PR #167, opened 2021-03-16
Implements account sugar functions

mvadari:

This PR implements the following account features:
- Get AccountData
- Get transaction history
    - Get payment history (just payment transactions)
- Account exists

### Context of Change

Python SDK sugar

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added tests for each new function. CI passes.


tests/integration/sugar/test_account.py
```diff
@@ -0,0 +1,45 @@
+from unittest import TestCase
+
+from tests.integration.it_utils import JSON_RPC_CLIENT
+from tests.integration.reusable_values import WALLET
+from xrpl.account import does_account_exist, get_account_transactions, get_balance
+from xrpl.wallet import Wallet, generate_faucet_wallet
+
+NEW_WALLET = generate_faucet_wallet(JSON_RPC_CLIENT)
```
ledhed2222, line 8: maybe wait to merge this after mine - i wanna have two faucets in the shared values so we can make cases like this faster

ledhed2222: my open PR has a `DESTINATION` wallet that is also generated from faucet

tests/integration/sugar/test_account.py
```diff
@@ -0,0 +1,45 @@
+from unittest import TestCase
```
ledhed2222, line 1: i don't think we need a `sugar` dir for integration tests. i get it (and you get it), but it doesn't match with a source dir. i can see either having a 1:1 dir struct like we have for unit tests, or a _less_ granular struct for integration tests. but i'm open to ideas. we can research this

ledhed2222: i will say for now, because we have deadlines, lets _not_ do the sugar dir. i might be wrong.

xrpl/account/__init__.py
```diff
@@ -1,10 +1,21 @@
...
+)
 
-__all__ = ["get_fee", "get_next_valid_seq_number", "get_balance", "get_account_info"]
+__all__ = [
+    "get_next_valid_seq_number",
+    "get_balance",
+    "get_account_info",
+    "get_account_payment_transactions",
+    "get_account_transactions",
+    "does_account_exist",
```
ledhed2222, line 20: i can't think of a better name, but something about this bothers me. @hammertoe - do you have any thoughts on naming functions in python that return bools?

xrpl/account/main.py
```diff
@@ -1,26 +1,33 @@
...
+        XRPLTransactionFailureException: if the transaction fails.
     """
-    response = client.request(Fee())
-    result = cast(Dict[str, Any], response.result)
-    return cast(str, result["drops"]["minimum_fee"])
+    try:
+        get_account_info(address, client)
+        return True
+    except XRPLTransactionFailureException as e:
+        if e.error_code == "actNotFound":
```
ledhed2222, line 28: what is this case? might be helpful to add a comment here

xrpl/account/transaction_history.py
```diff
@@ -0,0 +1,44 @@
+"""High-level methods to obtain information about account transaction history."""
+from typing import Any, Dict, List, cast
+
+from xrpl.clients import Client, XRPLTransactionFailureException
+from xrpl.models.requests import AccountTx
+
+
+def get_account_payment_transactions(
```
ledhed2222, line 8: nit - maybe put this below the other func. total nit :)

ledhed2222 (commented):
see comments - looks good but a few things

xrpl/account/__init__.py
mvadari: Yeah I agree, bool functions typically start with `is`, but I couldn't figure out the right wording for that. For example, `is_account_existing` and `is_account_in_existence` feel a bit awkward.

tests/integration/sugar/test_account.py
mvadari: This was created to make it easier to run all these tests at once - if they're in a folder that has an __init__, then you can run them all with one command. Happy to take it out though.

xrpl/account/main.py
ledhed2222: nevermind lol

xrpl/clients/exceptions.py
```diff
@@ -0,0 +1,24 @@
+"""General XRPL Transaction Exceptions."""
+from __future__ import annotations
+
+from xrpl import XRPLException
+
+
+class XRPLTransactionFailureException(XRPLException):
```
ledhed2222, line 7: for now, let's make this even more general so that it means a _request_ failure

xrpl/clients/exceptions.py
```diff
@@ -0,0 +1,24 @@
+"""General XRPL Transaction Exceptions."""
+from __future__ import annotations
+
+from xrpl import XRPLException
+
+
+class XRPLRequestFailureException(XRPLException):
+    """XRPL Transaction Exception, when the transaction fails."""
```
ledhed2222, line 8: please update

xrpl/clients/exceptions.py
```diff
@@ -0,0 +1,24 @@
+"""General XRPL Transaction Exceptions."""
```
ledhed2222, line 1: please update

ledhed2222 (approved):

PR #188, opened 2021-03-24
Add conversion utilities

mDuo13:
Adds some conversion utilities for feature-parity with ripple-lib:

- Convert from decimal (`int`, `float`, or `decimal.Decimal`) XRP to drops of XRP and back (to Decimal)
- Convert from XRP Ledger's "Ripple Epoch" time format to Python-native time types and back.

# Type of Change

- [x] New Feature
- [x] Tests
- [x] Documentation

### DIFF
--- a/xrpl/utils/time_conversions.py
+++ b/xrpl/utils/time_conversions.py
@@ -6,10 +6,14 @@ data types
 from datetime import datetime, timezone
 from typing import Union
 
+from typing_extensions import Final
+
 from xrpl import XRPLException
 
-RIPPLE_EPOCH = 946684800  #: The "Ripple Epoch" of 2000-01-01T00:00:00 UTC
-MAX_XRPL_TIME = 2 ** 32  #: The maximum time that can be expressed on the XRPL
+RIPPLE_EPOCH: Final[int] = 946684800  #: The "Ripple Epoch" of 2000-01-01T00:00:00 UTC
+MAX_XRPL_TIME: Final[int] = (
+    2 ** 32
+)  #: The maximum time that can be expressed on the XRPL
 
 
 def ripple_time_to_datetime(ripple_time: int) -> datetime:
--- a/xrpl/utils/time_conversions.py
+++ b/xrpl/utils/time_conversions.py
@@ -1,0 +1,130 @@
+"""
+Conversions between the XRP Ledger's 'Ripple Epoch' time and native time
+data types
+"""
+
+from datetime import datetime, timezone
+from typing import Union
+
+from xrpl import XRPLException
+
+RIPPLE_EPOCH = 946684800  #: The "Ripple Epoch" of 2000-01-01T00:00:00 UTC
+MAX_XRPL_TIME = 2 ** 32  #: The maximum time that can be expressed on the XRPL
+
+
+def ripple_time_to_datetime(ripple_time: int) -> datetime:
+    """
+    Convert from XRP Ledger 'Ripple Epoch' time to a UTC `datetime
+    <https://docs.python.org/3/library/datetime.html#datetime-objects>`_ object.
+
+    Args:
+        ripple_time: Whole seconds since the Ripple Epoch of 2001-01-01T00:00Z
+
+    Returns:
+        The equivalent time as a ``datetime`` instance.
+
+    Raises:
+        XRPLTimeRangeException: if the given ``ripple_time`` is not valid
+    """
+    if ripple_time < 0:
+        raise XRPLTimeRangeException(f"{ripple_time} is before the Ripple Epoch.")
+    if ripple_time > MAX_XRPL_TIME:
+        raise XRPLTimeRangeException(
+            f"{ripple_time} is larger than any time that can be expressed on"
+            f"the XRP Ledger."
+        )
+    timestamp = ripple_time + RIPPLE_EPOCH
+    return datetime.fromtimestamp(timestamp, timezone.utc)
+
+
+def datetime_to_ripple_time(dt: datetime) -> int:
+    """
+    Convert from a `datetime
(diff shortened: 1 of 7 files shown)


xrpl/utils/time_conversions.py
```diff
@@ -0,0 +1,130 @@
...
+data types
+"""
+
+from datetime import datetime, timezone
+from typing import Union
+
+from xrpl import XRPLException
+
+RIPPLE_EPOCH = 946684800  #: The "Ripple Epoch" of 2000-01-01T00:00:00 UTC
+MAX_XRPL_TIME = 2 ** 32  #: The maximum time that can be expressed on the XRPL
```
ledhed2222, line 12: please make these `Final`

```python

RIPPLE_EPOCH: Final[int] = .....
```

xrpl/utils/xrp_conversions.py
```diff
@@ -0,0 +1,101 @@
...
+
+from decimal import Context, Decimal, InvalidOperation, setcontext
+from re import match
+from typing import Union
+
+from xrpl import XRPLException
+
+ONE_DROP = Decimal("0.000001")  #: Indivisible unit of XRP
+MAX_XRP = Decimal(10 ** 11)  #: 100 billion decimal XRP
+MAX_DROPS = Decimal(10 ** 17)  #: Maximum possible number of drops of XRP
```
ledhed2222, line 11: ditto with final

xrpl/utils/xrp_conversions.py
```diff
@@ -0,0 +1,101 @@
...
+        Equivalent amount in drops of XRP
+
+    Raises:
+        XRPTypeException: if ``xrp`` is given as a string
+        XRPRangeException: if the given amount of XRP is invalid
+    """
+    if type(xrp) == str:  # type: ignore
+        # This protects people from passing drops to this function and getting
+        # a million times as many drops back.
+        raise XRPTypeException(
```
ledhed2222, line 37: nice!

xrpl/utils/xrp_conversions.py
```diff
@@ -0,0 +1,101 @@
...
+    setcontext(_DROPS_CONTEXT)
+    try:
+        xrp_d = Decimal(xrp)
+
+        if xrp_d < ONE_DROP and xrp_d != 0:
+            raise XRPRangeException(f"XRP amount {xrp} is too small.")
+        if xrp_d > MAX_XRP:
+            raise XRPRangeException(f"XRP amount {xrp} is too large.")
+    except InvalidOperation:
+        raise XRPRangeException(f"Not a valid amount of XRP: '{xrp}'")
```
ledhed2222, line 50: it's weird to me to `raise` within a `try`. also i think it's good to keep the body of `try` blocks as small as possible so readers can see exactly what code could raise. suggestion:
```python
try:
  xrp_d = Decimal(xrp)
except InvalidOperation:
  raise....
  
if xrp_d < ONE_DROP and xrp_d != 0:
    raise ....
    
if xrp_d > MAX_XRP:
   raise ....
   
 drops_amount = ....
 ```

File: xrpl/utils/xrp_conversions.py [hunk 5]
```diff
@@ -0,0 +1,101 @@
...
+
+    drops_amount = (xrp_d / ONE_DROP).quantize(Decimal(1))
+    drops_str = str(drops_amount)
+
+    # This should never happen, but is a precaution against Decimal doing
+    # something unexpected.
+    if not match(_DROPS_REGEX, drops_str):
```
ledhed2222, line 57: why not `fullmatch` instead of `match`

xrpl/utils/xrp_conversions.py
```diff
@@ -0,0 +1,101 @@
...
+        drops: String representing indivisible drops of XRP
+
+    Returns:
+        Decimal representation of the same amount of XRP
+
+    Raises:
+        XRPRangeException: if the given number of drops is invalid
+    """
+    setcontext(_DROPS_CONTEXT)
+    if not match(_DROPS_REGEX, drops):
```
ledhed2222, line 80: fullmatch

xrpl/utils/xrp_conversions.py
ledhed2222: but IMO this should be a vanilla `TypeException`, since that's exactly what it is

xrpl/utils/xrp_conversions.py
```diff
@@ -0,0 +1,101 @@
...
+class XRPRangeException(XRPLException):
+    """Exception for invalid XRP amounts."""
+
+    pass
+
+
+class XRPTypeException(XRPLException):
+    """Exception for string XRP amounts (use a Decimal for decimal XRP)."""
+
+    pass
```
ledhed2222, line 101: i don't think we need this second exception

ledhed2222 (changes requested):
a few comments - thanks for this!

xrpl/utils/xrp_conversions.py
mDuo13: Originally I had it like you wrote here, but the comparison itself (`xrp_d < ONE_DROP`) can raise if the caller passes in the string "NaN"

xrpl/utils/xrp_conversions.py
mDuo13: No particular reason; I'm used to using `match`, and wrote the regular expression to require a full match anyway the `^` and `$` in it. But I can switch it to fullmatch and simplify the regex instead.

xrpl/utils/xrp_conversions.py
mDuo13: Changed this to have an explicit check `if not xrp_d.is_finite():` instead

mvadari (approved):

ledhed2222 (approved):

PR #194, opened 2021-03-24
Adds request integration tests, fixes a few misc issues uncovered in testing

ledhed2222:

Title says it all. 

### Context of Change

Python SDK cleanup.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Tests pass.


xrpl/wallet/main.py
```diff
@@ -1,44 +1,43 @@
...
 
-from xrpl import CryptoAlgorithm
 from xrpl.core.keypairs import derive_classic_address, derive_keypair, generate_seed
 
 
 class Wallet:
     """The information needed to control an XRPL account."""
 
-    def __init__(self: Wallet, seed: str) -> None:
+    def __init__(self: Wallet, pub_key: str, priv_key: str) -> None:
```
ledhed2222, line 12: the reason i did this is to handle a case where a user wants to use a wallet that they did not generate in this session. maybe they don't know the seed? i'll confirm this though

xrpl/wallet/main.py
```diff
@@ -1,44 +1,43 @@
...
+            seed: A seed from which to derive public and private keys. If none
+                is provided, one will be generated for you automatically (using
+                Ed25519).
 
         Returns:
             The wallet that is generated from the given seed.
         """
-        seed = generate_seed(algorithm=crypto_algorithm)
-        return cls(seed)
+        pub_key, priv_key = derive_keypair(seed or generate_seed())
```
ledhed2222, line 39: why did you default SECP256K1? I _believe_ ED25519 is preferred.

xrpl/wallet/main.py
```diff
@@ -47,7 +46,10 @@ def __str__(self: Wallet) -> str:
         Returns:
             A string representation of a Wallet.
         """
-        return (
-            f"seed: {self.seed}\npub_key: {self.pub_key}\n"
-            f"priv_key: {self.priv_key}\nclassic_address: {self.classic_address}\n"
+        return "\n".join(
+            [
+                f"pub_key: {self.pub_key}",
+                "priv_key: -HIDDEN-",
```
ledhed2222, line 52: felt like a good idea not to print this, in the case users are turning this into a log

tests/integration/requests/test_submit_multisigned.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class TestSubmitMultisigned(TestCase):
+    def test_basic_functionality(self):
+        #
+        # Perform multisign
+        #
+        # NOTE: If you need to use xrpl-py for multisigning, please create an issue on
+        # the repo. We'd like to gauge interest in higher level multisigning
+        # functionality.
```
ledhed2222, line 56: added this b/c i'm embarrassed by how ugly this is lol

xrpl/wallet/main.py
```diff
@@ -1,44 +1,43 @@
...
+            pub_key: The public key for this Wallet.
+            priv_key: The private key for this Wallet.
+
         """
-        self.seed = seed
-        self.pub_key, self.priv_key = derive_keypair(self.seed)
+        self.pub_key = pub_key
+        self.priv_key = priv_key
         self.classic_address = derive_classic_address(self.pub_key)
         self.next_sequence_num = 0
```
ledhed2222, line 24: how are we going to handle this if users are loading an existing account? what happens in ripple-lib?

xrpl/models/transactions/transaction.py
```diff
@@ -114,7 +147,7 @@ class Transaction(BaseModel):
     memos: Optional[List[Memo]] = None
     signers: Optional[List[Signer]] = None
     source_tag: Optional[int] = None
-    signing_pub_key: Optional[str] = None
+    signing_pub_key: str = ""
```
ledhed2222, line 150: allows us to avoid a bunch of checks for `None`. AFAICT it can never be `None` anyway

xrpl/models/transactions/transaction.py
```diff
@@ -88,6 +88,39 @@ class Signer(BaseModel):
...
+        )
+
+    def to_dict(self: Signer) -> Dict[str, Any]:
+        """
+        Returns the dictionary representation of a Signer.
+
+        Returns:
+            The dictionary representation of a Signer.
+        """
+        return {"signer": super().to_dict()}
```
ledhed2222, line 122: this object needs to be wrapped in a `"signer"` key

xrpl/models/transactions/transaction.py
mDuo13: This is probably OK.

The XRP Ledger _does_ make a distinction between the binary form of transactions where `SigningPubKey` is present but empty (multi-signed transactions only) vs. ones where it is absent (unsigned transaction instructions). For multi-signed transactions the field needs to be serialized as empty, and in the latter case it needs to be added with the appropriate key when you serialize the transaction for signing.

xrpl/models/transactions/transaction.py
mDuo13: This is the convention `rippled` uses whenever it serializes an "Object in an array", so you'll see it in other places too like in [transaction metadata](https://xrpl.org/transaction-metadata.html#example-metadata). It's always:

```json
{
  "ObjectTypeName": {
      "object fields here": true,
      "etc.": "..."
  }
}
```

xrpl/wallet/main.py
mDuo13: The conventional way to create a wallet is from the seed, but it would be reasonable to have a `@classmethod` that derives the wallet from just a private key instead (the `pub_key` is derived from the private key anyway). In that case, the seed would be unknown but that should be fine.

xrpl/wallet/main.py
mDuo13: ripple-lib has:

- [generateXaddress](https://xrpl.org/rippleapi-reference.html#generatexaddress) - optionally takes explicitly specified entropy, doesn't let you specify the seed.
- [generateAddress (deprecated, nearly the same anyway)](https://xrpl.org/rippleapi-reference.html#generateaddress)
- [isValidSecret](https://xrpl.org/rippleapi-reference.html#isvalidsecret) to check a seed
- [deriveKeypair](https://xrpl.org/rippleapi-reference.html#derivekeypair) from seed
- [deriveAddress](https://xrpl.org/rippleapi-reference.html#deriveaddress) from pubkey

rippled's wallet_propose accepts the kitchen sink in seed formats, but weirdly specifying an actual private key isn't an option, and it doesn't expose any methods for going from private keys to public keys or from public keys to addresses.

xrpl/wallet/main.py
mDuo13: rippled does something similar in logs, but I forget what magic string it puts instead. It's `REDACTED` or something.

mDuo13 (changes requested):
My take is that the default `Wallet` constructor should have no required arguments, and should generate a seed from [its own source of strong entropy](https://docs.python.org/3/library/secrets.html) plus optionally more entropy the user supplies, and there should be `@classmethod` constructors for creating a Wallet from an existing seed or private key.

Also, all said constructors should accept a `test` flag and give you X-addresses with the test flag set accordingly.

xrpl/core/binarycodec/main.py
```diff
@@ -120,4 +116,4 @@ def _serialize_json(
     if suffix is not None:
         buffer += suffix
 
-    return cast(bytes, buffer.hex().upper())
+    return buffer.hex().upper()
```
mvadari, line 119: oof

xrpl/wallet/main.py
mvadari: I believe that was a part of the code Amie wrote, and I left it as-is because I didn't know that. That's fine.

mvadari (commented):
pls add an actual PR description

xrpl/wallet/main.py
ledhed2222: right right i get all that, but i mean let's say you have an account and its keys or seed. you've already made some transactions on it. now you want to wrap those keys in a `Wallet` object so you can use it with this lib. how do we correctly initialize the `next_sequence_num`, since in this case it is not `0`?

mDuo13: Ah, I think I see what you mean. IMO the user should call some account info lookup function and then set the next_sequence_num attribute directly.

Is it too late to rename that field to just `sequence`?

tests/integration/requests/test_submit_multisigned.py
```diff
@@ -0,0 +1,118 @@
...
+        tx_json = transaction_json_to_binary_codec_form(tx.to_dict())
+        first_sig = sign(
+            bytes.fromhex(
+                encode_for_multisigning(
+                    tx_json,
+                    FIRST_SIGNER.classic_address,
+                )
+            ),
+            FIRST_SIGNER.priv_key,
+        )
```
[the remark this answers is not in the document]
mDuo13: IMO `.encode()` should be an attribute on BaseModel. On transactions it could even decide whether it's for (regular) signing or not based on whether there's already a (filled out) `TxnSignature` field. For multi-signing, you might have to pass a specific option or something.

xrpl/wallet/main.py
```diff
@@ -15,48 +13,43 @@ class Wallet:
     details.
     """
 
-    def __init__(self: Wallet, seed: str) -> None:
+    def __init__(self: Wallet, public_key: str, private_key: str) -> None:
```
mDuo13, line 16: Before this PR:


```py
from xrpl.core.keypairs import generate_seed
my_wallet = Wallet(generate_seed())
```

After this PR:

```py
from xrpl.core.keypairs import generate_seed, derive_keypair
my_wallet = Wallet(*derive_keypair(generate_seed()))
```

How it should be:

```py
my_wallet = Wallet()
```

xrpl/wallet/main.py
```diff
@@ -15,48 +13,43 @@ class Wallet:
...
         """
-        #: The core value that is used to derive all other information about
-        #: this wallet. MUST be kept secret!
-        self.seed = seed
-
-        pk, sk = derive_keypair(self.seed)
         #: The public key that is used to identify this wallet's signatures, as
         #: a hexadecimal string.
-        self.pub_key = pk
+        self.public_key = public_key
```
mDuo13, line 26: Public key should be derivable from the private key anyway

mDuo13 (changes requested):

xrpl/wallet/main.py
mDuo13: To elaborate further:

It makes sense to me to have a classmethod that creates a wallet from some provided seed, and maybe another one for creating from a private key. But the default constructor should do a reasonable thing when you use it in the simplest way possible.

xrpl/wallet/main.py
```diff
@@ -15,48 +13,43 @@ class Wallet:
...
 
         #: The next available sequence number to use for transactions from this
         #: wallet.
         self.next_sequence_num = 0
 
     @classmethod
-    def create(
-        cls: Type[Wallet], crypto_algorithm: CryptoAlgorithm = CryptoAlgorithm.SECP256K1
-    ) -> Wallet:
+    def create(cls: Type[Wallet], seed: Optional[str] = None) -> Wallet:
```
mDuo13, line 39: I think you should still be able to pick a crypto algorithm explicitly. There have been a couple cases in the past and are likely to be more in the future where the user might want to choose a specific type. Also, people who are importing secp256k1 keys generated by other software might want to use this class to do that.

xrpl/wallet/main.py
mvadari: Could we initialize it to `None` instead, so it'll error if you try to use it on your own?

xrpl/wallet/main.py
mvadari: With the current API, you can actually do this:
```py
my_wallet = Wallet.create()
```

xrpl/models/requests/submit_multisigned.py
```diff
@@ -34,17 +30,5 @@ class SubmitMultisigned(Request):
 
     method: RequestMethod = field(default=RequestMethod.SUBMIT_MULTISIGNED, init=False)
     #: This field is required.
-    transaction: Transaction = REQUIRED  # type: ignore
+    tx_json: Transaction = REQUIRED  # type: ignore
```
mvadari, line 33: I believe the reason I renamed this was because it felt weird to have a param called `tx_json` that wasn't a JSON-like object.

mDuo13: That makes sense. The underlying API takes actual JSON of course, but if this is expecting an instance of the Transaction model then it makes sense to be different.

xrpl/wallet/main.py
mDuo13: Right, and that's fine, but what I'm saying is that it's backwards.

The _default constructor_ should do the most sensible thing with as few arguments as possible, and classmethods can provide more niche ways of instantiating a wallet.

xrpl/wallet/main.py
mDuo13: I still think this should derive the public key from the private key rather than requiring the user to pass in both.

xrpl/wallet/main.py
mDuo13: I was serious about renaming it to `sequence`, by the way. I would even be in favor of `seq` since people seem to make typos on the word "sequence" surprisingly often, and it's still pretty unambiguous.

Yeah, it's a breaking change, but if it significantly improves the developer experience (which I think it does) then better to do that now rather than changing it later when it breaks more people's code.

xrpl/wallet/main.py
mDuo13: I still think you should be able to request a non-default crypto algorithm here.

mDuo13 (changes requested):
I'm willing to be overruled on some of the changes I've suggested that haven't been addressed, but I would like to at least hear the reasoning first.

xrpl/models/requests/submit_multisigned.py
ledhed2222: for now let's please leave as is. i know what you guys are saying but at the moment we have a thing going with being 1:1 with the rippled API. can always change

xrpl/wallet/main.py
ledhed2222: @mDuo13 is right - my change earlier here was wrong. we should be using `seed` only for now. can later add support for private key if someone needs it

xrpl/wallet/main.py
ledhed2222: idea: add a required field to `__init__` of `seq_num`. then `create` will pass `0` to that explicitly whereas user will know they need to determine it

xrpl/wallet/main.py
ledhed2222: yeah i think my chagnes to `Wallet` need to be backed out, with the exception of defaulting to ED25519

xrpl/wallet/main.py
mvadari: Can we do this in the wallet-cleanup PR (I'll work on it today)? I'd like to get the rest of these changes pushed.

mvadari (approved):

mDuo13 (approved):
With the Wallet changes backed out, this is :+1:

Issue #189, opened 2021-03-24
Improve transaction signing/submission flow


mDuo13:
There are a lot of different ways of doing transaction signing and submission, but we've learned from experience what models work better than others, and the way I most recommend doing things is as follows:

1. Prepare Transaction (online)

    The main point of this is that you can inspect what you're signing before you even go get your secret keys out of secure storage. For offline transaction construction, you have to skip this step and provide things like `Fee` and `Sequence` yourself. (The `Wallet` class in xrpl-py is well-equipped to handle the Sequence part at least.)

    **Input:**

    - Incomplete transaction instructions (JSON or otherwise)
    - network client
    - (optional) instructions for auto-filling, such as max ledger offset (e.g. a number to add to the latest validated ledger to get `LastLedgerSequence`), fee options, path options, etc.
    
    **Output:**

    - Complete transaction instructions (JSON or otherwise) with _all_ required signing fields filled out, X-addresses converted to classic address & tag, and any other preprocessing.
        - Should provide a sensible `LastLedgerSequence` field by default, as well as any other best practices, with options to explicitly configure/disable them.

2. Sign Transaction (offline OK)

    **Input:**

    - Complete transaction instructions (JSON or otherwise) with _all_ required signing fields filled out
    - Seed or private key

    **Output:**

    - Signed transaction blob
    - (Optional) Signed transaction JSON.
    - Transaction ID (hash).

3. Submit-and-Verify (online)

    It's not a bad thing to also have a simple submit function that returns the preliminary result only.

    **Input:**

    - Signed transaction blob (or, potentially, signed transaction in JSON format) with a `LastLedgerSequence` parameter.
    - (Optional) Minimum ledger index. If not provided, assume that the minimum ledger is the latest validated ledger +1 at time of submission. (**Note:** _Not_ the current open ledger; during times of instability, there could be several closed but unvalidated ledgers, and a transaction could theoretically end up getting into a replacement for one of the closed ledgers.) This is useful because you can use this function to resubmit the signed transaction any number of times and it'll still work consistently and idempotently as long as you always provide the same minimum ledger index. To implement truly robust transaction submission, one would need to look up and save the original min ledger index alongside the signed blob _before_ submitting the transaction at all.

    Output:

    - The transaction's [final result code](https://xrpl.org/finality-of-results.html). Optionally it could raise an exception if the result code is anything other than `tesSUCCESS`.
 
----

There are several ways in which the current version of xrpl-py **DOES NOT** work this way:

[...]

PR #197, opened 2021-03-24
Adds autofilling capabilities to `safe_sign_transaction`

mvadari:

Title says it all. Also did some cleanup in tests to make them more reliable/less flaky.

### Context of Change

Python SDK cleanup.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

CI passes.


xrpl/transaction/reliable_submission.py
```diff
@@ -71,12 +71,6 @@ def send_reliable_submission(
         XRPLReliableSubmissionException: if the transaction fails or is misisng a
             `last_ledger_sequence` param.
     """
-    if transaction.last_ledger_sequence is None:
-        raise XRPLReliableSubmissionException(
-            "Reliable submission requires that the transaction has a "
-            "`last_ledger_sequence` parameter."
-        )
-
```
mvadari, line 79: It's auto-added in `safe_sign_transaction`

xrpl/transaction/main.py
```diff
@@ -29,25 +35,26 @@ def safe_sign_and_submit_transaction(
...
     """
-    tx_blob = safe_sign_transaction(transaction, wallet)
+    tx_blob = safe_sign_transaction(transaction, wallet, client)
     return submit_transaction_blob(tx_blob, client)
 
 
-def safe_sign_transaction(transaction: Transaction, wallet: Wallet) -> str:
+def safe_sign_transaction(
+    transaction: Transaction, wallet: Wallet, client: Client
+) -> str:
```
mvadari, line 44: I'm not sure it's a good idea to have a client variable in `sign` when it's supposed to be local (seems to be a little bit of mixed messages), the only other idea I have is an optional variable `should_autofill` that defaults to True, and if False, doesn't do any autofill (and then you don't need to provide a Client either).

mDuo13: See my discussion of the preferred transaction signing and submission flow (  #189 ).

Autofill functionality should move to a "prepare" step while signing remains completely offline. (Optionally there can be an online method that combines "prepare" and "sign" too.)

mDuo13 (changes requested):
I agree it's important to have this sort of functionality but it should be kept separate from signing, as described in #189. Having the preparation step be separate also gives more opportunity for flexibly filling in reasonable best-practice defaults for more fields (paths, flags, LastLedgerSequence, etc.).

xrpl/transaction/main.py
```diff
@@ -73,6 +80,83 @@ def submit_transaction_blob(
     return client.request(submit_request)
 
 
+def _prepare_transaction(
```
mDuo13, line 83: Some version of this function should be exposed directly to clients

xrpl/transaction/main.py
```diff
@@ -79,7 +105,83 @@ def submit_transaction(
...
+    transaction_json = transaction.to_dict()
+    if "sequence" not in transaction_json:
+        sequence = get_next_valid_seq_number(transaction_json["account"], client)
+        transaction_json["sequence"] = sequence
+    if "fee" not in transaction_json:
+        transaction_json["fee"] = get_fee(client)
+
+    if "last_ledger_sequence" not in transaction_json:
+        ledger_sequence = get_latest_validated_ledger_sequence(client)
+        transaction_json["last_ledger_sequence"] = ledger_sequence + _LEDGER_OFFSET
```
mDuo13, line 159: Future-feature: clients should be able to explicitly specify that they don't want a last ledger sequence on a transaction. ("I want all the defaults except don't let my transaction expire. Also, I want a number 4, no pickles.")

xrpl/transaction/main.py
```diff
@@ -79,7 +105,83 @@ def submit_transaction(
...
+        json[tag_field] = tag
+
+
+def _convert_to_classic_address(json: Dict[str, Any], field: str) -> None:
+    """
+    Mutates JSON-like dictionary to convert the given field from an X-address (if
+    applicable) to a classic address.
+    """
+    if field in json and is_valid_xaddress(json[field]):
+        json[field] = xaddress_to_classic_address(json[field])
```
mDuo13, line 184: Should we raise an exception if the user provides an X-address with a tag to a field that doesn't relate to tags? I'm honestly not sure, but it feels weird to use an X-address with a tag for `Authorize` for example when it would really authorize anything that shares the same underlying classic address.

mDuo13 (approved):
I still think the prepare function should be expanded and exposed to users, but this is good for v1.0.

Props for making sure the prepare/autofill stuff still works if you're using a regular key. :+1:

xrpl/transaction/main.py
mvadari: I agree it's weird, but I could see the case where you use an X-Address for everything so you just stick it in here too.

xrpl/transaction/main.py
```diff
@@ -44,17 +55,32 @@ def safe_sign_transaction(transaction: Transaction, wallet: Wallet) -> Transacti
...
-    # Increment the wallet sequence number, since we're about to use one.
-    wallet.next_sequence_num += 1
-    transaction_json = transaction_json_to_binary_codec_form(transaction.to_dict())
-    transaction_json["SigningPubKey"] = wallet.public_key
+    transaction_json = _prepare_transaction(transaction, wallet)
     serialized_for_signing = encode_for_signing(transaction_json)
     serialized_bytes = bytes.fromhex(serialized_for_signing)
     signature = sign(serialized_bytes, wallet.private_key)
     transaction_json["TxnSignature"] = signature
     return cast(Transaction, Transaction.from_xrpl(transaction_json))
```
ledhed2222, line 63: in the future, we're going to want to have generalized ability to copy a model and change/add/remove specific fields on the copy.

ledhed2222 (approved):

xrpl/transaction/main.py
mvadari: Added a backlog ticket

xrpl/transaction/main.py
mDuo13: Using an X-address with a tag and without a tag has different implications. (This gave me an interesting idea for how Deposit Preauthorization could be limited to a specific source tag, but it can't work that way currently.) So it might be surprising for users who think they've authorized only "Alice at Bitstamp" when really they've authorized "Bitstamp" so anyone could send to them from there.

Doesn't need to change now, but worth keeping an eye on.

Issue #190, opened 2021-03-24
generate_faucet_wallet() can stall


mDuo13:
I've noticed that `generate_faucet_wallet()` has started stalling until it fails with a timeout (40 seconds).

I suspect the `sleep()` loop has a race condition that can cause it to loop forever, but I'm not sure entirely where.

mDuo13:
Actually, I think I figured it out. It happens if the `client` you provide is connected to a different network than the testnet. (There is a devnet faucet, btw, so it might be useful to have that as an option. `https://faucet.devnet.rippletest.net/accounts`)

PR #200, opened 2021-03-24
Fixes issues with faucet generation

mvadari:

Responds to issue #190 and handles https://github.com/ripple/rippled/issues/3795

### Context of Change

Python SDK cleanup.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

CI passes. Added a test to check for the devnet case, the other case is too nondeterministic to really be able to test well.


xrpl/wallet/wallet_generation.py
```diff
@@ -48,33 +59,41 @@ def generate_faucet_wallet(client: Client, debug: bool = False) -> Wallet:
...
-            try:
-                wallet.next_sequence_num = get_next_valid_seq_number(address, client)
-            except XRPLRequestFailureException as e:
-                if e.error_code != "actNotFound":
-                    raise
-                # try again after waiting a bit
-                sleep(1)
-                wallet.next_sequence_num = get_next_valid_seq_number(address, client)
+            break
+        i += 1
```
ledhed2222, line 78: not understanding why it was necessary to move to a `while` loop here?

ledhed2222: oh i see it's to have a global timeout. this isn't a big deal, but what do you think about this refactor:

```python
is_funded = false
for _ in range(_TIMEOUT_SECONDS):
  sleep(1)
  if not is_funded:
    is_funded = _try_to_fund()
  else:
    next_seq_num = _try_to_get_next_seq()
    if next_seq_num is not None:
      wallet.net_sequence_num = nex_seq_num
      return
```

ledhed2222: not a big deal - more of a suggestion. this way you can have:

1) only one loop relating to the same TIMEOUT, and not have to keep track of elapsed seconds
2) move the actual network calls to nice small functions (you know how much i like those lol)

ledhed2222 (approved):
approved - i have a thought for you in there but take it or leave it

xrpl/wallet/wallet_generation.py
```diff
@@ -33,7 +37,14 @@ def generate_faucet_wallet(client: Client, debug: bool = False) -> Wallet:
         XRPLFaucetException: if an address could not be funded with the faucet.
         XRPLRequestFailureException: if a request to the ledger fails.
     """
-    timeout_seconds = 40
+    if "dev" in client.url:  # devnet
+        faucet_url = _DEV_FAUCET_URL
+    elif "altnet" in client.url or "test" in client.url:  # testnet
+        faucet_url = _TEST_FAUCET_URL
```
mDuo13, line 43: 👏

xrpl/wallet/wallet_generation.py
```diff
@@ -48,33 +59,41 @@ def generate_faucet_wallet(client: Client, debug: bool = False) -> Wallet:
         starting_balance = 0
 
     # Ask the faucet to send funds to the given address
-    post(url=FAUCET_URL, json={"destination": address})
+    post(url=faucet_url, json={"destination": address})
```
mDuo13, line 62: This call can return an error status code like 400 or 503 without raising. We should catch the result and check those.

xrpl/wallet/wallet_generation.py
mDuo13: I agree with this suggestion. The starting balance isn't strictly necessary, but the sequence number is (as long as we get it from validated data, since if you rely on non-validated data then #3795 applies again)

mDuo13 (changes requested):
I think we should have more error checking on the actual POST to the faucet, but otherwise looks good. :+1:

mDuo13 (approved):
I think there miiiiight be a race condition with the extra `balance_check()` on lines 60-63. I was a little confused by the need for this extra call at first, but eventually I figured out it's because this can method can be used to re-fund an existing account. If the delay from the faucet takes a while to come in, this could happen after the funding transaction has already succeeded, so then the "starting" balance won't change later in the sleep loop.

I guess a simple fix would be to put the initial balance check before the `post()` to the faucet.

mDuo13 (approved):
The simplified code is a nice improvement. :tada: 

Seems like we could easily add an optional parameter to put a wallet in the `generate_faucet_wallet()` call so one could use this exact same code to refill an existing address. The method name doesn't quite fit that scenario but the actual functionality is basically identical (the only difference being not calling `Wallet.create()`)

We don't need to do that, but it seems like an easy boost in utility at little cost.

PR #203, opened 2021-03-25
Binary codec: allow X-addresses alongside matching SourceTag/DestinationTag

mDuo13:
Replaces #172 since that PR broke when the repo moved.


ledhed2222 (approved):
hell yeah - ty!

PR #214, opened 2021-03-29
Cleans up sugar methods, fixes bugs in `from_dict`

### LINKED ISSUE #202: Sugar methods should raise on non-success response.
The "sugar" methods in the `xrpl.transaction` module should raise an exception when the server returns a error response rather than just silently returning a Response model object with `is_successful() = False`. (The methods in the `xrpl.account` module already do.) Possibly all responses should do that. 

It's weird and tiresome to have to check is_successful() on every call if you're doing a bunch of calls back to back; I'd rather just put a whole block of calls in a try/except and handle network errors or what-have-you all at once.

```
prelim_result = xrpl.transaction.submit_transaction_blob(signed, client)
print("Preliminary transaction result:", prelim_result)
if not prelim_result.is_successful():
    exit("Submit failed.")
tx_id = prelim_result.result["tx_json"]["hash"]
```

If applied to all responses, this would render #192 moot.

mvadari:

This PR:
* Raises an error in all sugar methods that make calls to rippled (resolves #202)
* Changes the structure of `sign` and `submit` (`sign` now returns a `Transaction` object instead of a blob, `submit` now takes in a `Transaction` object instead of a blob)
* Fixes bugs in `from_dict` (it didn't properly handle things like Lists, and this was uncovered while working on this PR)

### Context of Change

Python SDK cleanup

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added tests that caught the `from_dict` bugs.


xrpl/models/transactions/payment.py
```diff
@@ -55,7 +55,7 @@ class Payment(Transaction):
     destination: str = REQUIRED  # type: ignore
     destination_tag: Optional[int] = None
     invoice_id: Optional[str] = None  # TODO: should be a 256 bit hash
-    paths: Optional[List[Any]] = None
+    paths: Optional[List[Any]] = None  # TODO: should be better typed
```
ledhed2222, line 58: I believe this is just:

```python
from xprl.models.requests.path_find import PathStep

....

paths: Optional[List[List[PathStep]]]
```

we can fix this later though :)

xrpl/models/transactions/signer_list_set.py
```diff
@@ -56,6 +56,23 @@ def is_dict_of_model(cls: Type[SignerEntry], dictionary: Dict[str, Any]) -> bool
...
+            value: The value to construct the SignerEntry from.
+
+        Returns:
+            A new SignerEntry object, constructed using the given parameters.
+        """
+        if len(value) == 1 and "signer_entry" in value:
+            return cast(
+                SignerEntry, super(SignerEntry, cls).from_dict(value["signer_entry"])
+            )
+        return cast(SignerEntry, super(SignerEntry, cls).from_dict(value))
```
ledhed2222, line 74: i'm thinking that all this mussing of types and values is a good sign we don't have the optimal design. this solution we have is too complicated for the problem.

but again - for a later release

xrpl/transaction/main.py
```diff
@@ -52,25 +52,34 @@ def safe_sign_transaction(transaction: Transaction, wallet: Wallet) -> str:
...
+def submit_transaction(
+    transaction: Transaction,
     client: Client,
 ) -> Response:
     """
     Submits a transaction blob to the ledger.
 
     Args:
-        transaction_blob: the transaction blob to be submitted.
+        transaction: the transaction blob to be submitted.
```
ledhed2222, line 66: since we are changing the meaning of the argument to be a `Transaction` itself, let's update the docstring

ledhed2222 (approved):

xrpl/models/transactions/signer_list_set.py
mvadari: Sorry, I don't follow - do you mean the casting?

Issue #196, opened 2021-03-24
Partial Payments protections


mDuo13:
Messing up [Partial Payments](https://xrpl.org/partial-payments.html) is one of the costliest, common mistakes people make when building on the XRP Ledger.

People look at the "Amount" field of a successful transaction and assume that's how much it delivered, when actually, it may have delivered a minuscule amount instead.

ripple-lib attempts to protect against this in a couple ways, which we should adapt or improve on:

- `getTransaction()` and `getTransactions()` remove the field (`specification.destination.amount` in the ripple-lib specific format) from the parsed transaction response. You have to pass an option (`includeRawTransaction`) to get the original version with the potentially-misleading raw field.
- The outcome of `getTransaction()` shows a `balance_changes` object that parses through the metadata to show how much currency (of varying types) was actually delivered.
- It doesn't have any protections when calling the `rippled` APIs directly using `.request(methodname, options)`.

One thing we could do fairly safely is to rename the field from `Amount` field of Payments to `DeliverMax` in some or all cases. The JSON format is not really "canonical" anyway, only the binary format, so you can call the fields whatever you want as long as you know what they correspond to.

See also: https://github.com/ripple/rippled/issues/3484

PR #215, opened 2021-03-30
warn when receiving partial payment obj

ledhed2222:

emits warning when receiving partial payment

### Context of Change

https://github.com/XRPLF/xrpl-py/issues/196

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

before - no warning emitted when you receive a partial payment. now, a warning is logged.

## Test Plan

i tested this with the commented-out code. it's extremely brittle and we'll have to look at fixing that brittleness later. i commented this out because it's too brittle to be included in our overall suite.

### DIFF
--- a/xrpl/models/response.py
+++ b/xrpl/models/response.py
@@ -5,6 +5,7 @@ Represents fields common to all response types.
 """
 from __future__ import annotations
 
+import warnings
 from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Dict, List, Optional, Union
@@ -45,6 +46,16 @@ class Response(BaseModel):
     result: Union[List[Any], Dict[Any]] = REQUIRED  # type: ignore
     id: Optional[Union[int, str]] = None
     type: Optional[ResponseType] = None
+
+    def __post_init__(self: Response) -> None:
+        """Called by dataclasses immediately after __init__."""
+        super().__post_init__()
+        if self.contains_partial_payment():
+            warnings.warn(
+                """This response contains a partial payment, please confirm
+                the delivered amount is correct""",
+                stacklevel=2,
+            )
 
     def is_successful(self: Response) -> bool:
         """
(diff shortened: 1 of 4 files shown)


xrpl/models/response.py
```diff
@@ -53,3 +55,44 @@ def is_successful(self: Response) -> bool:
...
+        Returns whether the request contains at least one transactions with
+        the partial payment flag set.
+
+        Returns:
+            True if at least one transaction in this Response has the partial
+            payment flag set. False otherwise.
+        """
+        return self._do_contains_partial_payment(self.result)
+
+    def _do_contains_partial_payment(self: Response, val: Any) -> bool:
```
mvadari, line 70:
```suggestion
    def _do_contains_partial_payment(self: Response, val: Dict[str, Any]) -> bool:
```

mvadari (commented):
What makes the test super brittle? Can we make it better?

mvadari (approved):

xrpl/models/response.py
```diff
@@ -44,6 +47,16 @@ class Response(BaseModel):
     id: Optional[Union[int, str]] = None
     type: Optional[ResponseType] = None
 
+    def __post_init__(self: Response) -> None:
+        """Called by dataclasses immediately after __init__."""
+        super().__post_init__()
+        if self.contains_partial_payment():
+            warnings.warn(
+                """This response contains a partial payment, please confirm
```
mDuo13, line 55:
```suggestion
                """This response contains a partial payment. Please confirm
```

mDuo13 (approved):
Suggested a tiny text change, otherwise looks good.

I would prefer a stronger protection against partial payments since it's probably easy to miss the warnings, but it's fair that [there isn't exactly a consensus upstream about what should be done](https://github.com/ripple/rippled/issues/3484).

PR #225, opened 2021-04-04
Fixes issue with calculating IOU precision

N3TC4T:

Fixes issue with calculating precision for IOU values when value is negative or exponential notation

### Context of Change

If IOU value contains negative value or value is exponential notation and actual precision is as `_MAX_IOU_PRECISION`  this leads to `verify_iou_value` raise exception as calculation precision will return wrong value.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


N3TC4T:
I realized there is issue in calculating IOU precision values with scientific notation (5.443957975934530E-8), I will try to update this pull request with a fix for this issue as well

N3TC4T:
I'm still not sure if this is the right way as it won't work for value's with fixed points like `2E+2`.

probably this would be covert all the situations:

```
decimal_value = Decimal(value, Context(prec=100))
# fixed value
if(decimal_value == decimal_value.to_integral()):
    return len(decimal_value.quantize(Decimal(1), context=Context(prec=100)).as_tuple().digits)
return len(decimal_value.normalize(Context()).as_tuple().digits)
```

ledhed2222:
woo hoo! thanks for this PR. would you add test cases for scientific notation, as you pointed out? also, looks like this PR is not passing our linter. would you fix whatever it's complaining about? if you set up the `pre-commit` hooks as described in `CONTRIBUTING` you should see exactly what the issue is locally. let me know if you're having any issues setting that up :)

ledhed2222 (commented):
woo hoo! thanks for this PR. would you add test cases for scientific notation, as you pointed out? also, looks like this PR is not passing our linter. would you fix whatever it's complaining about? if you set up the pre-commit hooks as described in CONTRIBUTING you should see exactly what the issue is locally. let me know if you're having any issues setting that up :)

ledhed2222:
also, please add a note in `CHANGELOG` :)

ledhed2222:
hey @N3TC4T have you had a chance to take another look at this?

N3TC4T:
Hey Greg, sorry I was a little busy with releasing XUMM, I will try to update this pull request this weekend, thanks.

tests/unit/core/binarycodec/fixtures/data/data-driven-tests.json
```diff
@@ -2881,21 +2881,6 @@
...
-      },
-      "significant_digits": 1,
-      "type_id": 6,
-      "is_native": false,
-      "mantissa": "00038D7EA4C68000",
-      "type": "Amount",
-      "expected_hex": "E7C38D7EA4C6800000000000000000000000000055534400000000000000000000000000000000000000000000000001",
-      "is_negative": false,
-      "exponent": 62
-    },
```
mvadari, line 2898: Why was this test removed?

N3TC4T: as the calculation the IOU precision method is fixed, `100000000000000000000000000000000000000000000000000000000000000000000000000000` is not a valid IOU amount anymore, so it's failing under this method https://github.com/XRPLF/xrpl-py/blob/f2f80ccfd8385370362cc3344e4627b19995a030/xrpl/core/binarycodec/types/amount.py#L81-L109

mvadari (approved):
Awesome, thanks!

PR #229, opened 2021-04-06
Adds `get_hash` method to `Transaction` objects

mvadari:

This PR:
* Adds a `get_hash` method to `Transaction` objects
* Adds a `to_xrpl` method to `Transaction` objects
* Moves the `transaction_json_to_binary_codec_form` logic to `xrpl.models.transactions.transaction`
* Fixes a poorly-handled error case in `BaseModel.from_dict` that was caught in the process of implementing this change
* Uses `get_hash` in `send_reliable_submission`, which makes the method more reliable

### Context of Change

Python library 1.1.0

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added tests for the caught bug and for hashing. Ran integ tests to ensure that the reliable submission change still worked properly.


xrpl/models/base_model.py
```diff
@@ -120,6 +120,15 @@ def _from_dict_single_param(
...
             return param_value
 
+        if param_type.__module__ != "typing":
+            # Should be in `typing` if it reaches here
+            # If the type is a builtin type here, it means there was a mismatch
+            # somewhere above
+            raise XRPLModelException(
+                f"{param} expected a {param_type.__name__}, received a "
+                f"{type(param_value).__name__}"
+            )
```
mvadari, line 130: Ran into a poorly handled error while writing this test

xrpl/transaction/__init__.py
```diff
@@ -1,11 +1,11 @@
 """Top-level exports for the transaction package."""
+from xrpl.models.transactions.transaction import transaction_json_to_binary_codec_form
```
mvadari, line 2: I know this is bad form but I want to remove `transaction_json_to_binary_codec_form` from the public API at some point - I don't think it's necessary now that `to_xrpl` does that work. If this is the wrong way to handle this, I'm happy to change that

ledhed2222: instead of this, i think it would be clearer if you kept `transaction_json_to_binary_codec_form` but just had that call your new function

ledhed2222 (approved):
one suggestion

PR #233, opened 2021-04-08
Adds pseudo-transaction models

### LINKED ISSUE #201: Pseudo-transaction models
To be able to instantiate arbitrary transactions from the ledger, you need to also be able to recognize and handle [pseudo-transactions](https://xrpl.org/pseudo-transaction-types.html).

We should have models for these and a way to instantiate "any transaction you may have found in a ledger" into the appropriate model. That way people could monitor all ledger transactions, and interact with them in a Pythonic way.

mvadari:

This PR adds a PseudoTransaction base model, each of the 3 pseudo-transaction models, and tests for each models. It also adds some bugfixes that were uncovered when testing. It also moves `TransactionType` and `PseudoTransactionType` to its own folder in `xrpl.models.transactions`, to avoid circular import problems.

### Context of Change

Python library 1.1.0

Resolves #201 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added tests that convert to/from XRPL JSON format (since people can't send these to the ledger and will only receive them from the ledger).


xrpl/models/transactions/pseudo_transactions/enable_amendment.py
```diff
@@ -0,0 +1,66 @@
...
+    transaction_type: PseudoTransactionType = field(
+        default=PseudoTransactionType.ENABLE_AMENDMENT,
+        init=False,
+    )
+
+    #: The Flags value of the EnableAmendment pseudo-transaction indicates the status
+    #: of the amendment at the time of the ledger including the pseudo-transaction.
+    #: A Flags value of 0 (no flags) or an omitted Flags field indicates that the
+    #: amendment has been enabled, and applies to all ledgers afterward.
+    flags: int = 0
```
ledhed2222, line 66: this should be inherited from the base `Transaction`, no?

ledhed2222 (commented):
one question - otherwise excellent work!

xrpl/models/transactions/pseudo_transactions/enable_amendment.py
mvadari: Yeah, it is, but I wanted to override the docs.

ledhed2222: should be a `Union[int, List[int]]`

ledhed2222 (approved):

Issue #191, opened 2021-03-24
xrpl.transaction.get_transaction_from_hash() should take optional min_ledger and max_ledger params


mDuo13:
The `min_ledger` and `max_ledger` params to the [tx method](https://xrpl.org/tx.html) help to reliably determine the outcome of a transaction. The `get_transaction_from_hash()` helper function should at least take `min_ledger` (it can choose `max_ledger` based on the `LastLedgerSequence`) and use them to more smartly categorize failures:

- Transaction isn't in specified range (`searched_all` is true)
- Server doesn't have sufficient ledger history, which can be further split into more results if you know the server's current max validated ledger:
    - Result isn't final (max_ledger is not yet validated)
    - Server is missing necessary history

florent-uzio:
I have been working on this one and I am waiting to have access rights to push my local branch.
@mvadari fyi

PR #234, opened 2021-04-09
Included binary, min_ledger and max_ledger for get_transaction_from_hash

florent-uzio:

In this PR I added the `binary`, `min_ledger` and `max_ledger` of the `tx` method as described in https://xrpl.org/tx.html#tx

I also added some tests in `tests/integration/sugar/test_transaction.py`.

It solves this issue: https://github.com/XRPLF/xrpl-py/issues/191

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

The `get_transaction_from_hash` method now accepts `binary`, `min_ledger` and `max_ledger` parameters.

## Test Plan

I included three new tests:
1. We submit a simple payment and then retrieve this transaction with `get_transaction_from_hash` without using the new parameters.
2. We submit a simple payment and then retrieve this transaction with `get_transaction_from_hash` with `binary=true`. We make sure the response includes a `hash` field.
3. We submit a simple payment and then retrieve this transaction with `get_transaction_from_hash` with `binary=false`, `min_ledger=payment ledger index - 500` and `max_ledger=payment ledger index + 500`.

### DIFF
--- a/xrpl/transaction/ledger.py
+++ b/xrpl/transaction/ledger.py
@@ -1,22 +1,36 @@
 """High-level methods that fetch transaction information from the XRP Ledger."""
 
-from typing import Any, Dict, cast
+from typing import Any, Dict, Optional, cast
 
 from xrpl.clients import Client, XRPLRequestFailureException
 from xrpl.models.requests import Tx
 from xrpl.models.response import Response
 
 
-def get_transaction_from_hash(tx_hash: str, client: Client, binary: bool = False, min_ledger: int = None, max_ledger: int = None) -> Response:
+def get_transaction_from_hash(
+    tx_hash: str,
+    client: Client,
+    binary: bool = False,
+    min_ledger: Optional[int] = None,
+    max_ledger: Optional[int] = None,
+) -> Response:
     """
     Given a transaction hash, fetch the corresponding transaction from the ledger.
 
     Args:
         tx_hash: the transaction hash.
         client: the network client used to communicate with a rippled node.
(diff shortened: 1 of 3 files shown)


xrpl/transaction/ledger.py
```diff
@@ -1,27 +1,51 @@
...
+    min_ledger: Optional[int] = None,
+    max_ledger: Optional[int] = None,
+) -> Response:
     """
     Given a transaction hash, fetch the corresponding transaction from the ledger.
 
     Args:
         tx_hash: the transaction hash.
         client: the network client used to communicate with a rippled node.
+        binary: (Optional) If true, return transaction data and metadata as binary
```
ledhed2222, line 23: I don't think there's any need to mark it as optional because the docs will show the type of the param.

ledhed2222 (commented):
Great! Please add a comment about this change in the CHANGELOG 💪

xrpl/transaction/ledger.py
florent-uzio: I removed `(Optional)`

florent-uzio:
Thanks @ledhed2222 , I made the changes you asked for in your comments.

mvadari:
Not sure what happened when you rebased/merged but it looks like you've somehow picked up some extra commits - can you fix that?

florent-uzio:
@mvadari , I fixed it, let me know if that's ok for the review.

mvadari (approved):
Looks good to me! Thanks!

PR #235, opened 2021-04-09
Adds option for a Transaction's flags to be a List of ints

mvadari:

Title says it all.

### Context of Change

Python library 1.1.0

### Type of Change

## Test Plan

Added simple test. CI passes.

### DIFF
--- a/xrpl/models/transactions/transaction.py
+++ b/xrpl/models/transactions/transaction.py
@@ -168,9 +168,9 @@ class Transaction(BaseModel):
     #: details.
     account_txn_id: Optional[str] = None
 
-    #: A bitwise map of flags modifying this transaction's behavior. See `Flags
-    #: Field <https://xrpl.org/transaction-common-fields.html#flags-field>`_ for
-    #: more details.
+    #: A List of flags, or a bitwise map of flags, modifying this transaction's
+    #: behavior. See `Flags Field
+    #: <https://xrpl.org/transaction-common-fields.html#flags-field>`_ for more details.
     flags: Union[int, List[int]] = 0
 
     #: The highest ledger index this transaction can appear in. Specifying this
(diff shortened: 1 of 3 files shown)


xrpl/models/transactions/transaction.py
```diff
@@ -209,7 +209,15 @@ def to_dict(self: Transaction) -> Dict[str, Any]:
...
-        return {**super().to_dict(), "transaction_type": self.transaction_type.value}
+        dictionary = {
+            **super().to_dict(),
+            "transaction_type": self.transaction_type.value,
+        }
+        if isinstance(self.flags, list):
+            dictionary["flags"] = 0
+            for flag in self.flags:
+                dictionary["flags"] |= flag
+        return dictionary
```
ledhed2222, line 220: suggestion:

```python
def to_dict(self: Transaction) -> Dict[str, Any]:
  return {**super().to_dict(), "transaction_type": self.transaction_type.value, "flags": self._flags_to_dict()}
  
def _flags_to_dict(self: Transaction) -> int:
  if isinstance(self.flags, int):
    return self.flags
  accumulator = 0
  for flag in self.flags:
    accumulator |= flag
  return accumulator
```

factoring out into helper - just readability imo. take it or leave it

xrpl/models/transactions/transaction.py
```diff
@@ -256,7 +264,10 @@ def has_flag(self: Transaction, flag: int) -> bool:
         Returns:
             Whether the transaction has the given flag value set.
         """
-        return self.flags & flag != 0
+        if isinstance(self.flags, int):
+            return self.flags & flag != 0
+        else:  # is List[int]
+            return flag in self.flags
```
ledhed2222, line 270: don't need the `else` here. could also be expressed as:

```python
return (isinstance(self.flags, int) and self.flags & flag != 0) or flag in self.flags
```

ledhed2222 (approved):

ledhed2222:
changelog!

xrpl/models/transactions/transaction.py
mvadari: I think that's messier to read

PR #236, opened 2021-04-13
Adds additional X-Address support

mvadari:

This PR:
- Adds X-address support to the account sugar and the binary codec (the binary codec previously did not support X-Addresses for e.g. the issuer field of issued currency)
- Adds a method to wallets to obtain an X-Address
- Renames some non-exposed methods/method variables to better reflect what they do

### Context of Change

Python library 1.1.0

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added tests for all added functionality. CI passes.


xrpl/account/transaction_history.py
```diff
@@ -2,16 +2,17 @@
...
 
 
-def get_latest_transaction(account: str, client: Client) -> Response:
+def get_latest_transaction(address: str, client: Client) -> Response:
     """
     Fetches the most recent transaction on the ledger associated with an account.
 
     Args:
-        account: the classic address of the account.
+        address: the account to query.
```
mvadari, line 15: Does this count as a breaking change because people might be using keyword args?

tests/unit/core/binarycodec/test_main.py
```diff
@@ -253,7 +253,7 @@ def test_xaddress_null_tag(self):
         self.assertEqual(encode(json_null_x), encode(json_null_r))
 
     def test_xaddress_invalid(self):
-        with self.assertRaises(ValueError):
+        with self.assertRaises(XRPLBinaryCodecException):
```
mvadari, line 256: This is technically a bugfix - it is better for this error to raise `XRPLBinaryCodecException` for this than to raise `ValueError`

xrpl/core/keypairs/helpers.py
```diff
@@ -15,11 +15,11 @@ def sha512_first_half(message: bytes) -> bytes:
...
 
 
-def get_account_id(public_key: bytes) -> bytes:
+def get_classic_address(public_key: bytes) -> bytes:
     """
-    Returns the account ID for a given public key. See
+    Returns the classic address for a given public key. See
     https://xrpl.org/cryptographic-keys.html#account-id-and-address
-    to learn about the relationship between keys and account IDs.
+    to learn about the relationship between keys and classic addresses.
```
mvadari, line 22: This is not an exposed method so I don't think this counts as a breaking change.

xrpl/core/binarycodec/types/account_id.py
```diff
@@ -59,7 +65,16 @@ def from_value(cls: Type[AccountID], value: str) -> AccountID:
         if _HEX_REGEX.fullmatch(value):
             return cls(bytes.fromhex(value))
         # base58 case
-        return cls(decode_classic_address(value))
+        if is_valid_classic_address(value):
+            return cls(decode_classic_address(value))
+        if is_valid_xaddress(value):
+            classic_address, _, _ = xaddress_to_classic_address(value)
```
natenichols, line 71: Do we handle the Destination/Source Tags somewhere else in the binarycodec?

natenichols:
I left one question, but otherwise all changes make sense to me.

xrpl/core/binarycodec/types/account_id.py
mvadari: Yep, happens in https://github.com/XRPLF/xrpl-py/blob/master/xrpl/core/binarycodec/types/serialized_dict.py

mvadari:
I went through all the other folders and couldn't find any other instances of using addresses.

natenichols (approved):
LGTM 👍

xrpl/account/transaction_history.py
ledhed2222: yes and i don't see the need to change the variable name here.

xrpl/core/keypairs/helpers.py
ledhed2222: agree but not sure why you are changing the name. `account_id` has specific meaning if you look at the linked article

ledhed2222 (commented):
a few comments

ledhed2222 (approved):

Issue #213, opened 2021-03-26
Protections against high fee values


mDuo13:
A common mistake people make is fat-fingering the `Fee` value of a transaction, or letting a script auto-fill it during a brief moment of very high load. There have been several cases of people burning whole XRP (or even thousands of XRP!) to send payments for fractions of that value.

By default, ripple-lib won't let you sign any payment with a max fee higher than a default limit of 2 XRP. (You can configure it in the [API constructor parameters, `maxFeeXRP`](https://xrpl.org/rippleapi-reference.html#parameters).) Even 2 XRP is a pretty high fee given that most transactions only need about 0.000010 XRP and even a [special case](https://xrpl.org/transaction-cost.html#special-transaction-costs) like an EscrowFinish with a big fulfillment probably doesn't go above 0.001000 XRP.

The only big exception is AccountDelete, which requires a full 5 XRP. But deleting accounts is (and probably should be) a rare, unusual operation so it makes sense to require the user to use a special override to sign a transaction with a `Fee` value that high.

In short, **We should raise an error when constructing or signing a transaction whose fee value is more than 2 whole XRP** unless the user explicitly specifies an option to allow a higher maximum fee.

florent-uzio:
I am working on this one @mvadari and @ledhed2222 fyi

PR #239, opened 2021-04-15
Optional check to verify the Transaction fee is not too high

florent-uzio:

This PR solves: https://github.com/XRPLF/xrpl-py/issues/213
1. The Transaction fee default value is 2 XRP. If the fee is higher, an `XRPLException` is raised.
2. The `Client` (JsonRpcClient) now takes a `ClientParameters` parameter which allows overriding some default values. Today only `max_fee` is in `ClientParameters` to replicate `maxFeeXRP` in [ripple-lib](https://xrpl.org/rippleapi-reference.html#parameters).

### Context of Change

It's a new feature. I took ripple-lib as an example and found it easy and intuitive to implement the `parameters` in the `Client`.

### Type of Change

- [X] Documentation Updates
- [ ] Release

## Test Plan

I added multiple tests in `tests/integration/transactions` to make sure the transaction submission was failing if `fee` is higher than 2 XRP with the default JsonRpcClient. 
Also if `max_fee` is set and higher than 2 XRP, the Transactions should be successful.


mvadari:
Looks like you've got the same merge issue as before

docs/source/xrpl.core.binarycodec.binary_wrappers.rst
```diff
@@ -0,0 +1,29 @@
+xrpl.core.binarycodec.binary\_wrappers package
+==============================================
+
+Submodules
```
mvadari, line 4: All of these docs files should be deleted - they're not needed, poetry just red-adds them every time an `__init__` file is edited

xrpl/transaction/main.py
```diff
@@ -40,24 +41,25 @@ def safe_sign_and_submit_transaction(
...
     else:
-        transaction = safe_sign_transaction(transaction, wallet)
+        transaction = safe_sign_transaction(transaction, wallet, client)
     return submit_transaction(transaction, client)
 
 
 def safe_sign_transaction(
-    transaction: Transaction,
-    wallet: Wallet,
+    transaction: Transaction, wallet: Wallet, client: Client
```
mvadari, line 49: This method should not require a `Client` - the point of this method is that it's not sending any requests to the ledger, it's signing the Transaction object entirely offline. This would also be a breaking change and shouldn't be introduced in a minor release.

xrpl/clients/client.py
```diff
@@ -3,21 +3,24 @@
...
+from xrpl.models.parameters import ClientParameters
 from xrpl.models.requests.request import Request
 from xrpl.models.response import Response
 
 
 class Client(ABC):
     """Interface for all network clients to follow."""
 
-    def __init__(self: Client, url: str) -> None:
+    def __init__(self: Client, url: str, client_parameters: ClientParameters) -> None:
```
mvadari, line 14: This is much more of a JS paradigm than a Python paradigm - for Python you should just have an additional optional param. 
This would also be considered a breaking change and shouldn't be introduced at this time, but if you switch to an optional param then it wouldn't be a breaking change.

xrpl/clients/client.py
```diff
@@ -3,21 +3,24 @@
...
         """
         Constructs a Client.
 
         Arguments:
             url: The URL of the rippled node to submit requests to.
+            client_parameters: The parameters to update some default values
         """
```
mvadari, line 21:
```suggestion
    def __init__(self: Client, url: str, max_fee: str = "2000000") -> None:
        """
        Constructs a Client.

        Arguments:
            url: The URL of the rippled node to submit requests to.
            max_fee: The default maximum fee that should be allowed by the client.
        """
```

xrpl/clients/client.py
mvadari: I'm also not convinced that this should be something in the client - I feel like this should be a constant that exists in `xrpl.transaction` with an optional (defaulted to True) check that determines if the fee is too high.

mvadari (changes requested):

docs/source/xrpl.core.binarycodec.binary_wrappers.rst
florent-uzio: I see thank you, I deleted them

xrpl/clients/client.py
florent-uzio: Thanks for your useful suggestion. I removed `client_parameters` from the Client and implemented that fee check, in `xrpl.transaction`. The check is an optional field (default to True).
I feel it's much cleaner and easier to read now, let me know what you think @mvadari

xrpl/transaction/main.py
florent-uzio: Understood thank you, I removed the client.

florent-uzio:
The Client doesn't include the ability to define `max_fee` anymore. Instead, the fee check is happening in `xrpl.transactions.main.py`.
There is an optional boolean (default to True), which allows this check to happen or not during the transaction signing.

tests/integration/transactions/test_account_delete.py
```diff
@@ -25,7 +26,7 @@ def test_all_fields(self):
             destination=DESTINATION.classic_address,
             destination_tag=DESTINATION_TAG,
         )
-        response = submit_transaction(account_delete, WALLET)
+        response = submit_transaction(account_delete, WALLET, JSON_RPC_CLIENT, False)
```
mvadari, line 29: The client doesn't need to be passed in here
```suggestion
        response = submit_transaction(account_delete, WALLET, check_fee=False)
```

tests/integration/transactions/test_account_delete.py
```diff
@@ -38,3 +39,17 @@ def test_all_fields(self):
...
+            # GIVEN a new AccountDelete transaction
+            account_delete = AccountDelete(
+                account=ACCOUNT,
+                # WITH fee higher than 2 XRP
+                fee=FEE,
+                sequence=WALLET.sequence,
+                destination=DESTINATION.classic_address,
+                destination_tag=DESTINATION_TAG,
+            )
+            submit_transaction(account_delete, WALLET)
```
mvadari, line 55: This test should be in `tests/integration/sugar/test_transaction`, because it's testing transaction sugar

tests/integration/transactions/test_account_set.py
```diff
@@ -43,3 +45,21 @@ def test_all_fields_minus_set_flag(self):
...
+            email_hash=EMAIL_HASH,
+            message_key=MESSAGE_KEY,
+            transfer_rate=TRANSFER_RATE,
+            tick_size=TICK_SIZE,
+            # WITH fee higher than 2 XRP
+            fee=FEE,
+        )
+        # We expect an XRPLException to be raised
+        with self.assertRaises(XRPLException):
+            submit_transaction(account_set, WALLET)
```
mvadari, line 65: same here

tests/integration/transactions/test_check_cancel.py
```diff
@@ -28,3 +30,16 @@ def test_all_fields(self):
...
+        check_cancel = CheckCancel(
+            account=ACCOUNT,
+            sequence=WALLET.sequence,
+            check_id=CHECK_ID,
+            # WITH fee higher than 2 XRP
+            fee=FEE,
+        )
+        # We expect an XRPLException to be raised
+        with self.assertRaises(XRPLException):
+            submit_transaction(check_cancel, WALLET)
```
mvadari, line 45: There doesn't need to be a test for every single transaction type - one test should be enough

xrpl/transaction/main.py
```diff
@@ -33,31 +39,36 @@ def safe_sign_and_submit_transaction(
         wallet: the wallet with which to sign the transaction.
         client: the network client with which to submit the transaction.
         autofill: whether to autofill the relevant fields. Defaults to True.
+        check_fee: whether to check the fee is higher than 2 XRP. Defaults to True.
```
mvadari, line 42:
```suggestion
        check_fee: whether to check if the fee is higher than 2 XRP. Defaults to True.
```

xrpl/transaction/main.py
```diff
@@ -198,3 +212,16 @@ def transaction_json_to_binary_codec_form(dictionary: Dict[str, Any]) -> Dict[st
...
+def _check_fee(transaction: Transaction) -> None:
+    """Checks the Transaction fee"""
+    # Checks that no Transaction can have fee higher than 2 XRP
+    if transaction.fee is not None and int(transaction.fee) > int(_MAX_FEE):
+        raise XRPLException(
+            "Fee value: "
+            + str(drops_to_xrp(transaction.fee))
+            + " XRP exceeds the default "
+            + str(drops_to_xrp(_MAX_FEE))
+            + " XRP limit."
```
mvadari, line 226:
```suggestion
            + " maximum XRP fee limit."
```

mvadari (changes requested):

tests/integration/transactions/test_account_delete.py
florent-uzio: Thanks, I updated that

tests/integration/transactions/test_account_delete.py
florent-uzio: I moved it.

tests/integration/transactions/test_check_cancel.py
florent-uzio: I removed all of them except two: AccountDelete and AccountSet

xrpl/transaction/main.py
florent-uzio: Updated

xrpl/transaction/main.py
florent-uzio: Thanks, it's updated

florent-uzio:
Hi @mvadari , I made the changes requested. Let me know if you have more comments. Thanks

N3TC4T:
Hey @florent-uzio, thanks for the pull request. there is some things I need to mention.

as I see the `_MAX_FEE` is hard coded, the base network fee can be changed in future [Tweet Link](https://twitter.com/joelkatz/status/1380974867965648897?s=21) so maybe in future this value would consider as too low or too hight, and also the transaction fee can be higher in different transaction types in different situations, so I think applying this should not fail for example the  `AccountDelete` transaction with fixed `5` XRP fee.

I think better approach would be to fetch the base network fee and calculate the fee, base on the network fee and also transaction conditions and if calculated fee + ROUNUP is higher than current fee then raise exception.

This is a sample code in js:

```
// NOTE: this value is better to be fetched from network
// default to 12 drops
const netFee = 12

// 10 drops × (33 + (Fulfillment size in bytes ÷ 16))
if (this.Type === 'EscrowFinish' && this.Fulfillment) {
    baseFee = new BigNumber(netFee).multipliedBy(
        new BigNumber(Buffer.from(this.Fulfillment).length).dividedBy(16).plus(33),
    );
}

// AccountDelete tx have fixed 5 XRP fee
if (this.Type === 'AccountDelete') {
    baseFee = new BigNumber(5).multipliedBy(1000000);
}

// 10 drops × (1 + Number of Signatures Provided)
if (this.Signers.length > 0) {
    baseFee = new BigNumber(this.Signers.length).plus(1).multipliedBy(netFee).plus(baseFee);
}

if (!baseFee) {
    baseFee = new BigNumber(netFee);
}

return baseFee.toFixed(0, BigNumber.ROUND_UP);
```

N3TC4T:
I also need to mention this in most case's the `2` XRP fee is higher than enough and also good to prevent users to spend too much fee when network is busy . so probably a simple check for `AccountDelete` transaction and using `5000000` drop instead of `_MAX_FEE` would be still a good solution.

xrpl/transaction/main.py
```diff
@@ -198,3 +230,67 @@ def transaction_json_to_binary_codec_form(dictionary: Dict[str, Any]) -> Dict[st
...
+    else:
+        net_fee = int(get_fee(client))  # Usually 0.00001 XRP (10 drops)
+
+    base_fee = net_fee
+
+    # EscrowFinish Transaction with Fulfillment
+    # https://xrpl.org/escrowfinish.html#escrowfinish-fields
+    if transaction.transaction_type == TransactionType.ESCROW_FINISH:
+        escrow_finish = cast(EscrowFinish, transaction)
+        if escrow_finish.fulfillment:
```
mvadari, line 281:
```suggestion
        if escrow_finish.fulfillment is not None:
```

xrpl/transaction/main.py
```diff
@@ -198,3 +230,67 @@ def transaction_json_to_binary_codec_form(dictionary: Dict[str, Any]) -> Dict[st
...
+            fulfillment_bytes = escrow_finish.fulfillment.encode("ascii")
+            # 10 drops × (33 + (Fulfillment size in bytes / 16))
+            base_fee = math.ceil(net_fee * (33 + (len(fulfillment_bytes) / 16)))
+
+    # AccountDelete Transaction
+    if transaction.transaction_type == TransactionType.ACCOUNT_DELETE:
+        base_fee = 5000000
```
mvadari, line 288: This should be a named constant instead of a buried magic number

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Instance method in `Transaction` objects to calculate their hashes locally
 - Option for `Transaction.flags` to be a `List` of `int`s instead of just an `int`
 - Optional parameters for `get_transaction_from_hash`: `binary`, `min_ledger` and `max_ledger`
+- Optional parameter to check if the `Transaction` fee is not too high before signing it
+- Internal function to calculate the expected fee based on the `Transaction` type
```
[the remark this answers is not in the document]
florent-uzio: Understood, I removed it

tests/integration/it_utils.py
```diff
@@ -15,9 +15,16 @@
...
+    wallet: Wallet,
+    client: Client = JSON_RPC_CLIENT,
+    check_fee: bool = True,
 ) -> Response:
     """Signs and submits a transaction to the XRPL."""
+    if check_fee is False:
+        return safe_sign_and_submit_transaction(
+            transaction, wallet, client, True, check_fee
+        )
     return safe_sign_and_submit_transaction(transaction, wallet, client)
```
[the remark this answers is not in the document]
florent-uzio: Yes, I updated it, thanks

tests/integration/sugar/test_transaction.py
```diff
@@ -177,3 +190,104 @@ def test_get_transaction_from_hash_with_min_max_ledgers(self):
...
+            fee=FEE,
+            sequence=WALLET.sequence,
+            destination=DESTINATION,
+            destination_tag=DESTINATION_TAG,
+        )
+        # We expect the calculated fee to be 5000000 drops (5 XRP)
+        self.assertEqual(
+            _calculate_fee_per_transaction_type(account_delete, JSON_RPC_CLIENT),
+            "5000000",
+        )
```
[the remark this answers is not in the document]
florent-uzio: Ok I see, I will keep that in mind for the future.

tests/integration/transactions/test_trust_set.py
```diff
@@ -14,7 +14,7 @@ def test_basic_functionality(self):
             TrustSet(
                 account=WALLET.classic_address,
                 sequence=WALLET.sequence,
-                flags=TrustSetFlag.TF_SET_NO_RIPPLE,
+                flags=TrustSetFlag.TF_SET_NO_RIPPLE.value,
```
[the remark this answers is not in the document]
florent-uzio: Ok I removed it

xrpl/transaction/main.py
```diff
@@ -198,3 +230,67 @@ def transaction_json_to_binary_codec_form(dictionary: Dict[str, Any]) -> Dict[st
...
+
+    Args:
+        transaction: the Transaction to be submitted.
+        client: the network client with which to submit the transaction.
+
+    Returns:
+        The expected Transaction fee in drops
+    """
+    # Reference Transaction (Most transactions)
+    if not client:
```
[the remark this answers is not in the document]
florent-uzio: Thanks, I updated it

xrpl/transaction/main.py
florent-uzio: I created a constant for it

florent-uzio:
Hi @mvadari, I made the changes. As usual, let me know if you have more comments.
I am also using public methods in the tests instead of private methods.
Thanks

mvadari (approved):

PR #241, opened 2021-04-26
Implements a WebSocket client

mvadari:

Title says it all. This version handles both quick requests (like most JSON RPC connections) as well as longer connections (such as for `subscribe`).

### Context of Change

Python library v1.1

### Type of Change

## Test Plan

Added a simple websocket test, and one transaction test. CI passes.

## Future Tasks
* Make sure all tests that are run on the JSON client are also run on the websocket client
* Add sugar methods for subscribe/unsubscribe


xrpl/clients/websocket_client.py
```diff
@@ -0,0 +1,119 @@
...
+            The response from the server, as a Response object.
+        """
+        formatted_request = request_to_websocket(request_object)
+        if "id" not in formatted_request:
+            formatted_request["id"] = "request_{}".format(formatted_request["command"])
+        async with websockets.connect(self.url) as websocket:
+            await websocket.send(json.dumps(formatted_request))
+            response = await websocket.recv()
+            response_dict = json.loads(response)
+            assert response_dict["id"] == formatted_request["id"]
```
ledhed2222, line 47: we should be raising a custom error here, IMO, instead of just an `AssertionError`

xrpl/clients/websocket_client.py
```diff
@@ -0,0 +1,119 @@
...
+            formatted_request["id"] = "request_{}".format(formatted_request["command"])
+
+        async with websockets.connect(self.url) as websocket:
+            await websocket.send(json.dumps(formatted_request))
+            try:
+                async for message in websocket:
+                    await websocket.send(json.dumps(formatted_request))
+                    response = await websocket.recv()
+                    response_dict = json.loads(response)
+                    assert response_dict["id"] == formatted_request["id"]
```
ledhed2222, line 91: ditto. probably should be a private func becuase used multiple times

xrpl/clients/websocket_client.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Arguments:
+            request_object: An object representing information about a rippled request.
+            handler: The method that handles the Response objects from the server.
+                Takes a Response object as a parameter and returns None.
+        """
+        asyncio.get_event_loop().run_until_complete(
+            self.listen_async(request_object, handler)
+        )
+        # asyncio.get_event_loop().run_forever()
```
ledhed2222, line 114: did you mean to remove this?

xrpl/clients/websocket_client.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def close(self: WebsocketClient) -> None:
+        """Closes any open WebSocket connections."""
+        for task in asyncio.all_tasks():
```
ledhed2222, line 118: pretty sure this will cancel _all_ tasks regardless of whether they are relating to sockets. that's not right.

xrpl/clients/websocket_client.py
mvadari: Whoops, meant to do that, missed it

xrpl/clients/websocket_client.py
ledhed2222: you may need to store open requests related to the object, on the object to solve this

xrpl/clients/websocket_client.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Arguments:
+            request_object: An object representing information about a rippled request.
+
+        Returns:
+            The response from the server, as a Response object.
+        """
+        return asyncio.get_event_loop().run_until_complete(
+            self.request_async(request_object)
+        )
```
mvadari, line 85: Converts from async to synchronous

xrpl/clients/websocket_client.py
```diff
@@ -0,0 +1,150 @@
...
+        responses from the server.
+
+        Arguments:
+            request_object: An object representing information about a rippled request.
+            handler: The method that handles the Response objects from the server.
+                Takes a Response object as a parameter and returns None.
+        """
+        asyncio.get_event_loop().run_until_complete(
+            self.listen_async(request_object, handler)
+        )
```
ledhed2222, line 140: how does this work in practice? won't this just end up running forever?

xrpl/clients/websocket_client.py
```diff
@@ -0,0 +1,150 @@
...
+
+    async def close_async(self: WebsocketClient) -> None:
+        """Asynchronously closes any open WebSocket connections."""
+        while len(self.websockets) != 0:
+            websocket = self.websockets.pop()
+            await websocket.close()
```
ledhed2222, line 146: worried that this is waiting for each connection to close before requesting the next closure, which is a performance miss. also, confused by this approach since we should have a single socket connection, not multiple

pyproject.toml
```diff
@@ -46,6 +47,7 @@ darglint = "^1.5.8"
 nox-poetry = "^0.7.1"
 Sphinx = "^3.5.1"
 sphinx-rtd-theme = "^0.5.1"
+later = "^20.10.1"
```
ledhed2222, line 50: TODO not sure about this

pyproject.toml
```diff
@@ -25,11 +25,12 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.7.1"
```
ledhed2222, line 28: TODO not sure about this

tests/integration/test_websocket.py
```diff
@@ -0,0 +1,36 @@
+import asyncio
+from unittest import IsolatedAsyncioTestCase
+
+from tests.integration.it_utils import JSON_RPC_CLIENT, WEBSOCKET_CLIENT
```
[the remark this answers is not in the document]
ledhed2222: oh these tests don't work at all lol. let me remove them

mvadari (commented):
Looks good to me, but I'm not allowed to approve it since it's technically still my PR

ledhed2222 (approved):

PR #245, opened 2021-05-24
Async HTTP Integration Tests

mvadari:

This PR adds all the tests for the async HTTP client. It essentially copies the structure of the old tests, just making them async. I also made necessary refactors for now-flaky tests.

### Context of Change

Python 1.1 async stuff

### Type of Change

## Test Plan

CI passes.


tests/integration/it_utils.py
```diff
@@ -1,6 +1,15 @@
...
 from xrpl.asyncio.clients import AsyncJsonRpcClient, AsyncWebsocketClient
+from xrpl.asyncio.transaction import (
+    safe_sign_and_autofill_transaction as sign_and_autofill_async,
+)
+from xrpl.asyncio.transaction import (
+    safe_sign_and_submit_transaction as sign_and_submit_async,
+)
+from xrpl.asyncio.transaction import (
+    send_reliable_submission as send_reliable_submission_async,
+)
```
mvadari, line 12: The linter seems to prefer this way of doing this, instead of bundling them

tests/integration/reqs/test_account_info.py
```diff
@@ -1,15 +1,23 @@
-from unittest import TestCase
+from unittest import IsolatedAsyncioTestCase
```
ledhed2222, line 1: p sure this isn't available until 3.8, which is why i was playing with the `later` package that backports to 3.7.1. that would require us officially not supporting 3.7.0, but i think that's ok

tests/integration/reqs/test_account_objects.py
```diff
@@ -1,15 +1,23 @@
...
+class TestAccountObjects(IsolatedAsyncioTestCase):
+    def test_basic_functionality_sync(self):
         response = JSON_RPC_CLIENT.request(
             AccountObjects(
                 account=WALLET.classic_address,
             )
         )
         self.assertTrue(response.is_successful())
+
+    async def test_basic_functionality_async(self):
```
ledhed2222, line 17: i was trying to find a way to accomplish what you have here using decorators. i think it's worth it so that we don't have to duplicate code in the future. would you look into that please?

tests/integration/reqs/test_account_info.py
mvadari: Oh you're right, lemme see if I can get `later` to work

florent-uzio (commented):
That LGTM, happy to have another look once you look at Greg's comments.

ledhed2222 (approved):

PR #246, opened 2021-05-28
Adds Websocket Tests

mvadari:

This PR adds tests for the sync and async websocket network clients. Essentially, all the tests that are run on the JSON clients are also now run on the websocket clients. Whatever bugs were found were also fixed.

### Context of Change

Python websockets, v1.1

### Type of Change

## Test Plan

CI passes.


tests/integration/it_utils.py
```diff
@@ -12,18 +26,16 @@
 from xrpl.wallet import Wallet
 
 JSON_RPC_URL = "https://s.altnet.rippletest.net:51234"
+DEV_JSON_RPC_URL = "https://s.devnet.rippletest.net:51234"
```
florent-uzio, line 29: By curiosity, why do we use the Devnet ?

florent-uzio (approved):
Awesome work! I don't have any particular comment, just one question.

xrpl/models/requests/request.py
```diff
@@ -78,6 +84,66 @@ class Request(BaseModel):
...
+                    method == "submit"
+                    and cls.__name__ in ("SignAndSubmit", "SubmitOnly")
+                ):
+                    raise XRPLModelException(
+                        f"Using wrong constructor: using {cls.__name__} constructor "
+                        f"with Request method {method}."
+                    )
+                value = {**value}
+                del value["method"]
+            return cast(Request, super(Request, cls).from_dict(value))
```
ledhed2222, line 119: nit - you can remove an indentation level from this already-far-indented `else` block by removing the `else`, since you `return` in the `if`

PR #254, opened 2021-06-10
Fix SignerList serialization bug

mvadari:

This PR fixes another issue with serialization of SignerLists and multisigned transactions, which was introduced by #253.

### Context of Change

#253 introduced a new bug that was uncovered when running integ tests on master.

### Type of Change

## Test Plan

Added a test that failed when the bug was still in place.


xrpl/models/requests/__init__.py
```diff
@@ -1,4 +1,5 @@
 """Request models."""
+from xrpl.models.path import PathStep
```
ledhed2222, line 2: nice

ledhed2222 (approved):

PR #255, opened 2021-06-10
test: adds basic tests for subscribe

mayurbhandary:
Adds basic integration tests for stream subscriptions. 

### Context of Change
This test subscribes to the ledger stream and ensures that multiple(3) messages are received. The `test_async_and_sync` decorator was modified to accept a parameter to test using websockets clients only. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Before: 
No testing for subscriptions exists.

After:
Basic testing for subscriptions exists.


tests/integration/it_utils.py
```diff
@@ -89,12 +89,15 @@ def _choose_client_async(use_json_client: bool) -> Client:
...
-def test_async_and_sync(original_globals, modules=None, dev=False):
+def test_async_and_sync(
+    original_globals, modules=None, dev=False, websockets_only=False
+):
     def decorator(test_function):
         lines = _get_non_decorator_code(test_function)
         sync_code = (
             "".join(lines)
             .replace("async def", "def")  # convert method from async to sync
+            .replace("async for", "for")  # convert for from async to sync
```
ledhed2222, line 100: nice. while you're here you should also replace `async with` for just `with`

tests/integration/reqs/test_subscribe.py
```diff
@@ -0,0 +1,24 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+from tests.integration.it_utils import test_async_and_sync
+from xrpl.models.requests import StreamParameter, Subscribe
+from xrpl.models.requests.unsubscribe import Unsubscribe
+
+MESSAGE_LIMIT = 3
```
ledhed2222, line 6: add leading `_` to indicate that this is private

tests/integration/reqs/test_subscribe.py
```diff
@@ -0,0 +1,24 @@
...
+        await client.send(req)
+        count = 0
+        async for message in client:
+            if count > MESSAGE_LIMIT:
+                break
+            if count == 0:
+                self.assertIsInstance(message["result"]["ledger_index"], int)
+            else:
+                self.assertIsInstance(message["ledger_index"], int)
+            count = count + 1
```
ledhed2222, line 22: it's more idiomatic to do 
```python
count += 1
```

tests/integration/reqs/test_subscribe.py
```diff
@@ -0,0 +1,24 @@
...
+        req = Unsubscribe()
+        await client.send(req)
```
ledhed2222, line 24: just a style nit - since you don't need `req` this is more clean and clear to me

```python
await client.send(Unsubscribe())
```

tests/integration/reqs/test_subscribe.py
```diff
@@ -0,0 +1,24 @@
...
+from xrpl.models.requests.unsubscribe import Unsubscribe
+
+MESSAGE_LIMIT = 3
+
+
+class TestSubscribe(IntegrationTestCase):
+    @test_async_and_sync(globals(), websockets_only=True)
+    async def test_basic_functionality(self, client):
+        req = Subscribe(streams=[StreamParameter.LEDGER])
+        await client.send(req)
```
ledhed2222, line 13: style nit, since you don't need to reuse `req` this is cleaner and clearer to me

```python
await client.send(Subscribe(streams=[StreamParameter.LEDGER]))
```

tests/integration/reqs/test_subscribe.py
```diff
@@ -0,0 +1,24 @@
...
+        count = 0
+        async for message in client:
+            if count > MESSAGE_LIMIT:
+                break
+            if count == 0:
+                self.assertIsInstance(message["result"]["ledger_index"], int)
+            else:
+                self.assertIsInstance(message["ledger_index"], int)
```
ledhed2222, line 21: i wonder if we need stricter assertions here. probs if we're just going to have the one test

ledhed2222 (changes requested):
see comments

tests/integration/reqs/test_subscribe.py
ledhed2222: we should also test different subscription types, in separate tests within this class

tests/integration/reqs/test_subscribe.py
```diff
@@ -0,0 +1,24 @@
...
+
+class TestSubscribe(IntegrationTestCase):
+    @test_async_and_sync(globals(), websockets_only=True)
+    async def test_basic_functionality(self, client):
+        req = Subscribe(streams=[StreamParameter.LEDGER])
+        await client.send(req)
+        count = 0
+        async for message in client:
+            if count > MESSAGE_LIMIT:
+                break
```
ledhed2222, line 17: minor, but you are waiting for an extra message by having this condition here. instead, do this at the end of each iteration

```python
if count == MESSAGE_LIMIIT:
  break
count += `
```

tests/integration/reqs/test_subscribe.py
ledhed2222: actually - we need to test that you can unsubscribe from this specific stream, and should test that we get an unsubscribe confirmation

tests/integration/reqs/test_subscribe.py
mayurbhandary: Would you like me to make it so that each ledger index is strictly 1 greater than the last?

tests/integration/reqs/test_subscribe.py
mayurbhandary: I've added tests for validations, consensus, transactions, and transactions_proposed in my latest. I did not implement tests for the manifests, peer_status, and server streams. 

The manifests stream is slow because manifests don't come around all that often. The peer_status stream requires an admin connection. The server stream is also slow since the status of the server doesn't change that frequently. Let me know if the stream that I implemented are enough.

ledhed2222 (approved):

tests/integration/reqs/test_subscribe.py
```diff
@@ -0,0 +1,94 @@
...
+        async for message in client:
+            if count == 0:
+                self.assertIsInstance(message["result"]["ledger_index"], int)
+            else:
+                self.assertIsInstance(message["ledger_index"], int)
+            if count == _MESSAGE_LIMIT:
+                break
+            count += 1
+
+        await client.send(Unsubscribe(streams=[StreamParameter.LEDGER]))
```
ledhed2222, line 23: you could use `await client.request(Unsubscribe(streams=[StreamParameter.LEDGER]))` to avoid another `for` loop...i believe

ledhed2222: if i'm right - let's replace the unsubscribe parts with this pattern since it's more terse

mayurbhandary: I still need a loop to check if the messages contain a successful unsubscription response.

PR #257, opened 2021-06-11
Exports all models at the top models level

mvadari:

Title says it all. Makes it easier to import models.

### Context of Change

v1.1

### Type of Change

## Test Plan

CI passes. I also checked the docs and made sure they didn't go wonky.

### DIFF
--- a/xrpl/models/__init__.py
+++ b/xrpl/models/__init__.py
@@ -1,14 +1,22 @@
 """Top-level exports for the models package."""
 from xrpl.models import amounts, currencies, requests, transactions
+from xrpl.models.amounts import *  # noqa: F401, F403
+from xrpl.models.currencies import *  # noqa: F401, F403
 from xrpl.models.exceptions import XRPLModelException
 from xrpl.models.path import Path, PathStep
+from xrpl.models.requests import *  # noqa: F401, F403
+from xrpl.models.transactions import *  # noqa: F401, F403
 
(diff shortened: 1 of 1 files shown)


xrpl/models/__init__.py
```diff
@@ -1,14 +1,22 @@
 """Top-level exports for the models package."""
 from xrpl.models import amounts, currencies, requests, transactions
+from xrpl.models.amounts import *  # noqa: F401, F403
+from xrpl.models.currencies import *  # noqa: F401, F403
```
ledhed2222, line 4: you don't need to do this. once you have `amounts` you can access `amounts.__all__`. as it is, this import doesn't do anything AFAICT

mvadari: That doesn't work (I just tried it to make sure) - it still needs to be available in the `__init__` file as an imported thing in order for it to be valid in its `__all__`.

ledhed2222: I tried in the console myself. Can you show what you tried? Maybe I'm missing something.

mvadari: I commented out one of the `import *` lines and then tried to import something from that directory from `xrpl.models`, and it had an `ImportError`

ledhed2222: oh i see why

ledhed2222 (approved):

PR #258, opened 2021-06-11
fix: UInt64.from_value() accepts lower case hex

natenichols:
Allow for `UInt64.from_value()` to accept lower case hex w/ length 1-16.

### Context of Change
This PR allows for parsing lower case hex strings w/ length 1-16. It does this w/ an `rjust`, which pads the string `value` w/ leading zeros until it is length 16. This allows for it to be encoded like normal. 

Also, the `HEX_REGEX` was adjusted to allow for lower case hex strings.

### Type of Change
- [ ] Documentation Updates
- [ ] Release

## Test Plan
Added a test to verify that encoding would work on both hex-strings w/ lowercase letters and hex-strings w/ length <16.

### DIFF
--- a/tests/unit/core/binarycodec/test_main.py
+++ b/tests/unit/core/binarycodec/test_main.py
@@ -240,8 +240,8 @@ class TestMainSimple(TestCase):
             },
             "LowNode": "0",
             "PreviousTxnID": (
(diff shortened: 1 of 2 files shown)


mvadari:
lgtm, but looks like you're failing some tests

mvadari (approved):

ledhed2222:
Hell yeah

PR #262, opened 2021-06-14
Fixes integration test issues

mvadari:

This PR fixes issues that Github Actions had with installing the `aiounittest` module in python3.7, which is used for testing. It also fixes some other errors that were found along the way.

### Context of Change

@ledhed2222 ran integration tests on the latest commit on master last week and discovered that the tests were failing. We need to pass these tests before we can release v1.1.

### Type of Changes

## Test Plan

Integration test CI passes.


xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -38,7 +38,7 @@ async def request_impl(self: JsonRpcBase, request: Request) -> Response:
 
         :meta private:
         """
-        async with AsyncClient() as http_client:
+        async with AsyncClient(timeout=10.0) as http_client:
```
mvadari, line 41: default timeout is 5.0s, some integration tests were failing intermittently because they timed out

ledhed2222 (approved):

PR #267, opened 2021-06-24
Fix install command

intelliot:

I was following https://xrpl-py.readthedocs.io/en/stable/ and the install command didn't work:

```
➜  pip3 install xrpl
ERROR: Could not find a version that satisfies the requirement xrpl (from versions: none)
ERROR: No matching distribution found for xrpl
```

This PR fixes that.

### Type of Change


mvadari (approved):
Thanks!

PR #269, opened 2021-07-01
improve currency fix

ledhed2222:

Improve fix from #266 based on learnings!

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/xrpl/constants.py
+++ b/xrpl/constants.py
@@ -1,6 +1,7 @@
 """Collection of public constants for XRPL."""
 from enum import Enum
-from re import Pattern, compile
+from re import compile
+from typing import Pattern
 
 from typing_extensions import Final
 
--- a/xrpl/constants.py
+++ b/xrpl/constants.py
@@ -1,5 +1,8 @@
 """Collection of public constants for XRPL."""
+import re
 from enum import Enum
+
+from typing_extensions import Final
 
 
 class CryptoAlgorithm(str, Enum):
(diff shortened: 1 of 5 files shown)


xrpl/constants.py
```diff
@@ -13,3 +16,7 @@ class XRPLException(Exception):
     """Base Exception for XRPL library."""
 
     pass
+
+
+ISO_CURRENCY_REGEX: Final[Pattern[str]] = compile(f"^[A-Z0-9]{3}$")
+HEX_CURRENCY_REGEX: Final[Pattern[str]] = compile(f"^[A-F0-9]{40}$")
```
mvadari, line 22: This will show up in the docs under the "Global Variables" tab - do we want to add docs or remove this from the docs?

xrpl/core/binarycodec/types/currency.py
```diff
@@ -85,7 +83,7 @@ def __init__(self: Currency, buffer: Optional[bytes] = None) -> None:
         code_bytes = self.buffer[12:15]
         # Determine whether this currency code is in standard or nonstandard format:
         # https://xrpl.org/currency-formats.html#nonstandard-currency-codes
-        if self.buffer[:2] != bytes(2):
+        if self.buffer[:1] != bytes(1):
```
mvadari, line 86:
```suggestion
        if self.buffer[0] != 0:
```

mvadari:
Looks like this is also not passing 3.9 tests

mvadari (approved):

PR #271, opened 2021-07-05
Change assert statement when not for invariant in mypy check, include development node recommendation

### LINKED ISSUE #270: Async client takes a while to close the connection
For example if I run the following script with the `websockets` package within a context manager managing the session, it closes quickly. 

https://gist.github.com/yyolk/42b67ec5f1b1a17e273b5497e9eaad88#file-01__websockets__xrpl_stat_table-py

If I run the same code, with `xrpl-py` which is set up to handle the websocket identically, the websocket does close but it takes a while. 

https://gist.github.com/yyolk/42b67ec5f1b1a17e273b5497e9eaad88#file-02__xrpl-py__xrpl_stat_table-py

My first instinct was to see where there was a difference in how `websockets` handled it and how `xrpl-py` does. For the most part it's identical, following the dunder method for `__aexit__` to `close()` when called from the `AsyncWebsocketClient.__aexit__` https://github.com/XRPLF/xrpl-py/blob/e5bbdf458ad83e6670a4ebf3df63e17fed8b099f/xrpl/asyncio/clients/async_websocket_client.py#L58-L65

It led me to notice a few assert statements in the code that seemed pretty reasonable for conditional checks
https://github.com/XRPLF/xrpl-py/blob/e5bbdf458ad83e6670a4ebf3df63e17fed8b099f/xrpl/asyncio/clients/websocket_base.py#L84-L90

However if I run with `python -O` or in optimized mode, which will drop all `assert` statements, I do see the socket program close as swiftly as the one with `websockets` which I've seen very little difference in it's implementation. The only thing that stuck out to me at first glance was using the convenience of [`asyncio.shield(...)`](https://github.com/aaugustin/websockets/blob/b2a95c45fae19fff0a3473158dd02afe2ca42604/src/websockets/legacy/protocol.py#L671) over [spinning up a background thread](https://github.com/XRPLF/xrpl-py/blob/e5bbdf458ad83e6670a4ebf3df63e17fed8b099f/xrpl/clients/websocket_client.py#L92-L113) and these `assert` statements.

I did a quick search and saw we only had `asserts` in a few places and tried dropping it here: https://github.com/yyolk/xrpl-py/tree/patch-20210705

`-O` command line arg: https://docs.python.org/3/using/cmdline.html#cmdoption-o

yyolk:

Closes #270

### Context of Change
See #270.

When running [this script](https://gist.github.com/yyolk/42b67ec5f1b1a17e273b5497e9eaad88#file-02__xrpl-py__xrpl_stat_table-py):

closing (with <kbd>^C</kbd>, a `KeyboardInterrupt` closes the socket as quickly as the implementation using `websockets` when running `python -O ...` but not as quick when running `python ...`. 

This may result in sending another `KeyboardInterrupt` which isn't what you want to do.

### Type of Change

### Before
- An assert statement that is used for raising when an import doesn't have a requirement
- no runtime recommendations

### After
- An `assert` in a `__init__` was also dropped for `ImportError` as it is an error as a result of an `import` :D
- runtime recommendation to include `-O` in production or in development when running locally with CPython, `assert` statements are dropped with [`-O`](https://docs.python.org/3/using/cmdline.html#cmdoption-o)

## Test Plan

- use the test suite as-is for the ImportError change
- feedback on relevant documentation changes.

### DIFF
--- a/xrpl/core/keypairs/__init__.py
+++ b/xrpl/core/keypairs/__init__.py
@@ -13,10 +13,11 @@ from xrpl.core.keypairs.main import (
     sign,
 )
 
-assert (
-    "ripemd160" in algorithms_available
-), """Your OpenSSL implementation does not include the RIPEMD160 algorithm,
-    which is required by XRPL"""
+if "ripemd160" not in algorithms_available:
+    raise ImportError(
+        "Your OpenSSL implementation does not include the RIPEMD160 algorithm,"
+        " which is required by XRPL."
+    )
 
 __all__ = [
     "derive_classic_address",


yyolk:
first glance looks like dropping some `# mypy` statements caused the type checks to fail; looking into it

ledhed2222:
Yes @yyolk - those assert statements annotated with mypy are for the type checker. They can't be removed. Changing the assert into an import error should be ok I suppose.

yyolk:
@ledhed2222 gotcha. I suppose the reason is to pass mypy type checks for when there's an invariants only and `-O` should be used in production, I'm ok with that.

There's quite a bit a slowdown when running without `-O`, at least in the slow-down I experience which could be unrelated, but only happens without the changes in https://github.com/yyolk/xrpl-py/tree/309a915e57ea6d675f8711ffbfe93409786d1562 or for complete clarity when you don't run without `-O` to illustrate, the `assert` statements are completely dropped:

```sh
$ cat script.py
assert False

$ python -O script.py
$ echo $?
0
$ python script.py
  File "script.py", line 1, in <module>
    assert False
AssertionError
$ echo $?
1
```

I'll go ahead and revert the drop `assert` statements that I now understand are for `mypy`, (which were commented as such before. I didn't know about 

image via https://mypy.readthedocs.io/en/stable/cheat_sheet.html?highlight=assert#built-in-types

image via https://mypy.readthedocs.io/en/stable/common_issues.html#unexpected-errors-about-none-and-or-optional-types


I think to close the issue this references, I could also include an update the appropriate docs to recommend running with `-O` at runtime when using this library, at least using CPython.

ledhed2222:
Love the idea of adding documentation on this! Would you please add something? That would make this PR more complete IMO. 

Overall, xrpl-py isn't optimized for performance. It's trying to optimize usability and access to features. In my opinion, the developer that needs to optimize xrpl performance is sophisticated enough and has reason enough to run their own xrpl node.

yyolk:
A quick update on my dive into the referenced issue, it's been a while; I apologize for the delay. 

I'm sometimes seeing this to work when passing the `-O` flag, it also seems to happen sometimes even with that flag, which I'd rather dive into more thoroughly with a debugger before recommending just the `-O` to fix the issue I'm experiencing.

I don't think I'm going to include the option for `-O` since I'm not always seeing it work and would like to dive in further before seeing why it sometimes *does* work.

I appreciate you mentioning library's goal to optimize usability and access to features; I definitely agree with this and also take this into account with recommendations.

To your other point; I see what you're saying re: running a node for development. Would be happy to add this to the doc!

This would certainly alleviate my concern of terminating the slow thread taking care of the cleanup of the websocket (which could be to public node). It could be to my development node. I wouldn't so much killing the thread to close the websocket properly if I'm being impatient. I may also suggest running that development mode on the testnet.

yyolk:
Closing this :cold_sweat:    sorry for the cruft! :bow:

PR #272, opened 2021-07-08
update docs for clients

ledhed2222:

Improves docs on websocket clients and uses OO for common client initializes

### Type of Change

- [x] Documentation Updates
- [ ] Release

### DIFF
--- a/xrpl/core/binarycodec/types/serialized_type.py
+++ b/xrpl/core/binarycodec/types/serialized_type.py
@@ -14,38 +14,21 @@ class SerializedType(ABC):
 
     @classmethod
     @abstractmethod
-    def from_parser(
+    def from_parser(  # noqa: D102
         cls: Type[SerializedType],
         # TODO: Resolve Any (can't be ``BinaryParser`` because of circular imports)
         parser: Any,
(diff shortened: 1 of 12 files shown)


xrpl/core/binarycodec/types/serialized_type.py
```diff
@@ -14,38 +14,21 @@ def __init__(self: SerializedType, buffer: bytes = bytes()) -> None:
...
-
-        Args:
-            parser: The parser to construct a SerializedType from.
-            length_hint: The number of bytes to consume from the parser.
-
-        Raises:
-            NotImplementedError: Always.
-        """
-        raise NotImplementedError("SerializedType.from_parser not implemented.")
+        pass
```
mvadari, line 24: Why the switch from `NotImplementedError` to `pass`?

ledhed2222: it's not necessary. since it's an `abstractmethod` it will raise if you try to instantiate anything that doesn't have it implemented. so i wanted to make all our `abstractmethod` stubs the same. some were using `pass` and others were raising

mvadari (approved):
lgtm

PR #273, opened 2021-07-12
Workaround for ripemd160 not being available on Google Colab

hammertoe:

Use PyCrypto for RIPEMD160 hash if not found in hashlib

### Context of Change

`hashlib` relies upon OpenSSL to provide RIPEMD160. It can occur that OpenSSL is not configured to have RIPEMD160, in particular it is not found in Google Colab. I had a good play around with installing/re-installing OpenSSL and various dependancies on Google Colab, but could not get it to find RIPEMD160. There is an implementation of it in `pycrypto`, and so if we can't find it in hashlib, then try looking in there instead.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Moved the check from `__init__.py` into `helpers.py`, define a lambda to abstract the call.

## Test Plan

Manual test and it seems to work ok.

### DIFF
--- a/xrpl/core/keypairs/__init__.py
+++ b/xrpl/core/keypairs/__init__.py
@@ -2,7 +2,6 @@
 Low-level functions for creating and using cryptographic keys with the XRP
 Ledger.
 """
-
 from xrpl.core.keypairs.exceptions import XRPLKeypairsException
 from xrpl.core.keypairs.main import (
     derive_classic_address,
@@ -11,7 +10,6 @@ from xrpl.core.keypairs.main import (
     is_valid_message,
     sign,
 )
-
 
 __all__ = [
     "derive_classic_address",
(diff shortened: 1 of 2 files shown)


ledhed2222:
so - i think this is a great idea, and yes RIPEMD support seems to be an issue on all 3.7.x versions. i'd love to work around this with another library like you've done! if we do this, i think we should commit to it for all versions and try either the added library or the `hashlib` version. if you do that, you'll need to add a runtime dependency to this library by running `poetry add <library>` to add to dependencies.

ledhed2222 (changes requested):
see comments :)

connorjchen:
Updated version of PR: https://github.com/XRPLF/xrpl-py/pull/490

connorjchen (commented):
^

PR #285, opened 2021-10-11
feat: ensure support for python 3.10

mvadari:

This PR ensures that xrpl-py supports Python 3.10.

### Context of Change

Python 3.10 just came out: https://docs.python.org/3/whatsnew/3.10.html

### Type of Change

## Test Plan

CI passes. I [ran integration tests separately](https://github.com/XRPLF/xrpl-py/actions/runs/1330735869), they also pass.

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -1,6 +1,10 @@
 name: Unit test
 
-on: [push, pull_request, workflow_dispatch]
+on:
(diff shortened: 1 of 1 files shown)


mvadari:
There's a bug in asyncio 3.10 that makes the integration tests fail: https://simonwillison.net/2021/Oct/9/finding-and-reporting-a-bug/

PR #293, opened 2021-11-01
fix: make Python 3.10 support more general

mvadari:

This PR allows users of xrpl-py who aren't using Python 3.10 to continue to use websockets v9.0 instead of needing to upgrade to v10.0.

### Context of Change

People shouldn't have to upgrade websockets if they don't need to just to use xrpl-py.

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,10 @@ base58 = "^2.1.0"
 ECPy = "^1.2.5"
 typing-extensions = "^3.7.4"
 httpx = "^0.18.1"
-websockets = "^10.0"
+websockets = [
+    {version = ">=9.0.1 <10.0", python = ">= 3.7, < 3.9"},
+    {version = "^10.0", python = "^3.10"}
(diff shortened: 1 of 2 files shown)


JST5000 (approved):

pyproject.toml
```diff
@@ -33,7 +33,10 @@ base58 = "^2.1.0"
 ECPy = "^1.2.5"
 typing-extensions = "^3.7.4"
 httpx = "^0.18.1"
-websockets = "^10.0"
+websockets = [
+    {version = ">=9.0.1 <11.0", python = ">= 3.7, < 3.10"},
+    {version = "^10.0", python = "^3.10"}
```
ledhed2222, line 38: this shoudl be for `python = ">= 3.10"` so we don't have to keep updating this

ledhed2222 (approved):

intelliot (approved):

PR #295, opened 2021-11-01
chore: remove CODEOWNERS

mvadari:

This PR removes codeowners.

### Context of Change

@ledhed2222 and I are no longer the owners of this repo

### Type of Change

- [x] Repo management

## Test Plan

Code isn't touched.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,0 @@
-* @mvadari @ledhed2222
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,10 +5,10 @@ description = "A complete Python library for interacting with the XRP ledger"
 readme = "README.md"
 repository = "https://github.com/XRPLF/xrpl-py"
 authors = [
-  "Amie Corso <acorso@ripple.com>",
-  "Ted Kalaw <tkalaw@ripple.com>",
   "Mayukha Vadari <mvadari@ripple.com>",
   "Greg Weisbrod <gweisbrod@ripple.com>",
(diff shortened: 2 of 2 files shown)


CODEOWNERS
```diff
@@ -1 +0,0 @@
-* @mvadari @ledhed2222
```
ledhed2222, line 1: shouldn't we replace with new owners? if none defined, doesn't seem bad to me to keep this until there are new ones defined

mvadari: We don't have any codeowners in the xrpl.js repo and it's fine - IMO this isn't really necessary for these repos.

mvadari: Unless you think it'll be actually useful given that different groups are going to be independently working in these repos? Either way, the decision should be the same across the two repos we control(led).

ledhed2222 (approved):

JST5000 (approved):
I think it makes sense to not have centralized code owners - Just getting a review from whoever committed that part of the code last/the maintainers makes sense to me.

mvadari:
> I think it makes sense to not have centralized code owners - Just getting a review from whoever committed that part of the code last/the maintainers makes sense to me.

The objective of codeowners is to represent those maintainers - basically, when this was in place, you had to get an approval from @ledhed2222 or I in order to merge your PR.

JST5000:
> The objective of codeowners is to represent those maintainers - basically, when this was in place, you had to get an approval from @ledhed2222 or I in order to merge your PR.

I see, if it becomes a problem, we'll re-implement it, but I don't think it's necessary for now :)

PR #296, opened 2021-11-02
v1.2.0

mvadari:

Bumps version and updates CHANGELOG for v1.1.2. Also adds mentions of the mailing list.

### Type of Change

- [x] Documentation
- [x] Release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [[Unreleased]]
 
-## [1.1.2] - 2021-07-02
+## [1.1.2] - 2021-11-02
 ### Added
 - Support for Python 3.10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [[Unreleased]]
 
-## [1.1.2] - 2021-11-02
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,9 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [[Unreleased]]
+
+## [1.1.2] - 2021-11-02
+### Added
+- Support for Python 3.10
```
ledhed2222, line 12: this should be 1.2.0 because we're adding something

mvadari: It still feels like a patch because no code was updated

ledhed2222:
in this case it's still not IMO because we are adding support for a new
version of python. if someone is using 3.10 they will need to pin to our
version >= 1.2.0 and it would be strange to pin to a patch version for that.

JST5000 (approved):

CHANGELOG.md
```diff
@@ -6,9 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [[Unreleased]]
+
+## [1.1.2] - 2021-11-02
```
intelliot, line 10:
```suggestion
## [1.2.0] - 2021-11-02
```

intelliot (approved):

PR #300, opened 2021-11-09
fix: make `response.result` type more specific

mvadari:

In `Response`, the result type was initially either a `List` or a `Dict`. But it turns out that there aren't actually any `rippled` responses where the result type is a `List`. So this PR removes that as an option in the type, which eliminates a lot of casting.

### Context of Change

This was discovered while working on the Sidechain Launch Kit.

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/xrpl/asyncio/account/main.py
+++ b/xrpl/asyncio/account/main.py
@@ -1,6 +1,6 @@
 """High-level methods to obtain information about accounts."""
 
-from typing import Any, Dict, Union, cast
+from typing import Dict, Union, cast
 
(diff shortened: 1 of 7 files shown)


intelliot (approved):
This sounds right. Just to be safe, it might be nice to confirm with 1 person on the Core Ledger team that rippled always returns a JSON object and never an array -- perhaps @cjcobb23 or @manojsdoshi could either confirm, or point us to someone who can?

This doesn't necessarily need to hold up merging though

JST5000 (approved):

ledhed2222 (approved):

PR #304, opened 2021-11-12
feat: add optional `ledger_index` param to account methods

mvadari:

This PR adds an optional parameter to the main account methods for the ledger index to query. It also sets the default ledger index for querying the sequence number to `"current"` instead of `"validated"`.

### Context of Change

https://github.com/XRPLF/xrpl.js/pull/1719

### Type of Change

## Test Plan

CI passes. Tests pass.


JST5000 (approved):
LGTM

mDuo13:
Nitpick: can we write "The default is ___" instead of "Defaults to ___"? Ripple's style guide prefers "The default is..." because (we think) it's easier to read and understand correctly even for non-native speakers.

wojake (approved):

ledhed2222 (approved):

PR #308, opened 2021-11-15
fix: remove erroring in `send_reliable_submission` if not success

### LINKED ISSUE #291: Transaction Hash for failed. , queued transactions (DGE-68)
send_reliable_submission method raises exception in cases where a transaction hash should have generated like terQUEUED. Thus, unable to get transaction hash, which is essential to query that transaction status later.

mvadari:

This PR removes erroring in `send_reliable_submission` if the response code is something other than `tesSUCCESS`. It also adds an error if there is no `last_ledger_sequence` in the transaction.

### Context of Change

Fixes #291. Also https://github.com/XRPLF/xrpl.js/pull/1704#pullrequestreview-773442168

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added a test for if the `last_ledger_sequence` is missing in a transaction.

### DIFF
--- a/xrpl/asyncio/transaction/reliable_submission.py
+++ b/xrpl/asyncio/transaction/reliable_submission.py
@@ -77,13 +77,6 @@ async def send_reliable_submission(
             `last_ledger_sequence` param.
     """
     transaction_hash = transaction.get_hash()
-    submit_response = await submit_transaction(transaction, client)
-    result = submit_response.result
-    if result["engine_result"] != "tesSUCCESS":
-        result_code = result["engine_result"]
-        result_message = result["engine_result_message"]
-        raise XRPLReliableSubmissionException(
-            f"Transaction failed, {result_code}: {result_message}"
-        )
+    await submit_transaction(transaction, client)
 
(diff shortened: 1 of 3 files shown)


xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -76,14 +76,11 @@ async def send_reliable_submission(
         XRPLReliableSubmissionException: if the transaction fails or is missing a
```
JST5000, line 76: This comment need to be updated since this no longer throws if the transaction fails.

JST5000: (Or does _wait_for_final_transaction_outcome throw if the transaction fails now?)

JST5000 (approved):
LGTM

xrpl/asyncio/transaction/reliable_submission.py
mvadari: "fails" also refers to the last ledger sequence passing.

khancode (approved):
LGTM

PR #309, opened 2021-11-16
feat: add client method `request_json`

mvadari:

This PR adds a new method to the clients, `request_json`. This allows devs to send a dictionary (that is properly formatted) directly to rippled, instead of needing to use a request model. 

This is useful for features where the models do not exist yet (such as NFT features or sidechain features) or for admin features (which don't have models right now).

### Context of Change

The clients need the ability to send raw JSON (in the form of dictionaries) to nodes. This was necessary for sidechain work and for docs work.

### Type of Change

## Test Plan

CI passes. Added tests to test `request_json`. Integration tests pass.

### DIFF
--- a/xrpl/asyncio/clients/json_rpc_base.py
+++ b/xrpl/asyncio/clients/json_rpc_base.py
@@ -26,7 +26,7 @@ class JsonRpcBase(Client):
         Base ``request_impl`` implementation for JSON RPC.
 
         Arguments:
-            request: An object representing information about a rippled request.
+            request: A Request object representing information about a rippled request.
 
         Returns:
             The response from the server, as a Response object.
@@ -43,10 +43,10 @@ class JsonRpcBase(Client):
         Base ``request_json_impl`` implementation for JSON RPC.
 
         Arguments:
-            request: An object representing information about a rippled request.
+            request: A dictionary representing information about a rippled request.
 
         Returns:
-            The response from the server, as a Response object.
+            The response from the server, as a dictionary.
 
         :meta private:
         """
--- a/xrpl/asyncio/clients/json_rpc_base.py
+++ b/xrpl/asyncio/clients/json_rpc_base.py
@@ -1,6 +1,8 @@
 """A common interface for JsonRpc requests."""
 from __future__ import annotations
 
+from typing import Any, Dict, cast
+
 from httpx import AsyncClient
 from typing_extensions import Final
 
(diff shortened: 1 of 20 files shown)


xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -24,16 +26,33 @@ async def request_impl(self: JsonRpcBase, request: Request) -> Response:
         Base ``request_impl`` implementation for JSON RPC.
 
         Arguments:
-            request: An object representing information about a rippled request.
+            request: A Request object representing information about a rippled request.
 
         Returns:
             The response from the server, as a Response object.
 
+        :meta private:
```
JST5000, line 34: What does`:meta private:` mean?

xrpl/asyncio/clients/websocket_base.py
```diff
@@ -18,19 +18,18 @@
...
+    if "id" in request and request["id"] is not None:
         return request
-    request_dict = request.to_dict()
-    request_dict["id"] = f"{request.method}_{randrange(_REQ_ID_MAX)}"
-    resp = Request.from_dict(request_dict)
-    assert resp.id is not None  # mypy
-    return resp
+    request_dict = cast(Dict[str, Any], json.loads(json.dumps(request)))  # deepcopy
+    command = request["command"]
+    request_dict["id"] = f"{command}_{randrange(_REQ_ID_MAX)}"
```
JST5000, line 31: Could this ever fail by chance? (i.e. rolling the same id as a previous transaction?)

mvadari: Possibly, but the chance is very rare.

JST5000: I think that issue would be really hard to debug though (and the odds would increase as more requests were sent) Maybe something like this would be better? 

```
...
_REQ_ID_MAX = 1000000
last_id = 0
...
last_id = (1 + last_id) % _REQ_ID_MAX
request_dict["id"] = f"{command}_{last_id}"
```

Although I think that would mess up in a multi-thread situation unless locks were added, so maybe just doing random is better.

mvadari: Happy to chat more about that in a separate issue, but this is out of scope for this PR.

JST5000 (approved):
LGTM

mvadari:
Closed in favor of #313

PR #313, opened 2021-11-22
feat: add `GenericRequest` model for unsupported requests

mvadari:

This PR adds a new request model, `UnknownRequest`. The objective of this request model is to handle all unsupported requests, such as admin requests and new proposed features that aren't on the ledger yet. It can convert from both WS formatting and JSON formatting through the `from_dict` class method.

### Context of Change

xrpl-py needs to somehow support request types that don't have models, such as for sidechains or new features (like hooks).

### Type of Change

## Test Plan

CI passes. Added tests for the constructor both JSON formatting and WS formatting.


intelliot:
maybe GenericRequest or CustomRequest might be a better name?

xrpl/models/requests/request.py
```diff
@@ -74,6 +74,9 @@ class RequestMethod(str, Enum):
     PING = "ping"
     RANDOM = "random"
 
+    # generic unknown/unsupported request
+    UNKNOWN_REQUEST = "unknown_request"
```
ledhed2222, line 78: can we add to this comment to say that it has no analogue on xrpl, IE it's only an xrpl-py thing?

xrpl/models/requests/unknown_request.py
```diff
@@ -0,0 +1,82 @@
...
+
+from xrpl.models.exceptions import XRPLModelException
+from xrpl.models.requests.request import Request, RequestMethod
+from xrpl.models.utils import require_kwargs_on_init
+
+
+@require_kwargs_on_init
+@dataclass(init=False, frozen=True)
+class UnknownRequest(Request):
+    """A request object representing all unsupported requests."""
```
ledhed2222, line 15: can we add something here explaining that this has no analogue on xrpl?

xrpl/models/requests/unknown_request.py
```diff
@@ -0,0 +1,82 @@
...
+            kwargs: All the arguments for the request.
+        """
+        # initialize all the dataclass stuff
+        super().__init__(
+            id=(cast(Union[str, int, None], kwargs["id"]) if "id" in kwargs else None),
+            method=RequestMethod.UNKNOWN_REQUEST,
+        )
+        # pass in all the kwargs into the object (so self.key == value)
+        for key, value in kwargs.items():
+            object.__setattr__(self, key, value)
```
ledhed2222, line 38: why does this work? is it actually frozen?

xrpl/models/requests/unknown_request.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self: UnknownRequest, **kwargs: Dict[str, Any]) -> None:
+        """
+        Initializes an UnknownRequest.
+
+        Arguments:
+            kwargs: All the arguments for the request.
+        """
+        # initialize all the dataclass stuff
+        super().__init__(
+            id=(cast(Union[str, int, None], kwargs["id"]) if "id" in kwargs else None),
```
ledhed2222, line 33: you could take the `None` out of the Union

xrpl/models/requests/unknown_request.py
mvadari: The way that freezing works in dataclasses is that they've overwritten the class's `__setattr__` method and had it error. This gets around that by using the object `__setattr__` method.

xrpl/models/requests/unknown_request.py
mvadari: I left it in in case someone specifically has `id=None` (which is theoretically possible).

mDuo13 (approved):
Aside from the name `UnknownRequest` the rest seems fine to me. Tested it out and was able to check amendment votes for example

mDuo13 (approved):

intelliot (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `ledger_index` optional param for all the main account methods
 - `TicketCreate` transaction model
+- `UnknownRequest` model for unsupported request types
```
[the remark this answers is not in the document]
ledhed2222: i don't love `CustomRequest` because this is _also_ being used by `Request.from_dict` if the request type cannot be determined. in that case, it's not really a custom request, it's more of an unknown request.

feels like we can have a name that captures both. `GenericRequest`?

tests/integration/it_utils.py
```diff
@@ -154,14 +154,14 @@ async def _run_async_test(self, client):
...
         def modified_test(self):
             if not websockets_only:
-                with self.subTest(version="sync", client="json"):
-                    _run_sync_test(self, JSON_RPC_CLIENT)
                 with self.subTest(version="async", client="json"):
                     asyncio.run(_run_async_test(self, ASYNC_JSON_RPC_CLIENT))
-            with self.subTest(version="sync", client="websocket"):
-                _run_sync_test(self, WEBSOCKET_CLIENT)
+                with self.subTest(version="sync", client="json"):
+                    _run_sync_test(self, JSON_RPC_CLIENT)
```
ledhed2222, line 160: why was thsi moved? just an accident?

xrpl/models/requests/custom_request.py
```diff
@@ -0,0 +1,86 @@
...
+    A request object representing all unsupported requests.
+
+    There is no analog in rippled - this is an xrpl-py-specific model.
+    """
+
+    method: RequestMethod = field(default=RequestMethod.CUSTOM_REQUEST, init=False)
+    """
+    This field is required.
+
+    :meta hide-value:
```
ledhed2222, line 25: doesn't this have to be `init=True`?

CHANGELOG.md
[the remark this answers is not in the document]
mvadari: I can do `GenericRequest`

xrpl/models/requests/custom_request.py
mvadari: No - the init of this method doesn't need it. It's only called in the super init, and `Request` does take a method param in its init.

xrpl/models/requests/request.py
```diff
@@ -74,6 +74,10 @@ class RequestMethod(str, Enum):
     PING = "ping"
     RANDOM = "random"
 
+    # generic unknown/unsupported request
+    # (there is no XRPL analog, this model is specific to xrpl-py)
+    CUSTOM_REQUEST = "custom_request"
```
ledhed2222, line 79: wondering if we should define this const at all, since using it would always be an error. perhaps the new request should just have an empty, optional field there

xrpl/models/requests/request.py
```diff
@@ -160,5 +169,5 @@ def to_dict(self: Request) -> Dict[str, Any]:
             The dictionary representation of a Request.
         """
         # we need to override this because method is using ``field``
-        # which will not include the value in the objects __dict__
+        # which will not include the value in the object's __dict__
```
ledhed2222, line 172: 🙏

xrpl/models/requests/request.py
mvadari: That'll mess up the rest of our typing - the `method` param requires a `TransactionType` enum type.

xrpl/models/requests/request.py
```diff
@@ -150,7 +159,7 @@ def get_method(cls: Type[Request], method: str) -> Type[Request]:
         if method in request_methods:
             return request_methods[method]
 
-        raise XRPLModelException(f"{method} is not a valid Request method")
+        return request_models.CustomRequest
```
ledhed2222, line 162: please update teh doc string. this method no longer raises and it will return `CustomRequest` if no suitable model is found.

xrpl/models/requests/request.py
ledhed2222: in that case, let's prepend the value with `zz`, to further show it's nto a real request and to not have any collision with any future actual request

tests/integration/it_utils.py
mvadari: It makes debugging easier - all async stuff is run before the sync stuff, and async stuff has a better stack trace.

xrpl/models/requests/request.py
mvadari: I think the phrase "request" in there does that. But I can do that.

tests/integration/reqs/test_custom_request.py
```diff
@@ -0,0 +1,41 @@
...
+    async def test_websocket_formatting(self, client):
+        response = await client.request(
+            GenericRequest.from_dict(
+                {
+                    "command": "tx_history",
+                    "start": 0,
+                }
+            )
+        )
+        self.assertTrue(response.is_successful())
```
ledhed2222, line 41: so this means you can submit a JSON-formatted request via websockets and vice-versa? i think that's cool actually just want to be sure i'm right on that

ledhed2222 (approved):

tests/integration/reqs/test_custom_request.py
mvadari: Yes. Seemed more complicated than it's worth to do something differently.

PR #315, opened 2021-11-22
test: use rippled standalone Docker for most tests

mvadari:

This PR rewrites the testing framework so that most of the tests run through the rippled standalone Docker container (used for xrpl.js testing). This makes the integration tests take about 5 minutes instead of about 12. 

The subscribe and reliable submission tests still run with the testnet instead of via the Docker container, because it would be quite complicated to set up a separate thread or something to close ledgers or trigger subscriptions during that time.

Because of this time reduction, it is now feasible to run the integration tests on each PR (especially because it's not as much load on the testnet), so this has been turned on.

### Context of Change

We used the standalone Docker container for xrpl.js testing, and the tests all ran very fast.

### Type of Change

## Test Plan

No functionality was changed. CI passes.

## Future Tasks
* Refactor so that each test has its own wallet to work with (don't need to worry about interactions between tests then)
* Document how to write tests for posterity


tests/integration/it_utils.py
```diff
@@ -86,8 +126,31 @@ def _choose_client_async(use_json_client: bool) -> Client:
...
+        else:
+            if is_json:
+                return JSON_RPC_TESTNET_CLIENT
+            else:
+                return WEBSOCKET_TESTNET_CLIENT
+    else:
+        if is_async:
+            return _choose_client_async(is_json)
+        else:
+            return _choose_client(is_json)
```
ledhed2222, line 145: removing these extraneous `else`s will make this much easier to read

```
def _get_client(....):
  if is_testnet:
    if is_async:
      if is_json:
        return ASYNC_JSON_RPC_TESTNET_CLIENT
      return ASYNC_WEBSOCKET_TESTNET_CLIENT
   if is_json:
      return JSON_RPC_TESTNET_CLIENT
   return WEBSOCKET_TESTNET_CLIENT
   
   if is_async:
     return _choose_client_async(is_json)
   return _choose_client(is_json)
```

but even better would be helper methods

ledhed2222: you could do this with a dictionary whose keys are bitmasks

```
# [is_testnet, is_async, is_json]
_CLIENT_CHOICES = {
  0x111: ASYNC_JSON_RPC_TESTNET_CLIENT,
  0x110: ASYNC_WEBSOCKET_TESTNET_CLIENT,
  ....
}

def _get_client(is_async: bool, is_json: bool, is_testnet: bool) -> Client:
  bit_array = [int(value) * 10 ** (i+1) for i, value in enumerate([is_json, is_async, is_testnet])]
  bit_combination = functools.reduce(operator.or_, bit_array)
  return _CLIENT_CHOICES[bit_combination]
  
```

I'm sure there's a more elegant way to do the bitwise stuff in python than what i've done, but you get what i mean

tests/integration/sugar/test_transaction.py
```diff
@@ -223,8 +220,8 @@ async def test_calculate_account_delete_fee(self, client):
             account_delete, WALLET, client
         )
 
-        # THEN we expect the calculated fee to be 5000000 drops (5 XRP)
-        expected_fee = "5000000"
+        # THEN we expect the calculated fee to be 50000000 drops (5 XRP)
+        expected_fee = "50000000"
```
khancode, line 224: Is this change expected? Changing fee from 5 to 50 XRP? If so, then the comment needs to be updated to `(50 XRP)`

khancode (approved):
LGTM besides one comment about changing XRP fee.

tests/integration/sugar/test_transaction.py
mvadari: Oh I'll look into that some more - I didn't actually calculate how many XRP it was, I assumed it was an error.

tests/integration/it_utils.py
```diff
@@ -60,34 +110,39 @@ def sign_and_reliable_submission(
...
-    else:
-        return ASYNC_WEBSOCKET_CLIENT
+    return _CLIENTS[(True, use_json_client, False)]
+
 
+def _get_client(is_async: bool, is_json: bool, is_testnet: bool) -> Client:
+    return _CLIENTS[(is_async, is_json, is_testnet)]
 
+
+# TODO: document how to write tests, for posterity
```
JST5000, line 139: Did you want to do this in this PR?

tests/integration/it_utils.py
```diff
@@ -5,35 +5,85 @@
...
+    (False, True, True): JSON_RPC_TESTNET_CLIENT,
+    (False, True, False): JSON_RPC_CLIENT,
+    (False, False, True): WEBSOCKET_TESTNET_CLIENT,
+    (False, False, False): WEBSOCKET_CLIENT,
+}
+
+MASTER_ACCOUNT = "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh"
+MASTER_SECRET = "snoPBrXtMeMyMHUVTgbuqAfg1SUTb"
+MASTER_WALLET = Wallet(MASTER_SECRET, 0)
+FUNDING_AMOUNT = "1200000000"
```
JST5000, line 58: This changed from what it used to be (1000000000 w/9 zeros), was that intentional?)

tests/integration/sugar/test_transaction.py
```diff
@@ -18,14 +25,17 @@
...
+
 CLEAR_FLAG = 3
 DOMAIN = "6578616D706C652E636F6D".lower()
 EMAIL_HASH = "10000000002000000000300000000012"
 MESSAGE_KEY = "03AB40A0490F9B7ED8DF29D246BF2D6269820A0EE7742ACDD457BEA7C7D0931EDB"
 SET_FLAG = 8
 TRANSFER_RATE = 0
 TICK_SIZE = 10
-FEE = "6000000"
+FEE = "60000000"
```
JST5000, line 38: Why did you increase the fee?

tests/integration/transactions/test_account_delete.py
```diff
@@ -19,7 +19,7 @@ class TestAccountDelete(IntegrationTestCase):
     async def test_all_fields(self, client):
         account_delete = AccountDelete(
             account=ACCOUNT,
-            fee=FEE,
+            # fee=FEE,
```
JST5000, line 22: Should we get rid of this line?

tests/integration/it_utils.py
mvadari: I think that's out of scope.

tests/integration/sugar/test_transaction.py
```diff
@@ -125,24 +127,19 @@ async def test_get_transaction_from_hash_with_min_max_ledgers(self, client):
...
         # THEN we expect to retrieve this transaction from its hash with
         # min_ledger and max_ledger parameters
         payment = await get_transaction_from_hash(
             payment_hash,
             client,
             False,
-            payment_ledger_index - 500,
-            payment_ledger_index + 500,
+            payment_ledger_index - 5,
+            payment_ledger_index + 5,
```
mvadari, line 142: This is because there are fewer ledgers in the standalone node - there aren't 500 ledgers to go back in time

tests/integration/it_utils.py
mvadari: A test was failing due to lack of funds, so I just increased the number of funds.

tests/integration/sugar/test_transaction.py
```diff
@@ -35,7 +36,7 @@
 SET_FLAG = 8
 TRANSFER_RATE = 0
 TICK_SIZE = 10
-FEE = "60000000"
+FEE = xrp_to_drops(6)
```
JST5000, line 39: This is actually 60 XRP, not 6

tests/integration/transactions/test_account_delete.py
```diff
@@ -28,7 +29,6 @@ async def test_all_fields(self, client):
             account_delete, WALLET, check_fee=False
         )
         self.assertEqual(response.status, ResponseStatus.SUCCESS)
-        WALLET.sequence += 1
```
JST5000, line 31: Why was this line deleted?

mvadari: The account delete doesn't actually work, because it requires 256 ledgers to go by. So the sequence number doesn't get incremented.

JST5000 (approved):
LGTM (And just to double check, merging into the `unknown-request` branch is intentional right?)

mvadari:
> (And just to double check, merging into the `unknown-request` branch is intentional right?)

Yes - it requires the `GenericRequest` for the `ledger_accept` methods on the standalone node

PR #319, opened 2021-11-24
feat: methods to convert between IssuedCurrency and IssuedCurrencyAmount

mvadari:

Title says it all.

### Context of Change

These methods make it way easier to work with these objects.

### Type of Change

## Test Plan

CI passes. Added tests for the new features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ledger_index` optional param for all the main account methods
 - `TicketCreate` transaction model
 - `GenericRequest` model for unsupported request types
+- Methods to convert between IssuedCurrency and IssuedCurrencyAmount
 
 ### Fixed
 - Makes the default ledger version for `get_next_valid_seq_number` `current` instead of `validated`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `ledger_index` optional param for all the main account methods
 - `TicketCreate` transaction model
 - `GenericRequest` model for unsupported request types
-- Methods to convert between IssuedCurrency and IssuedCurrencyAmount
+- Methods to convert between `IssuedCurrency` and `IssuedCurrencyAmount`
 
 ### Fixed
 - Makes the default ledger version for `get_next_valid_seq_number` `current` instead of `validated`
--- a/xrpl/models/amounts/issued_currency_amount.py
+++ b/xrpl/models/amounts/issued_currency_amount.py
@@ -51,7 +51,7 @@ class IssuedCurrencyAmount(IssuedCurrency):
             cls.from_dict({**issued_currency.to_dict(), "value": value}),
         )
 
-    def get_issued_currency(self: IssuedCurrencyAmount) -> IssuedCurrency:
+    def to_currency(self: IssuedCurrencyAmount) -> IssuedCurrency:
         """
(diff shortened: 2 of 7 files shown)


intelliot (approved):
LGTM

xrpl/models/amounts/issued_currency_amount.py
```diff
@@ -25,3 +28,34 @@ class IssuedCurrencyAmount(IssuedCurrency):
 
     :meta hide-value:
     """
+
+    @classmethod
+    def from_issued_currency(
+        cls: Type[IssuedCurrencyAmount], issued_currency: Currency, value: str
+    ) -> Union[IssuedCurrencyAmount, str]:
```
ledhed2222, line 35: this doesn't belong in this class because it doesn't necessarily return an IssuedCurrencyAmount. Instead i'd rather see a method on `XRP` and `IssuedCurrency` to convert to an `Amount`.

xrpl/models/amounts/issued_currency_amount.py
```diff
@@ -25,3 +28,34 @@ class IssuedCurrencyAmount(IssuedCurrency):
...
+            currency.
+        """
+        if isinstance(issued_currency, XRP):
+            return value
+        return cast(
+            IssuedCurrencyAmount,
+            cls.from_dict({**issued_currency.to_dict(), "value": value}),
+        )
+
+    def get_issued_currency(self: IssuedCurrencyAmount) -> IssuedCurrency:
```
ledhed2222, line 54: what about `to_issued_currency` instead?

ledhed2222: actually, `to_currency`. then your instance method on `Currency` can be `to_amount`

xrpl/models/amounts/issued_currency_amount.py
mvadari: That doesn't work type-wise - there's a circular import dependency then

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `ledger_index` optional param for all the main account methods
 - `TicketCreate` transaction model
+- Methods to convert between IssuedCurrency and IssuedCurrencyAmount
```
JST5000, line 12: nit
```suggestion
- Methods to convert between `IssuedCurrency` and `IssuedCurrencyAmount`
```

JST5000 (approved):
LGTM

khancode (approved):

xrpl/models/currencies/issued_currency.py
```diff
@@ -53,3 +54,23 @@ def _get_errors(self: IssuedCurrency) -> Dict[str, str]:
...
+            value: The amount of issued currency in the IssuedCurrencyAmount.
+
+        Returns:
+            An IssuedCurrencyAmount that represents the issued currency and the
+                provided value.
+        """
+        return cast(
+            xrpl.models.amounts.IssuedCurrencyAmount,
+            xrpl.models.amounts.IssuedCurrencyAmount.from_dict(
+                {**self.to_dict(), "value": value}
```
JST5000: Since it's just two fields (`Account` and `currency`) I feel like it would be better to just explicitly reconstruct a new `IssuedCurrencyAmount` to avoid any weirdness where adding a field to this object changes the amount. (But it's a style pref, this seems to work)

JST5000 (approved):
LGTM!

xrpl/models/currencies/issued_currency.py
JST5000: Hahaha, you're way ahead of me!

ledhed2222 (approved):

PR #322, opened 2021-12-03
fix: encoding issue with encoding non-traditional currency codes

mvadari:

This PR fixes an issue where some tokens with non-traditional currency codes weren't being encoded properly; they were being incorrectly encoded as `XRP`. 

### Context of Change

https://github.com/XRPLF/xrpl.js/pull/1857

### Type of Change

## Test Plan

Added a test that failed before the fix. CI passes.

### DIFF
--- a/xrpl/core/binarycodec/types/currency.py
+++ b/xrpl/core/binarycodec/types/currency.py
@@ -86,7 +86,7 @@ class Currency(Hash160):
         if self.buffer[0] != 0:
             # non-standard currency
             self._iso = None
-        elif code_bytes.hex() == "000000":
+        elif self.buffer.hex() == "0" * 40:  # all 0s
             # the special case for literal XRP
             self._iso = "XRP"
         else:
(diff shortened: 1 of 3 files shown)


xrpl/core/binarycodec/types/currency.py
```diff
@@ -86,7 +86,7 @@ def __init__(self: Currency, buffer: Optional[bytes] = None) -> None:
         if self.buffer[0] != 0:
             # non-standard currency
             self._iso = None
-        elif code_bytes.hex() == "000000":
+        elif self.buffer.hex() == "0" * 40:  # all 0s
```
khancode, line 89: Wow, I didn't know you can multiply a string in python. Haha very cool.

khancode (approved):
LGTM 👌

xrpl/core/binarycodec/types/currency.py
mvadari: It's one of the things that makes it so easy to work with strings in Python :)

JST5000 (approved):
LGTM

PR #323, opened 2021-12-03
fix: add workaround for rippled UNLModify encoding bug

mvadari:

There's a bug in rippled that means that UNLModify pseudotransactions aren't being encoded properly (the account isn't being included in the encoded pseudo-tx). This PR adds a workaround for that issue.

### Context of Change

https://github.com/XRPLF/xrpl.js/pull/1830

### Type of Change

## Test Plan

Added a test that failed before the fix. CI passes.

### DIFF
--- a/xrpl/core/binarycodec/types/serialized_dict.py
+++ b/xrpl/core/binarycodec/types/serialized_dict.py
@@ -29,6 +29,8 @@ _ACCOUNT: Final[str] = "Account"
 _SOURCE_TAG: Final[str] = "SourceTag"
 _DEST_TAG: Final[str] = "DestinationTag"
 
+_UNL_MODIFY_TX: Final[str] = "0066"
+
 
 def _handle_xaddress(field: str, xaddress: str) -> Dict[str, Union[str, int]]:
     """Break down an X-Address into a classic address and a tag.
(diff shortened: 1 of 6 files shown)


khancode (approved):

xrpl/core/binarycodec/types/serialized_dict.py
```diff
@@ -193,7 +195,18 @@ def from_value(
                 # keeps the original stack trace
                 e.args = (f"Error processing {field.name}: {e.args[0]}",) + e.args[1:]
                 raise
-            serializer.write_field_and_value(field, associated_value)
+            if field.name == "TransactionType" and str(associated_value) == "0066":
```
khancode, line 198: nit: put `"0066"` into a final variable with a name describing it's a UNLModify transaction type.

JST5000, line 198: When I run `print(str(0x0066))` in python it outputs `102`, so I'm not sure this is right

mvadari: That's by design - `associated_value.__str__` prints out the hex as a string

JST5000 (approved):
LGTM

PR #324, opened 2021-12-07
fix: use the ledger to determine the AccountDelete fee instead of hardcoding

mvadari:

This PR pulls the fee for an AccountDelete transaction from the ledger, instead of using a hardcoded value.

### Context of Change

A few months ago, the network voted to move the fee structure from 20/5 to 10/2. The fee for an autofilled AccountDelete transaction in xrpl-py wouldn't change, in that case, because it was hardcoded to 5 XRP. 

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/xrpl/asyncio/transaction/main.py
+++ b/xrpl/asyncio/transaction/main.py
@@ -11,9 +11,9 @@ from xrpl.constants import XRPLException
 from xrpl.core.addresscodec import is_valid_xaddress, xaddress_to_classic_address
 from xrpl.core.binarycodec import encode, encode_for_signing
 from xrpl.core.keypairs.main import sign
-from xrpl.models.requests import SubmitOnly
+from xrpl.models.requests import ServerState, SubmitOnly
 from xrpl.models.response import Response
-from xrpl.models.transactions.escrow_finish import EscrowFinish
+from xrpl.models.transactions import EscrowFinish
 from xrpl.models.transactions.transaction import Transaction
 from xrpl.models.transactions.transaction import (
     transaction_json_to_binary_codec_form as model_transaction_to_binary_codec,
@@ -25,7 +25,7 @@ from xrpl.wallet.main import Wallet
 _LEDGER_OFFSET: Final[int] = 20
 
 # TODO: make this dynamic based on the current ledger fee
-_ACCOUNT_DELETE_FEE: Final[int] = int(xrp_to_drops(5))
+_ACCOUNT_DELETE_FEE: Final[int] = int(xrp_to_drops(2))
 
 
 async def safe_sign_and_submit_transaction(
@@ -288,7 +288,10 @@ async def _calculate_fee_per_transaction_type(
 
     # AccountDelete Transaction
     if transaction.transaction_type == TransactionType.ACCOUNT_DELETE:
-        base_fee = _ACCOUNT_DELETE_FEE
+        if client is None:
+            base_fee = _ACCOUNT_DELETE_FEE
+        else:
+            base_fee = await _fetch_account_delete_fee(client)
 
     # Multi-signed Transaction
     # 10 drops × (1 + Number of Signatures Provided)
@@ -297,3 +300,9 @@ async def _calculate_fee_per_transaction_type(
 
     # Round Up base_fee and return it as a String
     return str(math.ceil(base_fee))
+
+
+async def _fetch_account_delete_fee(client: Client) -> int:
+    server_state = await client.request_impl(ServerState())
+    fee = server_state.result["state"]["validated_ledger"]["reserve_inc"]
+    return int(fee)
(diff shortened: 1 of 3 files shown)


mvadari:
The standalone Docker container is currently still set to a 20/5 fee structure - tests will pass when that's fixed.

xrpl/asyncio/transaction/main.py
```diff
@@ -25,7 +25,7 @@
 _LEDGER_OFFSET: Final[int] = 20
 
 # TODO: make this dynamic based on the current ledger fee
-_ACCOUNT_DELETE_FEE: Final[int] = int(xrp_to_drops(5))
+_ACCOUNT_DELETE_FEE: Final[int] = int(xrp_to_drops(2))
```
ledhed2222, line 28: should we just get rid of this now? this lib should probably always ask the ledger for the current fee for sugar methods. you'd use a transaction yourself if you want to specify a custom fee

mvadari: I agree - but as of right now, the autofill method does not require a client, so there needs to be some default value for that case. IMO the autofill logic needs a lot of re-working, based on our findings from xrpl.js. But that is way beyond the scope of this PR/ticket.

xrpl/asyncio/transaction/main.py
```diff
@@ -297,3 +300,9 @@ async def _calculate_fee_per_transaction_type(
 
     # Round Up base_fee and return it as a String
     return str(math.ceil(base_fee))
+
+
+async def _fetch_account_delete_fee(client: Client) -> int:
+    server_state = await client.request_impl(ServerState())
+    fee = server_state.result["state"]["validated_ledger"]["reserve_inc"]
```
JST5000, line 307: I think this should be `reserve_base` instead of `reserve_inc` for the account delete fee

https://xrpl.org/server_state.html#server_state

JST5000 (changes requested):
Could be wrong, but I think the field that corresponds to account delete fees may be incorrect.

xrpl/asyncio/transaction/main.py
mvadari: No, it's `reserve_inc` - `reserve_base` is the account creation fee/minimum account balance for an existing account. 
```
"validated_ledger": {
        "base_fee": 10,
        "close_time": 683153081,
        "hash": "B52AC3876412A152FE9C0442801E685D148D05448D0238587DBA256330A98FD3",
        "reserve_base": 20000000,
        "reserve_inc": 5000000,
        "seq": 65887201
      },
```
If you look at this example, you can see that `reserve_base` is 20 XRP and `reserve_inc` is 5 XRP. The account delete fee was 5 XRP, not 20.

JST5000: Ah I mis-counted the Zeros - I thought it was 2XRP/5XRP but that's not the case.

JST5000 (commented):
Happy to approve once the failing integration test is updated

mvadari:
> The standalone Docker container is currently still set to a 20/5 fee structure - tests will pass when that's fixed.
Turns out defaults are 200/50, and there's no real way to change that. So for now, the tests will account for that

mDuo13:
This should probably (carefully!) factor in load/escalation-based adjustments to the transaction cost. Since the cost to get into the open ledger or queue when they're full is based on "fee levels", not raw fee amounts, an AccountDelete paying 1 incremental reserve (e.g. 2 XRP right now) is treated the same as a reference transaction paying 10 drops.

JST5000 (approved):
LGTM!

mvadari:
> This should probably (carefully!) factor in load/escalation-based adjustments to the transaction cost. Since the cost to get into the open ledger or queue when they're full is based on ["fee levels"](https://xrpl.org/transaction-cost.html#fee-levels), not raw fee amounts, an AccountDelete paying 1 incremental reserve (e.g. 2 XRP right now) is treated the same as a reference transaction paying 10 drops.

I don't want to touch the AccountDelete load factoring right now - that's a bit too risky IMO. Also, I'm not sure how exactly it _should_ be raised - certainly not the same formula.

xrpl/asyncio/transaction/main.py
```diff
@@ -11,9 +11,9 @@
 from xrpl.core.addresscodec import is_valid_xaddress, xaddress_to_classic_address
 from xrpl.core.binarycodec import encode, encode_for_signing
 from xrpl.core.keypairs.main import sign
-from xrpl.models.requests import SubmitOnly
+from xrpl.models.requests import ServerState, SubmitOnly
 from xrpl.models.response import Response
-from xrpl.models.transactions.escrow_finish import EscrowFinish
+from xrpl.models.transactions import EscrowFinish
```
ledhed2222, line 16: why was this changed?

ledhed2222 (approved):

xrpl/asyncio/transaction/main.py
mvadari: It's neater

PR #325, opened 2021-12-07
fix: better error messages for reliable submissions

mvadari:

This PR makes the error messages for reliable submission easier to understand. Right now, all errors are "latest ledger sequence > last ledger sequence".

### Context of Change

This is a followup to #308. I noticed while dogfooding that the error messages are now really hard to understand and debug, because you don't get the preliminary result in the error messages.

This doesn't need a CHANGELOG update because it's incorporated in the line for #308.

### Type of Change

## Test Plan

CI passes. Tested the fix empirically.


JST5000 (approved):
LGTM

ledhed2222 (approved):

PR #326, opened 2021-12-08
fix: replace Decimal `setcontext` with `localcontext` context manager

mvadari:

This PR replaces all instances of Decimal's `setcontext` with the `localcontext` context manager.

Tests were failing for some weird dependency reason, so this PR also updates some dev dependencies as well.

### Context of Change

It's better practice to use a context manager than to set global state. Plus, it's less likely to cause problems.

### Type of Change

## Test Plan

CI passes (tests don't pass if the context is set incorrectly).

### DIFF
--- a/xrpl/core/binarycodec/types/amount.py
+++ b/xrpl/core/binarycodec/types/amount.py
@@ -4,7 +4,7 @@ See `Amount Fields <https://xrpl.org/serialization.html#amount-fields>`_
 """
 from __future__ import annotations
 
-from decimal import MAX_PREC, Context, Decimal, setcontext
+from decimal import MAX_PREC, Context, Decimal, localcontext
 from typing import Any, Dict, Optional, Type, Union
 
 from typing_extensions import Final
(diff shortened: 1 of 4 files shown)


mDuo13 (approved):
Not necessary as part of this PR, but might it be helpful to export the XRP and token contexts for clients to use for their own calculations as well?

JST5000 (approved):
LGTM!

xrpl/core/binarycodec/types/amount.py
```diff
@@ -24,8 +24,9 @@
 _MAX_MANTISSA: Final[int] = 10 ** 16 - 1
 
 # Configure Decimal
-setcontext(
-    Context(prec=_MAX_IOU_PRECISION, Emax=_MAX_IOU_EXPONENT, Emin=_MIN_IOU_EXPONENT)
+# TODO: export this context so others can use it
+_AMOUNT_CONTEXT: Final[Context] = Context(
+    prec=_MAX_IOU_PRECISION, Emax=_MAX_IOU_EXPONENT, Emin=_MIN_IOU_EXPONENT
```
ledhed2222, line 29: move this to `constants.py` and get rid of the todo

ledhed2222 (approved):

khancode (approved):
LGTM

ledhed2222 (approved):

PR #328, opened 2021-12-08
feat: scales fee based on ledger load

mvadari:

This PR refactors `get_fee` to incorporate the ledger load, so that the fee it returns scales appropriately when the ledger is under load, instead of always returning the minimum fee.

### Context of Change

A minimum 10-drop fee currently does not work on mainnet. This PR uses the [formula used in xrpl.js](https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/sugar/getFeeXrp.ts) to scale the fee based on ledger load.

### Type of Change

## Test Plan

CI passes. I tested by hand and confirmed that the fees are in the 5k drop range on mainnet right now.

### DIFF
--- a/xrpl/asyncio/ledger/main.py
+++ b/xrpl/asyncio/ledger/main.py
@@ -4,6 +4,7 @@ from typing import cast
 
 from xrpl.asyncio.clients import Client, XRPLRequestFailureException
 from xrpl.models.requests import Ledger, ServerInfo
+from xrpl.utils import xrp_to_drops
 
 
 async def get_latest_validated_ledger_sequence(client: Client) -> int:
@@ -54,7 +55,7 @@ async def get_fee(client: Client) -> str:
         client: the network client used to make network calls.
 
     Returns:
-        The minimum fee for transactions, in XRP.
+        The minimum fee for transactions, in drops.
 
     Raises:
         XRPLRequestFailureException: if the rippled API call fails.
@@ -67,7 +68,7 @@ async def get_fee(client: Client) -> str:
     if "validated_ledger" not in server_info:
         raise XRPLRequestFailureException(response.result)
 
-    base_fee = server_info["validated_ledger"]["base_fee_xrp"]
+    base_fee = int(xrp_to_drops(server_info["validated_ledger"]["base_fee_xrp"]))
 
     if "load_factor" not in server_info:
         load_factor = 1
@@ -76,4 +77,4 @@ async def get_fee(client: Client) -> str:
 
     fee = base_fee * load_factor
     # TODO: add cushion and max fee params to this method
-    return "%.6f" % fee
+    return str(int(fee))
--- a/xrpl/asyncio/ledger/main.py
+++ b/xrpl/asyncio/ledger/main.py
@@ -3,7 +3,7 @@
 from typing import cast
 
 from xrpl.asyncio.clients import Client, XRPLRequestFailureException
-from xrpl.models.requests import Fee, Ledger
+from xrpl.models.requests import Ledger, ServerInfo
 
 
 async def get_latest_validated_ledger_sequence(client: Client) -> int:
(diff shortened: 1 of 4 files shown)


xrpl/asyncio/ledger/main.py
```diff
@@ -54,13 +55,26 @@ async def get_fee(client: Client) -> str:
...
-    if response.is_successful():
-        return cast(str, response.result["drops"]["minimum_fee"])
+    response = await client.request_impl(ServerInfo())
+    if not response.is_successful():
+        raise XRPLRequestFailureException(response.result)
 
-    raise XRPLRequestFailureException(response.result)
+    server_info = response.result["info"]
+    if "validated_ledger" not in server_info:
+        raise XRPLRequestFailureException(response.result)
```
JST5000, line 69: Should these two conditions have different error messages? (Not successful vs `validated_ledger` missing)

JST5000 (approved):
LGTM

xrpl/asyncio/ledger/main.py
```diff
@@ -54,13 +55,26 @@ async def get_fee(client: Client) -> str:
         client: the network client used to make network calls.
 
     Returns:
-        The minimum fee for transactions.
+        The minimum fee for transactions, in drops.
```
mDuo13, line 58:
```suggestion
        The load-scaled transaction cost, in drops.
```

xrpl/asyncio/ledger/main.py
```diff
@@ -54,13 +55,26 @@ async def get_fee(client: Client) -> str:
...
+
+    base_fee = int(xrp_to_drops(server_info["validated_ledger"]["base_fee_xrp"]))
+
+    if "load_factor" not in server_info:
+        load_factor = 1
+    else:
+        load_factor = server_info["load_factor"]
+
+    fee = base_fee * load_factor
+    # TODO: add cushion and max fee params to this method
```
mDuo13, line 79: Max fee is super important. We _really_ don't want people to auto-fill a really high fee because they happened to catch a server at its busiest. People tend to get upset if sending a transaction burns 1000x as much as it delivers, and auto-filling doesn't always provide an opportunity to double-check that first.

mDuo13 (changes requested):

xrpl/asyncio/ledger/main.py
mvadari: Hm, would this be a breaking change? Should we just change how autofill does it?

mDuo13: I guess you could view it that way, but I think it's more like a documentation fix.

mDuo13 (approved):

xrpl/asyncio/ledger/main.py
mvadari: I was referring to the change itself - this seemed like a good illustration of my concern.

xrpl/asyncio/ledger/main.py
```diff
@@ -46,21 +48,44 @@ async def get_latest_open_ledger_sequence(client: Client) -> int:
...
         XRPLRequestFailureException: if the rippled API call fails.
     """
     response = await client.request_impl(Fee())
     if response.is_successful():
-        return cast(str, response.result["drops"]["minimum_fee"])
+        result = response.result["drops"]
+        if fee_type == "open":
+            fee = cast(str, result["open_ledger_fee"])
+        elif fee_type == "minimum":
+            fee = cast(str, result["minimum_fee"])
```
JST5000, line 79: Why do you use `cast(str,...)` here instead of `str(...)`?

JST5000 (approved):
LGTM!

xrpl/asyncio/ledger/main.py
mvadari: It's already a string, but mypy doesn't know that - `cast` is the same as `as string` in TS.

khancode (approved):
LGTM 👌

xrpl/asyncio/ledger/main.py
```diff
@@ -46,21 +48,44 @@ async def get_latest_open_ledger_sequence(client: Client) -> int:
     raise XRPLRequestFailureException(response.result)
 
 
-async def get_fee(client: Client) -> str:
+async def get_fee(
+    client: Client, *, max_fee: Optional[float] = 2, fee_type: str = "open"
```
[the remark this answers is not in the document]
mvadari: I put it there so people don't accidentally add additional args without meaning to.

PR #331, opened 2021-12-11
feat: make `autofill` a public method

mvadari:

This PR makes `_autofill_transaction` into a public method `autofill`. This exposes it to library users.

Tests were failing for some weird dependency reason, so this PR also updates some dev dependencies as well.

### Context of Change

Requested by @mDuo13 because there is only a sign-and-autofill public method, not just an autofill method.

### Type of Change

## Test Plan

CI passes. Adjusted a few tests to call `autofill_transaction` directly instead of sign-and-autofill.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Methods to convert between `IssuedCurrency` and `IssuedCurrencyAmount`
 - Support for ints and floats in the `IssuedCurrency` and `IssuedCurrencyAmount` models (and ints for `XRP`)
 - `max_fee` and `fee_type` optional params for `get_fee`
-- Makes `autofill_transaction` a publicly exposed method
+- Makes `autofill` a publicly exposed method
 
 ### Fixed
 - Makes the default ledger version for `get_next_valid_seq_number` `current` instead of `validated`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Methods to convert between `IssuedCurrency` and `IssuedCurrencyAmount`
 - Support for ints and floats in the `IssuedCurrency` and `IssuedCurrencyAmount` models (and ints for `XRP`)
 - `max_fee` and `fee_type` optional params for `get_fee`
-- Makes `autofill` a publicly exposed method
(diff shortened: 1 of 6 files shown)


mDuo13 (changes requested):
Haven't tried it yet, but the general idea looks right. But, I suggest naming it just `autofill`.

khancode (approved):
LGTM. I agree with Rome to rename it to just `autofill`

mvadari:
> But, I suggest naming it just autofill.
@mDuo13 I wanted to keep the naming consistent with everything else here. I agree that renames are needed, but that can happen in a separate PR.

mDuo13:
> I agree that renames are needed, but that can happen in a separate PR.

I disagree about doing it in a separate PR. I don't think it's wise to introduce a new public method, even if only temporarily/pre-release, if the plan is to change its name anyway. Better to get the name right the first time.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GenericRequest` model for unsupported request types
 - Methods to convert between `IssuedCurrency` and `IssuedCurrencyAmount`
 - Support for ints and floats in the `IssuedCurrency` and `IssuedCurrencyAmount` models (and ints for `XRP`)
+- Makes `autofill` a publicly exposed method
```
intelliot, line 16:
```suggestion
- `autofill`, a new public method that populates the `fee`, `sequence`, and `last_ledger_sequence` fields of a transaction, based on the current state retrieved from the server the Client is connected to. It also converts all X-Addresses to classic addresses.
```

- Most users probably weren't aware that there was a private autofill method before
- This matches the grammar of other lines (e.g. "`xrpl.utils.str_to_hex` and `xrpl.utils.hex_to_str` helpers") and reads naturally off of "Added", as in: Added `autofill`, a new public method…

tests/integration/sugar/test_transaction.py
```diff
@@ -268,9 +263,7 @@ async def test_calculate_payment_fee(self, client):
         )
 
         # AFTER autofilling the transaction fee
-        payment_signed = await safe_sign_and_autofill_transaction(
-            payment, WALLET, client
-        )
+        payment_signed = await autofill(payment, client)
```
JST5000, line 266: This no longer is signed, should `payment_signed` be renamed? (Or is this supposed to be signed still?

ledhed2222 (approved):

PR #332, opened 2021-12-13
quick regular expression use cleanup

ledhed2222:

Quick cleanup of regex usage. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/xrpl/constants.py
+++ b/xrpl/constants.py
@@ -1,7 +1,7 @@
 """Collection of public constants for XRPL."""
+import re
 from decimal import Context
 from enum import Enum
-from re import compile
 from typing import Pattern
 
 from typing_extensions import Final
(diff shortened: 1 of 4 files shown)


xrpl/constants.py
```diff
@@ -1,6 +1,6 @@
 """Collection of public constants for XRPL."""
+import re
 from enum import Enum
-from re import compile
```
JST5000, line 3: Why not import the specific function? (Seems like that's what we normally do)

JST5000: Nevermind, I see there's a different standard in the other files you edited.

JST5000 (approved):

xrpl/constants.py
ledhed2222: yeah, did that because imo in this case, like with the `json` package, it's not at all clear to see:

```python
from re import compile

.....

regex = compile("[A-Z]")
```

i think you want to see the full `re.compile`

mvadari (approved):

PR #334, opened 2021-12-16
chore: update dev dependencies so tests pass

mvadari:

This PR updates `mypy` and `typing-ast` (both dev dependencies) to their latest versions, and fixes some minor issues that mypy complained about.

### Context of Change

Tests started failing for some weird dependency reason that was fixed in `typing-ast` v1.5.0. The version of `mypy` we had on the repo did not support `typing-ast` v1.5.0, so that had to be updated as well.

### Type of Change

## Test Plan

CI passes.


JST5000 (approved):

khancode (approved):
LGTM

PR #342, opened 2022-01-09
Adding a transactionparser

LimpidCrypto:
Changes marked for commit:
	changed:       xrpl/utils/__init__.py
	new folder:     xrpl/utils/txn_parser

Add a transaction parser. It supports "parse_balance_changes", "parse_final_balances", "parse_previous_balances" and "parse_order_book_changes".
Also included the functions in the "xrpl/utils/__init__.py" file.
This is a rewritten parser of the https://github.com/ripple/ripple-lib-extensions/blob/master/transactionparser/src/balancechanges.js in python.


xrpl/utils/tx_parser.py
```diff
@@ -0,0 +1,197 @@
...
+    def __init__(self, difftype: str, entryType: str, ledgerIndex: str, newFields: dict, finalFields: dict, previousFields: dict) -> None:
+        self.DiffType = difftype
+        self.EntryType = entryType
+        self.LedgerIndex = ledgerIndex
+        self.NewFields = newFields
+        self.FinalFields = finalFields
+        self.PreviousFields = previousFields
+
+
+def __normalize_node(affectedNode):
```
mvadari, line 19: Internal functions should be prepended by just one `_` not two
```suggestion
def _normalize_node(affectedNode):
```
ditto for the other ones

xrpl/utils/tx_parser.py
```diff
@@ -0,0 +1,197 @@
+from xrpl.utils.xrp_conversions import drops_to_xrp
+import pydash
+
+"""
+This is a rewritten repository of https://github.com/ripple/ripple-lib-extensions/blob/master/transactionparser/src/balancechanges.js in python.
+"""
+
+### Utils ###
+class Normalized_Node():
```
mvadari, line 9: Class names shouldn't have underscores in them
```suggestion
class NormalizedNode():
```

xrpl/utils/tx_parser.py
```diff
@@ -0,0 +1,197 @@
...
+    def __init__(self, difftype: str, entryType: str, ledgerIndex: str, newFields: dict, finalFields: dict, previousFields: dict) -> None:
+        self.DiffType = difftype
```
mvadari, line 11:
```suggestion
        self.diff_type = difftype
```
Please use `snake_case` for variable names.

xrpl/utils/tx_parser.py
```diff
@@ -0,0 +1,197 @@
...
+def __parse_final_balance(node):
+    if "Balance" in node.NewFields.keys():
+        return __parse_value(node.NewFields["Balance"])
+    elif "Balance" in node.FinalFields.keys():
+        return __parse_value(node.FinalFields["Balance"])
+
+    return None
+
+
+def __parse_XRP_quantity(node, valueParser):
```
mvadari, line 92:
```suggestion
def _parse_xrp_quantity(node, valueParser):
```
Function names shouldn't have capital letters.

mvadari:
Here's the PR from xrpl.js for `getBalanceChanges`, for reference

xrpl/utils/__init__.py
```diff
@@ -8,6 +8,11 @@
     ripple_time_to_datetime,
     ripple_time_to_posix,
 )
+from xrpl.utils.tx_parser import (
+    XRPLMetadataExcetion,
```
mvadari, line 12: Typo here - should be `XRPLMetadataException`

xrpl/utils/tx_parser.py
```diff
@@ -0,0 +1,277 @@
+import pydash
```
mvadari, line 1: This new package needs to be added to the poetry dependencies. Instructions for how to use that are in `CONTRIBUTING.md`.

xrpl/utils/tx_parser.py
```diff
@@ -0,0 +1,277 @@
...
+from typing import Dict, List, Union
+from xrpl.constants import XRPLException
+from xrpl.utils.xrp_conversions import drops_to_xrp
+
+
+class XRPLMetadataExcetion(XRPLException):
```
mvadari, line 7:
```suggestion
class XRPLMetadataException(XRPLException):
```

xrpl/utils/tx_parser.py
```diff
@@ -0,0 +1,277 @@
...
+            values.append(_parse_xrp_quantity(node, valueParser))
+        elif node.entry_type == "RippleState":
+            values.append(_parse_trustline_quantity(node, valueParser))
+        else:
+            values.append([])
+
+    return _group_by_address(pydash.compact(pydash.flatten(values)))
+
+
+def parse_balance_changes(metadata: dict) -> Dict[str, List[Dict[str, str]]]:
```
mvadari, line 210: Can you add unit tests for these methods in `tests/unit/utils`?

mvadari:
It'll be easier to work with the linter if you set up the pre-commit hook. Instructions for that are in `CONTRIBUTING.md`.

mvadari:
You'll need to add the `pydash` package via Poetry, not just editing the `.lock` file.

LimpidCrypto:
Thanks for your patience @mvadari . As you may have noticed this is the first time I contribute to such a project :D Also the first time I worked with pyenv and poetry. 
So the unit test works fine now. The only error I still get when I run pre-commit is the following:
```xrpl/utils/tx_parser.py:8: error: Skipping analyzing "pydash": module is installed, but missing library stubs or py.typed marker```

Is this a critical error? As defined here https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports I would have to create a ```py.typed``` file in ```pydash```. Just as in dependencies like e.g. ```websockets```.

**Comment by mvadari** [comment]
It's not a critical error - it's not worth dealing with for this PR. 

I think if you can get around using pydash that would be preferable, but if you can't then that's fine.

**Comment by wojake** [comment]
When will this be merged? I'd love to use this.

**Comment by LimpidCrypto** [comment]
The PR needs to be reviewed first. I also just found a miscalculation in _flip_trustline_perspective. I will fix that tomorrow.

File: xrpl/utils/tx_parser/balance_changes.py [hunk 9]
```diff
@@ -0,0 +1,231 @@
+"""
+Parse balance changes, final balances and previouse balances of every
```
JST5000, line 2:
```suggestion
Parse balance changes, final balances and previous balances of every
```

xrpl/utils/tx_parser/balance_changes.py
```diff
@@ -0,0 +1,231 @@
...
+
+    all_balances: Dict[str, List[Dict[str, str]]] = {}
+    """All balance changes of all accounts
+    affected by the transaction.
+    """
+    source_account_balances: Dict[str, List[Dict[str, str]]] = {}
+    """All balance changes of the source account (Field: `Account`)
+    which was affected by the transaction.
+    """
+    destination_account_balances: Union[None, Dict[str, List[Dict[str, str]]]] = None
```
JST5000, line 35: nit: Instead of `Union[None, ...]` you can use `Optional[...]`

xrpl/utils/tx_parser/balance_changes.py
```diff
@@ -0,0 +1,231 @@
...
+    def _parse(self: Any) -> None:
+        is_valid_metadata(metadata=self.metadata)
+        parsedQuantities = parse_quantities(
+            metadata=self.metadata, valueParser=compute_balance_changes
+        )
+
+        result: Dict[str, List[Dict[str, str]]] = {}
+        for k, v in parsedQuantities.items():
+            address = k
+            change = v
```
JST5000, line 112:
```suggestion
        for address, change in parsedQuantities.items():
```
The values are based on the position, so you can name them to whatever you'd like :)

xrpl/utils/tx_parser/balance_changes.py
```diff
@@ -0,0 +1,231 @@
...
+        Attributes:
+            metadata (`Dict[str, Union[str, int, bool, Dict[str, Any]]]`):
+                Transaction metadata including the account that
+                sent the transaction and the affected nodes.
+        """
+        super().__init__(metadata)
+
+        self._parse()
+
+    def _parse(self: Any) -> None:
```
JST5000, line 162: If you'd like to have `self: ParseFinalBalances` instead of `Any`, you just need to put `import __future__` at the top of the file. That allows python to see that you defined `ParseFinalBalances` even if at this point in the file the definition isn't done yet.

xrpl/utils/tx_parser/balance_changes.py
```diff
@@ -0,0 +1,231 @@
...
+        is_valid_metadata(metadata=self.metadata)
+        parsedQuantities = parse_quantities(self.metadata, parse_final_balance)
+
+        result: Dict[str, List[Dict[str, str]]] = {}
+        for k, v in parsedQuantities.items():
+            address = k
+            change = v
```
JST5000, line 169: Same change as above

xrpl/utils/tx_parser/balance_changes.py
```diff
@@ -0,0 +1,231 @@
...
+    def _parse(self: Any) -> None:
+        is_valid_metadata(metadata=self.metadata)
+        balance_changes = ParseBalanceChanges(metadata=self.metadata).all_balances
+        final_balances = ParseFinalBalances(metadata=self.metadata).all_balances
+
+        for account, balances in balance_changes.items():
+            for count, balance in enumerate(balances):
+                final_balances_value = final_balances[account][count]["Value"]
+                final_balances[account][count]["Value"] = str(
+                    Decimal(final_balances_value) - Decimal(balance["Value"])
```
JST5000, line 228: Should this be addition? (Are the changes listed where +100 = Add 100 XRP to the previous value?)

xrpl/utils/tx_parser/orderbook_changes.py
```diff
@@ -0,0 +1,74 @@
...
+    """Parse order book changes from a transactions metadata
+    in a easy-to-read format.
+
+    Args:
+        metadata (`Dict[str, Union[str, int, bool, Dict[str, Any]]]`):
+            The transactions metadata.
+    """
+
+    all_orderbook_changes: Dict[str, List[Dict[str, Any]]] = {}
+    """All orderbook_changes"""
```
JST5000, line 23: What does the key represent?

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,847 @@
...
+
+def is_valid_metadata(metadata: Dict[str, Any]) -> None:
+    """Check if the given metadata is valid.
+
+    Args:
+        metadata (Dict[str, Any]): The transactions metadata
+
+    Raises:
+        XRPLMetadataException: If field Account is missing.
+        XRPLMetadataException: If field Account is missing.
```
JST5000, line 42: Duplicate line

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,847 @@
...
+        if isinstance(field_value, dict):
+            # if 'field_value' is a issued currency amount
+            if field_name in ["Balance", "LowLimit", "HighLimit"]:
+                field_value = Balance(
+                    counterparty=field_value["issuer"],
+                    currency=field_value["currency"],
+                    value=field_value["value"],
+                )
+            else:  # if 'field_value' type is dict but is no issued currency amount.
+                factured_field_name = factory(field_name)
```
JST5000, line 169: What does factured mean?

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,847 @@
...
+
+
+"""
+#####################
+Utils balance_changes
+#####################
+"""
+
+
+class Balance:
```
JST5000, line 234: This might be the same as an `IssuedCurrencyAmount`. 

You may also want to see if `IssuedCurrency.py` is a useful file for you.

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,847 @@
...
+        value = _parse_value(node.new_fields.Balance)
+    if hasattr(node.previous_fields, "Balance") and hasattr(
+        node.final_fields, "Balance"
+    ):
+        value = _parse_value(node.final_fields.Balance) - _parse_value(
+            node.previous_fields.Balance
+        )
+
+    # if value is not 'None' or '0' return 'value' else 'None'.
+    return None if value is None else None if value == 0 else value
```
JST5000, line 323:
```suggestion
     if( value and value != 0):
           return value
     else:
          return None
```
There's a lot of None's in this line and it's a bit confusing to read.

xrpl/utils/tx_parser/utils.py
JST5000: Or maybe not because the `counterparty` doesn't correspond to the currency, it corresponds to the payment?

xrpl/utils/tx_parser/balance_changes.py
LimpidCrypto: Ah, I wondered how I would do that 👍🏼

xrpl/utils/tx_parser/balance_changes.py
LimpidCrypto: Example:
`BalanceChanges`:
-100 XRP
`FinalBalance`:
900 XRP
`PreviousBalance`:
900 XRP - -100 XRP = 1000 XRP

If the balance change would increase the balance by 100 the calculation of the `PreviousBalance` would be:
900 XRP - 100 XRP = 800 XRP

So subtraction should be correct :)

xrpl/utils/tx_parser/orderbook_changes.py
LimpidCrypto: It represents the accounts address :)

xrpl/utils/tx_parser/utils.py
LimpidCrypto: I will rename it 👍🏼

xrpl/utils/tx_parser/orderbook_changes.py
JST5000: Could you add that to the docstring for future reference?

xrpl/utils/tx_parser/balance_changes.py
JST5000: That makes sense! Thanks for the example :)

LimpidCrypto:
Is it my fault the integration tests failed? They failed because the transaction tests did not succeed.

JST5000:
> Is it my fault the integration tests failed? They failed because the transaction tests did not succeed.

That's odd. I'll try to dig into this a bit before the end of today and get back to you.

JST5000:
@LimpidCrypto Turns out it was just a flaky test, sorry about that! (The sequence number didn't match up in the failing test case, and my hypothesis is that for some reason the mock-rippled instance wasn't being restarted, so the sequence number just kept going up. If it happens again I'll debug deeper)

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,850 @@
...
+        if (hasattr(node.final_fields, "Account"))
+        else node.new_fields.Account
+    )
+
+    return order_change.__dict__
+
+
+def parse_order_book_changes(
+    nodes: Iterable[NormalizedNode],
+) -> Any:
```
JST5000, line 835: Small improvement 
```suggestion
) -> List[Any]:
```

xrpl/utils/tx_parser/orderbook_changes.py
```diff
@@ -0,0 +1,78 @@
...
+        metadata: Dict[str, Union[str, int, bool, Dict[str, Any]]],
+    ) -> None:
+        """
+        Args:
+            metadata (`Dict[str, Union[str, int, bool, Dict[str, Any]]]`):
+                The transactions metadata.
+        """
+        self.metadata = metadata
+
+    def _define_results(self: Any, result: Dict[str, List[Dict[str, Any]]]) -> None:
```
JST5000, line 38:
```suggestion
    def _define_results(self: OrderBookChanges, result: Dict[str, List[Dict[str, Any]]]) -> None:
```

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,850 @@
...
+
+    if "AffectedNodes" not in metadata["meta"] or not metadata["meta"]["AffectedNodes"]:
+        raise XRPLMetadataException("Metadata incomplete: No nodes provided.")
+
+
+class NormalizedNode:
+    """A standard format for nodes."""
+
+    def __init__(
+        self: Any,
```
JST5000, line 69:
```suggestion
        self: NormalizedNode,
```

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,850 @@
...
+        diffType: str,
+        entryType: str,
+        ledgerIndex: str,
+        newFields: Any,
+        finalFields: Any,
+        previousFields: Any,
```
JST5000, line 75: Are these actually Any? (Or are they a list of some sort?)

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,850 @@
...
+            State of the field.
+        fields:
+            A dictionary of all fields that field state has.
+            E.g.: {'Balance': '62537659', 'OwnerCount': 16, 'Sequence': 67702065}
+
+    Returns:
+        The full field state with all its fields as objects.
+    """
+
+    def factory(name: str) -> Any:
```
JST5000, line 133: Would creating a base class you return from factory help cut down on the use of Any? 

(The main reason being that Mypy loses a lot of effectiveness the more `Any`s enter the code because they multiply)

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,850 @@
...
+
+    if value and value != 0:
+        return value
+    else:
+        return None
+
+
+def _parse_xrp_quantity(
+    node: NormalizedNode,
+    valueParser: Any,
```
JST5000, line 330: If this is a function, you can probably import `Callable` from typing

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,850 @@
...
+    balanceChanges: List[TrustLineQuantity],
+) -> Any:
+    """Groups the balances changes by address.
+
+    Args:
+        balanceChanges (List[TrustLineQuantity]):
+            A list of trust line quantities.
+
+    Returns:
+        Any: A dictionary with all balance changes grouped by addresses.
```
JST5000, line 433: This description implies this could be more specific in typing

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,850 @@
...
+
+
+def _get_expiration_time(node: NormalizedNode) -> Union[Any, None]:
+    """Formats the ripple timestamp to a easy to read format.
+
+    Args:
+        node (NormalizedNode): Normalized node.
+
+    Returns:
+        Union[Any, None]:
```
JST5000, line 714: This return type can be more specific (I think it's a string?)

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,850 @@
...
+            delattr(order, k)
+
+    return order
+
+
+def _calculate_received_and_paid_amount(
+    taker_gets: Any,
+    taker_pays: Any,
+    direction: str,
+) -> Any:
```
JST5000, line 757: When you return multiple things you can represent that with `Tuple[first_type, second_type]`

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,850 @@
...
+def parse_order_book_changes(
+    nodes: Iterable[NormalizedNode],
+) -> Any:
+    """Filter nodes by 'EntryType': 'Offer'.
+
+    Args:
+        nodes (Iterable[NormalizedNode]): Affected nodes.
+
+    Returns:
+        Any:
```
JST5000, line 842: This can also be a list

LimpidCrypto:
## Improve type hints commits:
- Added `normalize_metadata` to format metadata received from subscritions to have one standard format so that typing is easier
- Normalize nodes in the `Parse...` classes so the `metadata` type hint do not need to be repeated that often
- Renamed `Balance` to `AccountBalance` so it won't get in conflict with the `Balance` field in metadata
- Added a lot of type hints
- Formatted some arg/attr names
- Corrected some doc strings

Explanations of added `# type: ignore`:
- `# type: ignore` in `normalize_metadata` is because `mypy` is identifying a false type hint of metadata
- `# type: ignore` every time a function returns a return of `pydash` because `mypy` always identifies them as `Any`

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1054 @@
...
+                        str,
+                        Dict[
+                            str,
+                            Union[str, int, Dict[str, Union[str, int, Dict[str, str]]]],
+                        ],
+                    ],
+                ],
+            ],
+        ],
+    ],
```
JST5000, line 79: Holy crap that's one hell of a type! 

I'd recommend turning this into a class or type alias, that way you can define the type once and then use it everywhere else. 

(That would also let you turn instances of `metadata: Any` above into something like `metadata: Metadata` which is nice and readable :) )

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1054 @@
...
+
+
+def normalize_node(
+    affected_node: Dict[
+        str,  # Node state / diff_type
+        Dict[
+            str,
+            Union[str, int, Dict[str, Union[str, int, Dict[str, str]]]],
+        ],
+    ]
```
JST5000, line 265: This type is probably a useful sub-type to define for the Metadata type :)

xrpl/utils/tx_parser/utils.py
JST5000: Here's the doc on type aliases - https://mypy.readthedocs.io/en/stable/kinds_of_types.html#type-aliases

xrpl/utils/tx_parser/orderbook_changes.py
```diff
@@ -0,0 +1,113 @@
...
+
+class OrderBookChanges:
+    """Parse order book changes from a transactions metadata
+    in a easy-to-read format.
+    """
+
+    # all_orderbook_changes = {'account': ['order book changes']}
+    all_orderbook_changes: Dict[str, List[Dict[str, Any]]] = {}
+
+    def __init__(self: OrderBookChanges, metadata: Any) -> None:
```
JST5000, line 23: Does this metadata field re-use the type you defined elsewhere?

JST5000:
For some of the longer types I think it's worth using a type alias even if it's just used once. That'll let you keep the function definitions focused on the implementations instead of the type. (Then if folks want to learn what the type signature actually means they can opt-into that information)

xrpl/utils/tx_parser/utils.py
LimpidCrypto: I thought that this would be possible but did not find the right solution to that. Thank you I will change that 🙏

xrpl/utils/tx_parser/orderbook_changes.py
LimpidCrypto: I defined the parent classes `metadata` as `Any` because I did not wanted to define the huge type hint once again.
So every child class has the exact type hint but the parent classes don't. This also will be improved by using type aliases :)

docs/source/xrpl.utils.tx_parser.rst
```diff
@@ -0,0 +1,37 @@
+xrpl.utils.tx\_parser package
```
mvadari, line 1: This doc shouldn't be added - the docs for the new function will be in the `xrpl.utils` file

xrpl/utils/tx_parser/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+    ParsePreviousBalances,
+)
+from xrpl.utils.tx_parser.orderbook_changes import ParseOrderBookChanges
+from xrpl.utils.tx_parser.utils import XRPLMetadataException
+
+__all__ = [
+    "ParseBalanceChanges",
+    "ParseFinalBalances",
+    "ParsePreviousBalances",
+    "ParseOrderBookChanges",
```
mvadari, line 15: All of these methods should use `snake_case`, since they're Python functions

xrpl/utils/tx_parser/balance_changes.py
```diff
@@ -0,0 +1,332 @@
...
+    compute_balance_changes,
+    is_valid_metadata,
+    normalize_metadata,
+    normalize_nodes,
+    parse_final_balance,
+    parse_quantities,
+)
+
+
+class BalanceChanges:
```
mvadari, line 20: Why are these all classes instead of methods? IMO if they're only existing to process the metadata, they should be methods.

xrpl/utils/tx_parser/balance_changes.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+class ParsePreviousBalances(BalanceChanges):
+    """Parse the previous balances of all accounts affected
+    by the transaction before it occurred.
+    """
+
+    def __init__(
+        self: ParsePreviousBalances,
+        metadata: Dict[  # metadata received from a tx method
```
mvadari, line 259: You can define this type at the top or in your utils file, which will make it much neater.

mvadari: e.g.
```py
Metadata = Dict[str, ...]
```

mvadari: You can even nest types to make it neater

xrpl/utils/tx_parser/balance_changes.py
LimpidCrypto: That way I think it is easier to access e.g. `.source_account_balances` directly from the class. If you think this is not necessary or too confusing I can easily change them up to functions and just return all balance changes, final balances, ... :)

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+            "value": str(negated_balance),
+        },
+    )
+
+    return result
+
+
+def _parse_trustline_quantity(
+    node: NormalizedNode,
+    value_parser: Callable[[NormalizedNode], Union[Decimal, None]],
```
JST5000, line 539: nit:
```suggestion
    value_parser: Callable[[NormalizedNode], Optional[Decimal]],
```

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+) -> Union[List[TrustLineQuantity], None]:
```
JST5000, line 540: nit:
```suggestion
) -> Optional[List[TrustLineQuantity]]:
```

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+    mapped_group = map_values(
+        grouped, lambda group: map_(group, lambda node: node.balance)
+    )
+
+    return mapped_group  # type: ignore
+
+
+def parse_quantities(
+    nodes: List[NormalizedNode],
+    value_parser: Callable[[NormalizedNode], Union[Decimal, None]],
```
JST5000, line 600: ditto

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+    return AccountBalance(
+        currency=currency_amount.currency,
+        counterparty=currency_amount.issuer,
+        value=currency_amount.value,
+    )
+
+
+def _calculate_delta(
+    final_amount: Union[AccountBalance, None],
+    previous_amount: Union[AccountBalance, None],
```
JST5000, line 762: ditto

JST5000 (approved):
Looks good to me - small nitpick about using Optional, but very solid work!

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+]
+
+
+class XRPLMetadataException(XRPLException):
+    """Exception for invalid transaction metadata."""
+
+    pass
+
+
+def is_valid_metadata(
```
mvadari, line 82:
```suggestion
def validate_metadata(
```
Using the `is_valid` implies that it's going to be returning a boolean as to whether the metadata is valid or not, but that's not what this method does.

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+
+    Raises:
+        XRPLMetadataException: If field Account is missing.
+        XRPLMetadataException: If field meta is missing.
+        XRPLMetadataException: If field AffectedNodes is missing or is empty.
+    """
+    if "transaction" in metadata:
+        if "Account" not in metadata["transaction"]:  # type: ignore
+            raise XRPLMetadataException(
+                "Metadata incomplete: Metadata field 'Account' must be included."
```
mvadari, line 99:
```suggestion
                "Malformed metadata: Metadata field 'Account' must be included."
```
ditto to the other messages

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+
+    Args:
+        metadata (Union[METADATA_TYPE, SUBSCRIPTION_METADATA_TYPE]):
+            Transactions metadata.
+
+    Returns:
+        Metadata in standard format.
+    """
+    transaction = metadata["transaction"]
+    assert isinstance(transaction, dict)
```
mvadari, line 129: You can simplify this logic and avoid the `isinstance` if `METADATA_TYPE` and `SUBSCRIPTION_METADATA_TYPE` are defined with `TypedDict`: https://docs.python.org/3/library/typing.html#typing.TypedDict

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+class FinalFields(FieldState):
+    """FinalFields"""
+
+    pass
+
+
+class PreviousFields(FieldState):
+    """PreviousFields"""
+
+    pass
```
mvadari, line 182: Why are these defined as separate classes if they're all identical?

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+
+    Balance: Union[AccountBalance, str]
+    LowLimit: Union[AccountBalance, str]
+    HighLimit: Union[AccountBalance, str]
+    Account: str
+    TakerGets: Union[FieldName, str]  # Type: 'NewClass'
+    TakerPays: Union[FieldName, str]  # Type: 'NewClass'
+    Sequence: int
+    Flags: int
+    Expiration: Optional[Union[int, str]] = None
```
mvadari, line 164: These might be easier to work with if you make them dataclasses

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+        """
+        self.diff_type = diff_type
+        self.entry_type = entry_type
+        self.ledger_index = ledger_index
+        self.new_fields = new_fields
+        self.final_fields = final_fields
+        self.previous_fields = previous_fields
+
+
+def field_factory(
```
mvadari, line 217: This method feels very hacky. Typing doesn't need to be so restrictive in this package that we need to do some metaprogramming in order to use it.

xrpl/utils/tx_parser/utils.py
```diff
@@ -0,0 +1,1052 @@
...
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
+
+from pydash import compact, filter_, flatten, group_by, map_, map_values  # type: ignore
+from typing_extensions import Literal
+
+from xrpl.constants import XRPLException
+from xrpl.utils.xrp_conversions import drops_to_xrp
+
+"""
+###########################################
```
mvadari, line 15: This file is very long, and it looks like it has distinct parts. Why is it not split into multiple files?

mvadari (commented):
Please run `poetry run sphinx-apidoc -o docs/source xrpl` to generate the new documentation files for this PR.

xrpl/utils/tx_parser/utils.py
LimpidCrypto: I found a way to make this function way easier 👍

docs/source/xrpl.utils.txn_parser.rst
```diff
@@ -0,0 +1,37 @@
+xrpl.utils.txn\_parser package
```
mvadari, line 1: These two docs pages should actually both just contain `:orphan:` - since you're exporting the txn parser stuff at the `xrpl.utils` level, the docs will appear there already, and you don't need separate pages.

Sorry for the confusion.

tests/unit/utils/test_txn_parser.py
```diff
@@ -0,0 +1,476 @@
+import copy
+from unittest import TestCase
+
+import xrpl.utils
+
+TEST_TXN = {
+    "engine_result": "tesSUCCESS",
+    "engine_result_code": 0,
+    "engine_result_message": """
```
mvadari, line 9: Why does this have a `.strip()` instead of having the string just be there as is?

tests/unit/utils/test_txn_parser.py
```diff
@@ -0,0 +1,476 @@
...
+                        "LowLimit": {
+                            "currency": "USD",
+                            "issuer": "rpXhhWmCvDwkzNtRbm7mmD1vZqdfatQNEe",
+                            "value": "1000000000",
+                        },
+                        "LowNode": "0",
+                    },
+                    "LedgerEntryType": "RippleState",
+                    "LedgerIndex": """
+0B0388546E94E2B339A52C70CE9B7791FDF104F114C94CEACB2A9440819DD435
```
mvadari, line 40: ditto

tests/unit/utils/test_txn_parser.py
```diff
@@ -0,0 +1,455 @@
+import copy
+from unittest import TestCase
+
+import xrpl.utils
+
+TEST_TXN = {
+    "engine_result": "tesSUCCESS",
+    "engine_result_code": 0,
+    "engine_result_message": """The transaction was applied. Only final in a validated ledger.
+""",
```
mvadari, line 10:
```suggestion
    "engine_result_message": "The transaction was applied. Only final in a validated ledger.",
```
ditto for the others

xrpl/utils/txn_parser/balance_changes.py
```diff
@@ -0,0 +1,137 @@
...
+    by the transaction before it occurred.
+
+    Args:
+        transaction (Union[RawTxnType, SubscriptionRawTxnType]):
+            Raw transaction data including the account that
+            sent the transaction and the affected nodes.
+
+    Returns:
+        Dict[str, Any]:
+            All previous balances.
```
mvadari, line 35: nit: this should just be
```suggestion
    Returns:
        All previous balances.
```
ditto with the other docstrings

xrpl/utils/txn_parser/balance_changes.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def parse_previous_balances(
+    transaction: Union[RawTxnType, SubscriptionRawTxnType],
+) -> Dict[str, Any]:
+    """Parse the previous balances of all accounts affected
+    by the transaction before it occurred.
+
+    Args:
+        transaction (Union[RawTxnType, SubscriptionRawTxnType]):
```
mvadari, line 29: You don't need to put the types in the docstrings - they'll already be there in the generated documentation

xrpl/utils/txn_parser/balance_changes.py
```diff
@@ -0,0 +1,137 @@
...
+            Raw transaction data including the account that
+            sent the transaction and the affected nodes.
+
+    Returns:
+        Dict[str, Any]:
+            All final balances.
+    """
+    validate_transaction_fields(transaction_data=transaction)
+    if "transaction" in transaction:
+        transaction = cast(SubscriptionRawTxnType, transaction)
```
mvadari, line 114: You might be able to simplify this to:
```suggestion
    if isinstance(transaction, SubscriptionRawTxnType):
```

tests/unit/utils/test_txn_parser.py
```diff
@@ -0,0 +1,454 @@
...
+    def test_parse_balace_changes(self):
+        actual = xrpl.utils.parse_balance_changes(transaction=TEST_TXN)
+        expected = {
+            "rpXhhWmCvDwkzNtRbm7mmD1vZqdfatQNEe": [
+                {
+                    "Counterparty": "rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq",
+                    "Currency": "USD",
+                    "Value": "-0.7042071790427",
+                },
+                {"Counterparty": "", "Currency": "XRP", "Value": "1.000000"},
```
mvadari, line 289: We should keep the standard XRPL parlance, to avoid confusion - so `issuer` instead of `Counterparty`. It would also be more Pythonic/fit the library better if it's all lowercase instead of being capitalized.

mvadari: ditto with all of the other methods as well

xrpl/utils/txn_parser/utils/balance_changes_utils.py
```diff
@@ -0,0 +1,268 @@
...
+
+    Args:
+        node (NormalizedNode):
+            Normalized node.
+
+    Returns:
+        Optional[Decimal]:
+            The balance value as Decimal.
+    """
+    if node.new_fields and hasattr(node.new_fields, "Balance"):
```
mvadari, line 61:
```suggestion
    if node.new_fields is not None and hasattr(node.new_fields, "Balance"):
```

xrpl/utils/txn_parser/utils/balance_changes_utils.py
```diff
@@ -0,0 +1,268 @@
...
+        new_balance = cast(Union[AccountBalance, str], node.new_fields.Balance)
+        return _parse_value(value=new_balance)
+    if node.final_fields and hasattr(node.final_fields, "Balance"):
```
mvadari, line 64:
```suggestion
    if node.final_fields is not None and hasattr(node.final_fields, "Balance"):
```

tests/unit/utils/test_txn_parser.py
LimpidCrypto: If you parse balance changes for example, it returns the balance changes of all parties affected.
For example rPu2feBaViWGmWJhvaF5yLocTVD8FUxd2A bought 8.455426113856 CNY from RippleQK. The balance change for RippleQK looks like:
`'rPT74sUcTBTQhkHVD54WGncoqXEAMYbmH7': [{'Counterparty': 'rafjNPiCrvVniSGSj9UUpAPTvGnHsbeuKn', 'Currency': 'CNY', 'Value': '8.455426113856'}, {'Counterparty': 'rPu2feBaViWGmWJhvaF5yLocTVD8FUxd2A', 'Currency': 'CNY', 'Value': '-8.455426113856'}]}`
The counterparty for RippleQK is rPu2feBaViWGmWJhvaF5yLocTVD8FUxd2A but the account is not the issuer.
I think counterparty would be the better term :) I guess for the order book changes it would make sense.
It should be lowercase.

xrpl/utils/txn_parser/balance_changes.py
LimpidCrypto: Unfortunately that does not work.

xrpl/utils/txn_parser/balance_changes.py
```diff
@@ -0,0 +1,134 @@
...
+def parse_previous_balances(
+    transaction: Union[RawTxnType, SubscriptionRawTxnType],
+) -> Dict[str, Any]:
+    """
+    Parse the previous balances of all accounts affected
+    by the transaction before it occurred.
+
+    Args:
+        transaction: Raw transaction data including the account that
+            sent the transaction and the affected nodes.
```
mvadari, line 31:
```suggestion
            sent the transaction and the metadata.
```

xrpl/utils/txn_parser/balance_changes.py
```diff
@@ -0,0 +1,134 @@
...
+
+    balance_changes = parse_balance_changes(transaction=transaction)
+    final_balances = parse_final_balances(transaction=transaction)
+
+    for account, balances in balance_changes.items():
+        for count, balance in enumerate(balances):
+            final_balances_value = final_balances[account][count]["value"]
+            final_balances[account][count]["value"] = str(
+                Decimal(final_balances_value) - Decimal(balance["value"])
+            )
```
mvadari, line 49:
```suggestion
        for balance in balances:
            final_balances_value = balances["value"]
            balances["value"] = str(
                Decimal(final_balances_value) - Decimal(balance["value"])
            )
```

mvadari:
I know this is pretty far into the process, but could you split up this PR, with one exported method per PR? It would make it much easier to understand what's going on, because it seems like I find something I've previously missed on every review. The PRs can build on top of each other if that's easier.

LimpidCrypto:
> I know this is pretty far into the process, but could you split up this PR, with one exported method per PR? It's hard to catch everything on a 2k line PR. It would make it much easier to understand what's going on, because it seems like I find something I've previously missed on every review, and that just keeps dragging this out. The PRs can build on top of each other if that's easier.

@mvadari I can close this pull request then, right?

mvadari:
> @mvadari I can close this pull request then, right?

Yes

LimpidCrypto:
The four exported functions will gradually get their own PR

Issue #204, opened 2021-03-25
Better support for transaction flags


mDuo13:
Many transactions have specific [flags](https://xrpl.org/transaction-common-fields.html#flags-field) that affect how the transaction works. It's important to be able to set these flags when sending certain types of transactions, and when reading/processing certain types of transactions it's even more important to be able to check them to understand the functionality of the transaction (most importantly, Partial Payments, #196). Preferably, this could be done by setting and getting specific options as booleans by name.

The [map of flags to names I wrote a long time ago](https://github.com/mDuo13/txsplain/blob/master/txsplain.py#L22-L63) is still valid and, I believe, up-to-date (any new transaction types not mentioned there don't have any flags).

Furthermore, any bits in the `Flags` field of a transaction that don't represent defined flags MUST be 0. We should enforce that requirement on transactions too.

At the transaction model level, I think there is no reason to directly expose the flags field if we can expose, say, a dict of flag names to bools instead.

mvadari:
Just to clarify, this is basically what we implemented in xrpl.js, right?

PR #363, opened 2022-03-15
Add better support for txn flags

LimpidCrypto:
# Add better support for txn flags
Add support for accessing transaction flags directly from the transaction object.
Enabling to set transaction `flags` with `bools` to make it easier and more understandable for users.
Add a map of flags as suggested by @mDuo13.

Keep support of `list/int` of `flags` attribute in `Transaction`.

### Tasks:
- [x] Support Transaction.has_flag()

### Idea:
#204


LimpidCrypto:
I kept support for the `flags` attribute to make sure people’s code will still work.
Do you think it would be too confusing so it would be better to end the support of that attribute?
Or just add a note in the `...Flags` objects docstrings that they are deprecated?

xrpl/models/transactions/account_set.py
```diff
@@ -145,6 +145,60 @@ class AccountSet(Transaction):
     also set the AccountSetFlag.ASF_AUTHORIZED_MINTER flag.
     """
 
+    asf_account_tx_id: Optional[bool] = None
```
mvadari, line 148: This would be better implemented as a `TypedDict` of strings/enums to bools that is a part of `Flags`, instead of adding many additional instance variables to each class.

xrpl/asyncio/transaction/main.py
```diff
@@ -79,6 +80,11 @@ async def safe_sign_transaction(
     if check_fee:
         await _check_fee(transaction)
     transaction_json = _prepare_transaction(transaction, wallet)
+    transaction_json = {
+        field: value
+        for field, value in transaction_json.items()
+        if not field.startswith(TX_FLAG_PREFIXES)
```
JST5000, line 86: What does this do?

xrpl/models/transactions/transaction.py
```diff
@@ -336,7 +337,29 @@ def to_dict(self: Transaction) -> Dict[str, Any]:
...
+            if (attr.startswith(TX_FLAG_PREFIXES) and set_flag)
+        ]
+        if tx_flags:
+            return [TX_FLAGS[self.transaction_type][tx_flag] for tx_flag in tx_flags]
+        return []
+
     def _flags_to_int(self: Transaction) -> int:
+        tx_flags: List[int] = self._bool_flag_to_list()
+        if tx_flags and (
+            self.flags if isinstance(self.flags, list) else self.flags > 0
```
JST5000, line 353: This logic is a bit hard to read in my opinion. Maybe it'd be better to split it into a helper function with a descriptive name?

xrpl/models/transactions/transaction.py
```diff
@@ -336,7 +337,29 @@ def to_dict(self: Transaction) -> Dict[str, Any]:
...
+        ):
+            raise XRPLModelException(
+                "Please define flags either by setting bools or by using `flags` attr."
```
JST5000, line 356: nit:
```suggestion
                "Please define flags either by setting bools or by using `flags` attr, not both."
```

xrpl/models/transactions/account_set.py
LimpidCrypto: Just to clarify. You mean something like this, right?
```python
class AccountSetFlagInterface(BoolFlags):  # 'BoolFlags' == 'TypedDict'
    asf_account_tx_id: bool
    asf_authorized_minter: bool
    asf_default_ripple: bool
    asf_deposit_auth: bool
    asf_disable_master: bool
    asf_disallow_xrp: bool
    asf_global_freeze: bool
    asf_no_freeze: bool
    asf_require_auth: bool
    asf_require_dest: bool
```

xrpl/asyncio/transaction/main.py
LimpidCrypto: For now if you want to define a flag with bool you are defining an instance variable.
Because you would then have the flag name as key in the `transaction_json` I added that code to delete the flag keys.
Else you could not `encode_for_signing`.

**When I implemented the changes suggested by @mvadari that part will be omitted.**

JST5000 (approved):
Looks good to me!

JST5000:
Before merging you should add a bullet to HISTORY.md explaining this addition :)

LimpidCrypto:
> Before merging you should add a bullet to HISTORY.md explaining this addition :)

@JST5000 I would merge for the first time. Do you mean the CHANGELOG.md? The new version would be 1.5.0 right?
**The tests failed because I can not import `TypedDict` from `typing` in python 3.7.1. I will use `typing_extensions`. I have tested it and it works for all versions (3.7.1 - 3.10).**

JST5000:
> @JST5000 I would merge for the first time. Do you mean the CHANGELOG.md? The new version would be 1.5.0 right?
> **The tests failed because I can not import `TypedDict` from `typing` in python 3.7.1. I will use `typing_extensions`. I have tested it and it works for all versions (3.7.1 - 3.10).**

Ah yep, the CHANGELOG.md is correct :) - You can just add it under the "Unreleased" section with a "Added" header like in the other version notes below. (When we do a release we'll update the version from "Unreleased" to the proper version number).

JST5000:
> I kept support for the `flags` attribute to make sure people’s code will still work.
> Do you think it would be too confusing so it would be better to end the support of that attribute?
> Or just add a note in the `...Flag` (e.g. PaymentFlag) objects docstrings that they are deprecated?

Ah, sorry I didn't see this comment. 

Making it backwards compatible is good. 

I'm not opposed to people using the number format since that's the format that `rippled` uses. Having both makes it more usable in my opinion (Regardless of the format you have the information in, you can make it work without any fancy transformations)

tests/unit/models/transactions/test_better_transaction_flags.py
```diff
@@ -0,0 +1,413 @@
...
+                account=ACCOUNT,
+                destination=dest,
+                amount=amnt,
+            )
+            safe_sign_transaction(
+                transaction=pymnt,
+                wallet=WALLET,
+            )
+        except XRPLModelException:
+            self.fail("Could not sign transaction without flags")
```
JST5000, line 357: Oh, actually if no flags are defined, it should default to flags = 0 (It's an optional parameter)

LimpidCrypto: How about: "If no flags are defined 'flags' have to be 0. Txn could not be signed."?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [[Unreleased]]
 
+## [1.5.0] - 2022-03-18
+### Added
+- Support setting flags with booleans. For each transaction type supporting flags there is a `FlagInterface` to set the flags with booleans.
+
```
mvadari, line 13:
```suggestion
## [[Unreleased]]
### Added
- Support setting flags with booleans. For each transaction type supporting flags there is a `FlagInterface` to set the flags with booleans.

```

xrpl/models/transactions/offer_create.py
```diff
@@ -53,6 +54,20 @@ class OfferCreateFlag(int, Enum):
...
 
+class OfferCreateFlagInterface(FlagInterface):
+    """
+    Transactions of the OfferCreate type support additional values in the Flags field.
+    This TypedDict represents those options.
+
+    `See OfferCreate Flags <https://xrpl.org/offercreate.html#offercreate-flags>`_
+    """
+
+    tf_passive: bool
```
mvadari, line 65: IMO all these flags should be upper-case

khancode (approved):
LGTM

tests/unit/models/transactions/test_better_transaction_flags.py
JST5000: That works. Or maybe "Failed to sign a transaction which had no flags defined."

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [[Unreleased]]
+<<<<<<< HEAD
```
intelliot, line 9: You can fix this up by removing the lines containing merge conflict markers

JST5000:
@LimpidCrypto Looks great! Thanks for writing a very solid addition to the library :)

PR #366, opened 2022-03-28
feat: add sidechain-specific RPCs/sugar

mvadari:

This PR:
* Adds `federator_info` request and response objects
* Adds a sugar util method to create a cross-chain payment transaction from a normal payment transaction

### Context of Change

xrpl-py sidechain support

### Type of Change

## Test Plan

Added tests. CI passes.

### DIFF
--- a/xrpl/utils/sidechain.py
+++ b/xrpl/utils/sidechain.py
@@ -36,10 +36,10 @@ def create_cross_chain_payment(payment: Payment, dest_account: str) -> Payment:
         raise XRPLException(
             "Cannot have more than 2 memos in a cross-chain transaction."
         )
-    memos = [dest_account_memo] + memos
+    new_memos = [dest_account_memo] + memos
 
     payment_dict = payment.to_dict()
-    payment_dict["memos"] = memos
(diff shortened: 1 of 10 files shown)


xrpl/utils/sidechain.py
```diff
@@ -0,0 +1,45 @@
...
+
+def create_cross_chain_payment(payment: Payment, dest_account: str) -> Payment:
+    """
+    Creates a cross-chain payment transaction.
+
+    Args:
+        payment: The initial payment transaction. If the transaction is signed, then
+            it will need to be re-signed. There must be no more than 2 memos, since one
+            memo is used for the sidechain destination account. The destination must be
+            the sidechain's door account.
```
JST5000, line 18:
```suggestion
        payment: The initial payment transaction. If the transaction is signed, then
            it will need to be re-signed. There must be no more than 2 memos, since the
            first memo is used for the sidechain destination account. The destination
            must be the sidechain's door account.
```

JST5000 (approved):
LGTM

khancode (approved):
LGTM 👌

Issue #352, opened 2022-01-31
Adding the unit to the docs for `ledger.get_fee`


JST5000:
It's be good to add units to the `ledger.get_fee` function so folks know they are returning `drops`. It's also be nice to link to a description of what drops are.

khancode:
Closing since this has been added in the linked PR.

Issue #330, opened 2021-12-11
Use better fee-calculation formula


mvadari:
#328 adds the ability to scale fees based on ledger load. However, it gives a fee that's way larger than it needs to be.

Wietse has a better formula: https://gist.github.com/WietseWind/3e9f9339f37a5881978a9661f49b0e52

mvadari:
Closed by #370

PR #370, opened 2022-04-05
Better fee calculation

LimpidCrypto:

Use of better [`fee calculation formula`](https://gist.github.com/WietseWind/3e9f9339f37a5881978a9661f49b0e52) but keep supporting the current. The current way is marked as deprecated.
Add info to the docstring that `get_fee` is returning drops.

### Context of Change

As proposed in #330 and #352 

### Type of Change
- [x]  Bug fix (non-breaking change which fixes an issue)
- [x]  Documentation Updates

### Test Plan
None


LimpidCrypto:
If this PR should be merged should I add the changes as 'Added' or 'Fixed' in the `CHANGELOG.md`?

xrpl/asyncio/ledger/main.py
```diff
@@ -49,22 +51,32 @@ async def get_latest_open_ledger_sequence(client: Client) -> int:
 
 
 async def get_fee(
-    client: Client, *, max_fee: Optional[float] = 2, fee_type: str = "open"
+    client: Client,
+    *,
+    max_fee: Optional[float] = 2,
+    fee_type: Optional[Literal["open", "minimum"]] = None,
```
JST5000, line 57: This should default to "open" to be backwards compatible

xrpl/asyncio/ledger/main.py
```diff
@@ -49,22 +51,32 @@ async def get_latest_open_ledger_sequence(client: Client) -> int:
...
-        fee_type: The type of fee to return. The options are "open" (the load-scaled
+        fee_type: DEPRECATED.
+            The type of fee to return. The options are "open" (the load-scaled
             fee to get into the open ledger) or "minimum" (the minimum transaction
-            fee). The default is "open".
+            fee). The default is `None`.
+
+            Recommended: Do not define any type of return so the
+            fee is calculated more dynamically based on the queue size of the
+            node. It increases the chances for the transaction to succeed.
```
JST5000, line 75: Maybe it'd be better to have another option "dynamic" since "open" is the original default? (That way you don't have to use `None` to specify unique behavior, it could just be another mode)

xrpl/asyncio/ledger/main.py
```diff
@@ -49,22 +51,32 @@ async def get_latest_open_ledger_sequence(client: Client) -> int:
...
 
     Returns:
         The transaction fee, in drops.
+        `Read more about drops <https://xrpl.org/currency-formats.html#xrp-amounts>`_
```
JST5000, line 79: What does the backtick quotes and underscore do here? (I haven't seen that syntax before)

xrpl/asyncio/ledger/main.py
```diff
@@ -74,18 +86,66 @@ async def get_fee(
...
+        if fee_type == "open":
+            fee = cast(str, drops["open_ledger_fee"])
+        elif fee_type == "minimum":
+            fee = cast(str, drops["minimum_fee"])
+        else:
+            raise XRPLException(
+                f'`fee_type` param must be "open" or "minimum". {fee_type} is not a '
+                "valid option."
+            )
+    else:  # use https://gist.github.com/WietseWind/3e9f9339f37a5881978a9661f49b0e52
```
JST5000, line 101: This credit was a little confusing, maybe having it on a new line with a bit more context like:
```
# Code originally written here: https://gist.github.com/WietseWind/3e9f9339f37a5881978a9661f49b0e52
```

xrpl/asyncio/ledger/main.py
LimpidCrypto: Example: `See Payment Flags <https://xrpl.org/payment.html#payment-flags>`_ in xrpl/models/transactions/payment.py

xrpl/asyncio/ledger/main.py
```diff
@@ -74,18 +86,66 @@ async def get_fee(
...
+            ),
+        )
+        if queue_pct > 0.1:  # if 'current_queue_size' is >10 % of 'max_queue_size'
+            possible_fee_medium = round(
+                (minimum_fee + median_fee + open_ledger_fee) / 3
+            )
+        elif queue_pct == 0:  # if 'current_queue_size' is 0
+            possible_fee_medium = max(
+                10 * minimum_fee, min(minimum_fee, open_ledger_fee)
+            )
```
JST5000, line 122: This seems like it would always produce either `10 * minimum_fee` or `open_ledger_fee`. I'm not sure why the `min(minimum_fee, ...)` is required.

xrpl/asyncio/ledger/main.py
```diff
@@ -74,18 +86,66 @@ async def get_fee(
...
+                f'`fee_type` param must be "open" or "minimum". {fee_type} is not a '
+                "valid option."
+            )
+    else:  # use https://gist.github.com/WietseWind/3e9f9339f37a5881978a9661f49b0e52
+        current_queue_size = int(result["current_queue_size"])
+        max_queue_size = int(result["max_queue_size"])
+        queue_pct = current_queue_size / max_queue_size
+        minimum_fee = int(drops["minimum_fee"])
+        median_fee = int(drops["median_fee"])
+        open_ledger_fee = int(drops["open_ledger_fee"])
```
JST5000, line 107: If we extract this "else" section into it's own function with these as parameters, it'll let you write unit tests for this logic :) 

Some basic things that would be good to test:
* When given `queue_pct = .5` - Is that fee more than when given `queue_pct = 0`
* Similarly when `queue_pct = 1` is that fee more than `queue_pct = .9`

LimpidCrypto: Good idea 👍 I don't think it would make sense to add it to the async folder. I would add it to utils. What do you think? :)

xrpl/asyncio/ledger/main.py
LimpidCrypto: I will look into it :)

LimpidCrypto: You are right, just a little misconception. I will fix that.

xrpl/asyncio/ledger/main.py
mvadari: It's a link in rst syntax

xrpl/asyncio/ledger/main.py
mvadari, line 101:
```suggestion
    if fee_type == "open":
        fee = cast(str, drops["open_ledger_fee"])
    elif fee_type == "minimum":
        fee = cast(str, drops["minimum_fee"])
    elif fee_type is not None:
        raise XRPLException(
            f'`fee_type` param must be "open" or "minimum". {fee_type} is not a '
            "valid option."
        )
    else:  # use https://gist.github.com/WietseWind/3e9f9339f37a5881978a9661f49b0e52
```

xrpl/asyncio/ledger/main.py
mvadari: It should still be in this folder, because it's only a helper function used here, not a standalone method.

LimpidCrypto:
I am a little confused how the integration tests work. Every rippled api request you make is basically made via localhost, right? Why does the `result` of the `fee` method not include "max_queue_size"?
@JST5000 why does the `fee_type` has to be "open" as default to be backwards compatible? In my opinion the default `fee_type` should be dynamically calculated fee.

Thank you :) You guys already taught me so much 🙏

JST5000:
> @JST5000 why does the `fee_type` has to be "open" as default to be backwards compatible? In my opinion the default `fee_type` should be dynamically calculated fee.

My main worry with that is if updating xrpl.js version changes how much money people spend on fees without any warning. While I do think the dynamic calculation is better, I don't want to increase how much people are paying without their explicit opt-in. So my thought was that it'd be backwards compatible to keep the calculations the same unless folks opt-in to the better calculations. 

That being said, I haven't done the math to see if the new calculations are higher than the previous calculations. If they're always lower when the queue is empty, then I think it makes sense to just use the better fee calculation as the default.

LimpidCrypto:
> > @JST5000 why does the `fee_type` has to be "open" as default to be backwards compatible? In my opinion the default `fee_type` should be dynamically calculated fee.
> 
> My main worry with that is if updating xrpl.js version changes how much money people spend on fees without any warning. While I do think the dynamic calculation is better, I don't want to increase how much people are paying without their explicit opt-in. So my thought was that it'd be backwards compatible to keep the calculations the same unless folks opt-in to the better calculations.
> 
> That being said, I haven't done the math to see if the new calculations are higher than the previous calculations. If they're always lower when the queue is empty, then I think it makes sense to just use the better fee calculation as the default.

Ah I did not think about that. I agree to you, the "dynamic" option should not be the default to prevent unwanted higher fees :)

JST5000 (approved):
LGTM :)

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support setting flags with booleans. For each transaction type supporting flags there is a `FlagInterface` to set the flags with booleans.
 - `federator_info` RPC support
 - Helper method for creating a cross-chain payment to/from a sidechain
+- Helper function for dynamic fee calculation
```
mvadari, line 13: nit:
```suggestion
- Support for dynamic fee calculation
```

xrpl/ledger/main.py
```diff
@@ -40,24 +40,30 @@ def get_latest_open_ledger_sequence(client: SyncClient) -> int:
...
         client: the network client used to make network calls.
         max_fee: The maximum fee in XRP that the user wants to pay. If load gets too
             high, then the fees will not scale past the maximum fee. If None, there is
             no ceiling for the fee. The default is 2 XRP.
         fee_type: The type of fee to return. The options are "open" (the load-scaled
-            fee to get into the open ledger) or "minimum" (the minimum transaction
-            cost). The default is "open".
+            fee to get into the open ledger), "minimum" (the minimum transaction
+            fee) or "dynamic" (dynamic fee-calculation based on the queue size
+            of the node). The default is "open".
```
mvadari, line 59: ditto here

xrpl/asyncio/ledger/utils.py
```diff
@@ -0,0 +1,69 @@
...
+            fee_low * 15,
+            10000,
+        ),
+    )
+    fee_high = round(
+        min(
+            max(10 * minimum_fee, round(max(median_fee, open_ledger_fee) * 1.1)),
+            100000,
+        ),
+    )
```
mvadari, line 60: ditto

mvadari (approved):
Left a couple notes on where to leave some additional comments, looks good otherwise

CHANGELOG.md
```diff
@@ -10,8 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support setting flags with booleans. For each transaction type supporting flags there is a `FlagInterface` to set the flags with booleans.
 - `federator_info` RPC support
 - Helper method for creating a cross-chain payment to/from a sidechain
+- Support for dynamic fee calculation
 - Helper method for parsing an NFTokenID
 
+
```
mvadari, line 16: super nit: this line should probs be removed

mvadari (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support setting flags with booleans. For each transaction type supporting flags there is a `FlagInterface` to set the flags with booleans.
 - `federator_info` RPC support
 - Helper method for creating a cross-chain payment to/from a sidechain
+- Support for dynamic fee calculation
```
mvadari, line 15: Given the new released version, this should be moved up into the `Unreleased` section

PR #372, opened 2022-04-07
fix: allow lower-case currency codes

mvadari:

According to [the xrpl.org page on currency codes](https://xrpl.org/currency-formats.html#standard-currency-codes), we should support lower-case currency codes.

### Context of Change

Issue in the launch kit where lower-case currency codes were throwing an error.

### Type of Change

## Test Plan

Relevant test was fixed.


JST5000 (approved):
LGTM

khancode (approved):

PR #374, opened 2022-04-12
Add util to parse an NFTokenID

JST5000:

Add a utility method to parse NFTokenIDs into a readable format.

### Context of Change

In response to #1897 

### Type of Change

## Test Plan

Added unit tests for this new feature.

### DIFF
--- a/tests/unit/utils/test_parse_nftoken_id.py
+++ b/tests/unit/utils/test_parse_nftoken_id.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 from unittest import TestCase
 
+from xrpl import XRPLException
 from xrpl.utils.parse_nftoken_id import parse_nftoken_id
 
 
@@ -10,7 +11,7 @@ class TestParseNFTokenID(TestCase):
     """Test parse_nftoken_id."""
 
     def test_parse_nftoken_id_successful(self: TestParseNFTokenID) -> None:
-        nft_id = "000B0539C35B55AA096BA6D87A6E6C965A6534150DC56E5E12C5D09E0000000C"  # noqa:E501
+        nft_id = "000B0539C35B55AA096BA6D87A6E6C965A6534150DC56E5E12C5D09E0000000C"
         result = parse_nftoken_id(nft_id)
         expected = {
             "token_id": nft_id,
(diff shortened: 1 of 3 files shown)


mvadari (approved):

JST5000:
@mvadari @ckniffen Sorry, could you re-review this? I realized my tests weren't actually being run because I named them `_test...` instead of `test` and found some issues with my decoding. (I was decoding to strings, when I needed to decode the number values straight to numbers. (`1024 -> 0x400` instead of being encoded digit by digit like `01000204`)

ckniffen (approved):

tests/unit/utils/test_parse_nftoken_id.py
```diff
@@ -0,0 +1,27 @@
...
+from unittest import TestCase
+
+from xrpl.utils.parse_nftoken_id import parse_nftoken_id
+
+
+class TestParseNFTokenID(TestCase):
+    """Test parse_nftoken_id."""
+
+    def test_parse_nftoken_id_successful(self: TestParseNFTokenID) -> None:
+        nft_id = "000B0539C35B55AA096BA6D87A6E6C965A6534150DC56E5E12C5D09E0000000C"  # noqa:E501
```
mvadari, line 13: This shouldn't be triggering E501 - it's only 84 characters long and the limit is set to 88.

xrpl/utils/parse_nftoken_id.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Returns:
+        A decoded nft TokenID with all information encoded within
+
+    # noqa:E501
+    """
+    expected_length = 64
+    print(len(nft_id))
+    if len(nft_id) != expected_length:
+        raise RuntimeError(
```
mvadari, line 82: This should be `XrplError`

khancode (approved):

xrpl/utils/parse_nftoken_id.py
```diff
@@ -0,0 +1,100 @@
...
+    Raises:
+        XRPLException: when given an invalid Token ID as nft_id.
+
+    Returns:
+        A decoded nft TokenID with all information encoded within
+
+    # noqa:E501
+    """
+    expected_length = 64
+    print(len(nft_id))
```
mvadari, line 81: this should be removed

mvadari (approved):
lgtm other than the debug artifact

PR #379, opened 2022-04-25
Verify submitted transaction response is valid before waiting for its final outcome.

khancode:

Inside `send_reliable_submission`, verify submitted transaction response is valid before waiting for its final outcome.

### Context of Change

There's a bug in `reliable_submission` reported by users where it throws an `txnNotFound: Transaction not found exception`. From my investigation, this bug is reproduced when waiting for the final outcome on a submitted malformed transaction (`temMALFORMED`). Therefore, this fix verifies a submitted transaction is valid (not malformed) before waiting for its final outcome inside `send_reliable_submission`.

### Type of Change

## Test Plan

Integration and unit tests pass.

### DIFF
--- a/xrpl/asyncio/transaction/reliable_submission.py
+++ b/xrpl/asyncio/transaction/reliable_submission.py
@@ -89,6 +89,8 @@ async def send_reliable_submission(
     transaction_hash = transaction.get_hash()
     submit_response = await submit_transaction(transaction, client)
     prelim_result = submit_response.result["engine_result"]
+    if prelim_result != 'tesSUCCESS':
+        raise Exception(submit_response.result["engine_result_message"])
 
     return await _wait_for_final_transaction_outcome(
         transaction_hash, client, prelim_result


xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -89,6 +89,8 @@ async def send_reliable_submission(
     transaction_hash = transaction.get_hash()
     submit_response = await submit_transaction(transaction, client)
     prelim_result = submit_response.result["engine_result"]
+    if prelim_result != 'tesSUCCESS':
```
mvadari: This should only check for `temMALFORMED` or any of the error codes that don't propagate across the peer network. There are some "error" responses that eventually "go away" and the transaction becomes valid - such as `telINSUF_FEE_P`.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -89,6 +90,8 @@ async def send_reliable_submission(
     transaction_hash = transaction.get_hash()
     submit_response = await submit_transaction(transaction, client)
     prelim_result = submit_response.result["engine_result"]
+    if prelim_result != "tesSUCCESS":
```
mvadari, line 93: This should only check for `temMALFORMED` or any of the error codes that don't propagate across the peer network. There are some "error" responses that eventually "go away" and the transaction becomes valid - such as `telINSUF_FEE_P`.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -89,6 +90,8 @@ async def send_reliable_submission(
...
+        raise Exception(submit_response.result["engine_result_message"])
```
mvadari, line 94:
```suggestion
        raise XRPLReliableSubmissionException(submit_response.result["engine_result_message"])
```

mvadari, line 94: Error messages can also occur in other places in the response (this is a processing example: https://github.com/XRPLF/xrpl-py/blob/7572e939e271fa838057f6d2cc529200379af9ec/xrpl/asyncio/clients/exceptions.py#L19)

xrpl/asyncio/transaction/reliable_submission.py
khancode: Ah I see. I'll change this to only checking for [`tem` codes](https://xrpl.org/tem-codes.html#tem-codes) which relate to transaction malformed.

JST5000 (approved):
LGTM

mvadari (approved):
Don't forget to edit the changelog

ckniffen (approved):

Issue #298, opened 2021-11-08
Utility function to get balance changes from transaction metadata


mDuo13:
The [`getBalanceChanges()` method of xrpl.js](https://js.xrpl.org/modules.html#getBalanceChanges) makes it much easier for beginners to reliably get information about how a transaction affected their balances of XRP and tokens. Doing it "by hand" is error-prone and confusing since it involves parsing a bunch of metadata, which has some sneaky edge cases, and `RippleState` nodes in particular, where the sign of the balance depends on which accounts are the high/low nodes. 

It would be great to have a comparable function in xrpl-py. 

It would also be nice to do `getBalances()` (which calls account_info and account_lines and does a little processing on the results) as well.

JST5000:
Completed by LimpidCrypto in #383

PR #383, opened 2022-05-03
Add balance changes parser

LimpidCrypto:

The https://github.com/XRPLF/xrpl-py/pull/342 PR will be splitted in four parts. For each exported method (`parse_balance_changes`, `parse_final_balances`, `parse_previous_balances` and `parse_order_book_changes`) there will now be a separate PR to make it easier to review. This PR adds the functionality to parse the balance changes from a transaction's metadata.

### Context of Change

https://github.com/XRPLF/xrpl-py/issues/298

### Type of Change

## Test Plan

I wrote a test to ensure the parser is working well.

- [x] Add tests that should raise the `XRPLTxnFieldsException`


LimpidCrypto:
Let's start with `parse_balance_changes` :)

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,210 @@
...
+        previous_balance = cast(
+            Union[AccountBalance, str], node.previous_fields.Balance
+        )
+        value = _parse_value(value=final_balance) - _parse_value(
+            value=previous_balance,
+        )
+    if value and value != 0:
+        return value
+    else:
+        return None
```
JST5000, line 78: Why does this return `None` if value is 0?

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,196 @@
...
+    Returns:
+        Transaction data in standard format.
+    """
+    transaction = transaction_data["transaction"]
+    normalized_txn = {
+        tx_field_name: tx_field_value
+        for tx_field_name, tx_field_value in transaction.items()
+    }
+    meta = transaction_data["meta"]
+    normalized_txn["meta"] = meta
```
JST5000, line 68: nit: This should just be one line
```suggestion
    normalized_txn["meta"] = transaction_data["meta"]
```

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,196 @@
...
+    ledger_index = transaction_data["ledger_index"]
+    normalized_txn["ledger_index"] = ledger_index
```
JST5000, line 70: ditto

JST5000 (approved):
LGTM

tests/unit/utils/txn_parser/test_invalid_txn.py
```diff
@@ -0,0 +1,22 @@
...
+import xrpl.utils
+from tests.unit.utils.txn_parser.transaction import TXN
+
+meta_not_included_txn = deepcopy(TXN)
+meta_not_included_txn.pop("meta")
+no_nodes_txn = deepcopy(TXN)
+no_nodes_txn["meta"].clear()
+
+
+class TestInvalidTxn(TestCase):
```
mvadari, line 15: Tests that are testing the same method should be in the same file. If you want, you can still have both classes in that file, instead of combining them into one class.

tests/unit/utils/txn_parser/test_parse_balance_changes.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+from unittest import TestCase
+
+from tests.unit.utils.txn_parser.transaction import TXN
```
mvadari, line 5: nit: this transaction should probably be named something a little more descriptive than `TXN`

tests/unit/utils/txn_parser/test_parse_balance_changes.py
```diff
@@ -0,0 +1,41 @@
...
+                    "value": "10",
+                },
+            ],
+            "rME2LXH8Che2BZRbu5LCRKWju9U3ARaEPd": [
+                {
+                    "issuer": "rLBnhMjV6ifEHYeV4gaS6jPKerZhQddFxW",
+                    "currency": "5452535259000000000000000000000000000000",
+                    "value": "-10",
+                },
+                {"issuer": "", "currency": "XRP", "value": "-0.000012"},
```
mvadari, line 38: Dictionaries with info about XRP should not have an `issuer` key. See https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/test/utils/getBalanceChanges.ts#L18 for an example

tests/unit/utils/txn_parser/test_parse_balance_changes.py
```diff
@@ -0,0 +1,41 @@
...
+            ],
+        }
+        self.assertEqual(first=actual, second=excpected)
```
mvadari, line 41: nit:
```suggestion
        self.assertEqual(actual, expected)
```

tests/unit/utils/txn_parser/transaction.py
```diff
@@ -0,0 +1,170 @@
+TXN = {
```
mvadari, line 1: This should just be a JSON file.

tests/unit/utils/txn_parser/test_parse_balance_changes.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+from unittest import TestCase
+
+from tests.unit.utils.txn_parser.transaction import TXN
+from xrpl.utils import parse_balance_changes
+
+
+class TestParseBalanceChanges(TestCase):
```
mvadari, line 9: Can you add a test case for the subscription version of a tx JSON?

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+    meta = transaction_data["meta"]
+    if "AffectedNodes" not in meta or not meta["AffectedNodes"]:
+        raise XRPLTxnFieldsException("Malformed transaction fields: No nodes provided.")
+
+
+def normalize_transaction(
+    transaction_data: SubscriptionRawTxnType,
+) -> RawTxnType:
+    """
+    Formats the raw transaction data into one standard format.
```
mvadari, line 43:
```suggestion
    Formats the raw transaction data from a `transaction` subscription to match what is returned from a `tx` request.
```
Feel free to change the wording of this

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+        transaction_data: The raw transaction data.
+
+    Returns:
+        Transaction data in standard format.
+    """
+    transaction = transaction_data["transaction"]
+    normalized_txn = {
+        tx_field_name: tx_field_value
+        for tx_field_name, tx_field_value in transaction.items()
+    }
```
mvadari, line 55: normalized_txn = {
  **transaction,
  "meta": transaction_data["meta"],
  "ledger_index": transaction_data["ledger_index"]
}

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+    normalized_txn["meta"] = transaction_data["meta"]
+    normalized_txn["ledger_index"] = transaction_data["ledger_index"]
```
mvadari, line 57:
```suggestion
    normalized_txn = {
        **transaction,
        "meta": transaction_data["meta"],
        "ledger_index": transaction_data["ledger_index"]
    }
```

mvadari: This reformatting doesn't seem to actually match the types that are defined in the other file.

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+    return cast(RawTxnType, normalized_txn)
```
mvadari, line 58: This `cast` doesn't make sense - `normalized_txn` doesn't look like it's in the shape of `RawTxnType`

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+    Normalize 'NewFields', 'FinalFields' and 'ModifiedFields'.
+
+    Args:
+        fields: A dictionary of all fields that field state has.
+            E.g.: {'Balance': '62537659', 'OwnerCount': 16, 'Sequence': 67702065}
+
+    Returns:
+        The full field state with all its fields as objects.
+    """
+    balance = fields["Balance"] if "Balance" in fields else None
```
mvadari, line 74:
```suggestion
    balance = fields.get("Balance")
```
[`get` is a dictionary method](https://www.w3schools.com/python/ref_dictionary_get.asp) that returns the value if the key exists, and otherwise returns `None`

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+    flags = cast(int, fields["Flags"]) if "Flags" in fields else None
+    expiration = cast(int, fields["Expiration"]) if "Expiration" in fields else None
+    field_state = NormalizedFields(
+        Balance=AccountBalance(
+            issuer=balance["issuer"],
+            currency=balance["currency"],
+            value=balance["value"],
+        )
+        if isinstance(balance, dict)
+        else balance,
```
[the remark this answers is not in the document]
mvadari: ditto for the rest of these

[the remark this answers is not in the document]
mvadari: Might also be useful to create a small helper function for this, since you do it so much

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+    normalized_txn = {
+        tx_field_name: tx_field_value
+        for tx_field_name, tx_field_value in transaction.items()
+    }
+    normalized_txn["meta"] = transaction_data["meta"]
+    normalized_txn["ledger_index"] = transaction_data["ledger_index"]
+    return cast(RawTxnType, normalized_txn)
+
+
+def normalize_fields(
```
mvadari, line 61: Helper functions should be marked as private with a `_` in front

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+    Args:
+        transaction_data: Transactions raw data.
+
+    Returns:
+        The normalized nodes.
+    """
+    meta = transaction_data["meta"]
+    if isinstance(meta, dict):
+        affected_nodes = meta["AffectedNodes"]
+        if not affected_nodes:
```
mvadari, line 179:
```suggestion
        if affected_nodes is None:
```

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+            return []
+        return map_(affected_nodes, normalize_node)  # type: ignore
```
mvadari, line 181:
```suggestion
        return {key: normalize_node(node) for (key, node) in affected_nodes.items()}
```

xrpl/utils/txn_parser/utils/transaction.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Args:
+        affected_node: Affected node.
+
+    Returns:
+        NormalizedNode object.
+    """
+    diff_type = cast(
+        Literal["ModifiedNode", "CreatedNode", "DeletedNode"],
+        list(affected_node.keys())[0],
```
mvadari, line 139: What are you trying to get here? Why only the first one?

LimpidCrypto: This is getting the type of the affected node (ModifiedNode, CreatedNode, DeletedNode). The types are relevant for the `parse_order_book_changes` parser to determine whether an Offer was created, canceled, filled or partially-filled.
I made it an argument so it gets explained in the docstring :) To put in the arguments please check out lines 157 - 171 of my next commit. Maybe you know a more elegant solution.

xrpl/utils/txn_parser/utils/transaction.py
LimpidCrypto: The `normalized_txn` has the same format as the `RawTxnType`. The only difference is that the `normalized_txn` has the `owner_funds` field.

LimpidCrypto:
This is the closest as I could get to the JS library. I used the same [tests](https://github.com/XRPLF/xrpl.js/blob/0486037ee420e1649245183c655ea2dd3809214f/packages/xrpl/test/utils/getBalanceChanges.ts) that the JS library uses, with the same expected results.

.vscode/settings.json
```diff
@@ -4,5 +4,14 @@
...
+    "esbonio.sphinx.confDir": "${workspaceFolder}/docs",
+    "python.testing.unittestArgs": [
+        "-v",
+        "-s",
+        "./tests",
+        "-p",
+        "test_*.py"
+    ],
+    "python.testing.pytestEnabled": false,
+    "python.testing.unittestEnabled": true
```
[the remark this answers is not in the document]
LimpidCrypto: This gets created automatically, I forgot to delete it

tests/unit/utils/txn_parser/transaction_jsons/payment_token.json
```diff
@@ -0,0 +1,123 @@
+{
```
mvadari, line 1: nit: I'd call this `payment_iou.json`, since we mainly call tokens IOUs internally to the repo

tests/unit/utils/txn_parser/transaction_jsons/payment_token_multipath.json
```diff
@@ -0,0 +1,285 @@
+{
```
mvadari, line 1: ditto

xrpl/models/transactions/__init__.py
```diff
@@ -16,6 +16,12 @@
 from xrpl.models.transactions.escrow_cancel import EscrowCancel
 from xrpl.models.transactions.escrow_create import EscrowCreate
 from xrpl.models.transactions.escrow_finish import EscrowFinish
+from xrpl.models.transactions.metadata import (
+    CreatedNode,
+    DeletedNode,
+    ModifiedNode,
```
mvadari, line 22: IMO these don't need to be exported at the transaction level. You can still import them from `xrpl.models.transactions.metadata`

xrpl/utils/txn_parser/get_balance_changes.py
```diff
@@ -0,0 +1,31 @@
...
+    Parse all balance changes from a transaction's metadata.
+
+    Args:
+        metadata: Transactions metadata.
+
+    Returns:
+        All balance changes caused by a transaction.
+        The balance changes are grouped by the affected account addresses.
+    """
+    quantities = map(lambda node: get_quantites(node), normalize_nodes(metadata))
```
mvadari, line 29: This might be a little more readable like this:
```suggestion
    quantities = [get_quantities(node) for node in normalize_nodes(metadata)]
```

xrpl/utils/txn_parser/utils/get_balance_changes.py
```diff
@@ -0,0 +1,147 @@
...
+        balance=BalanceType(
+            currency=fields["Balance"]["currency"],  # type: ignore
+            issuer=fields["HighLimit"]["issuer"],
+            value=f"{value.normalize():f}",
+        ),
+    )
+    return [result, _flip_trustline_perspective(result)]
+
+
+def get_quantites(node: NormalizedNode) -> List[BalanceChangeType]:
```
mvadari, line 101: nit: should be `get_quantities`

xrpl/utils/txn_parser/utils/get_balance_changes.py
```diff
@@ -0,0 +1,147 @@
+"""Helper functions for `get_balance_changes`."""
```
mvadari, line 1: If these helper functions are only being used by `get_balance_changes`, and you won't need them for the other methods, you could put them in the same `get_balance_changes.py` file above as the main function

xrpl/utils/txn_parser/utils/nodes.py
```diff
@@ -0,0 +1,56 @@
...
+class Fields(TypedDict):
+    """Model for possible fields."""
+
+    Account: str
+    Balance: Union[BalanceType, str]
+    LowLimit: BalanceType
+    HighLimit: BalanceType
+
+
+class NormalizedNode(TypedDict):
```
mvadari, line 20: Some of these fields need to be optional, since not all of the node types have them.

xrpl/utils/txn_parser/utils/nodes.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Normalize all nodes of a transaction's metadata.
+
+    Args:
+        metadata: The transaction's metadata.
+
+    Returns:
+        The normalized nodes.
+    """
+    if len(metadata["AffectedNodes"]) == 0:
```
mvadari, line 54: This extra check isn't needed - the list comprehension will take care of it

xrpl/utils/txn_parser/utils/nodes.py
```diff
@@ -0,0 +1,56 @@
...
+    """A model representing an affected node in a standard format."""
+
+    NodeType: Literal["CreatedNode", "ModifiedNode", "DeletedNode"]
+    LedgerEntryType: str
+    LedgerIndex: str
+    NewFields: Fields
+    FinalFields: Fields
+    PreviousFields: Fields
+    PreviousTxnID: str
+    PreviouTxnLgrSeq: int
```
mvadari, line 30:
```suggestion
    PreviousTxnLgrSeq: int
```

xrpl/utils/txn_parser/utils/nodes.py
```diff
@@ -0,0 +1,56 @@
...
+from typing_extensions import Literal, TypedDict
+
+from xrpl.models import CreatedNode, DeletedNode, ModifiedNode, TransactionMetadata
+from xrpl.utils.txn_parser.utils.types import BalanceType
+
+
+class Fields(TypedDict):
+    """Model for possible fields."""
+
+    Account: str
```
mvadari, line 14: ditto here

xrpl/utils/txn_parser/utils/nodes.py
```diff
@@ -0,0 +1,56 @@
...
+
+def _normalize_node(
+    affected_node: Union[CreatedNode, ModifiedNode, DeletedNode]
+) -> NormalizedNode:
+    diff_type = cast(
+        Literal["CreatedNode", "ModifiedNode", "DeletedNode"],
+        list(affected_node.keys())[0],
+    )
+    node = affected_node[diff_type]  # type: ignore
+    return NormalizedNode(NodeType=diff_type, **node)  # type: ignore
```
mvadari, line 41: `NormalizedNode` is a `TypedDict` - it doesn't have a constructor. This should be something like:
```return {"NodeType": diff_type, **node}

**Comment by LimpidCrypto** [comment]
I don't know if it makes so much sense to make some keys in the `TypedDicts` `Optional`. All keys in a `TypedDict` that are not defined are not included (all keys are optional). If I define an `Optional` for every optional key, I must add many `assert $ is not None` as I did in `_compute_balance_change` or I have to add many `type: ignore`. Maybe you know a better solution? :)

File: xrpl/utils/txn_parser/get_balance_changes.py [hunk 32]
```diff
@@ -0,0 +1,75 @@
...
+    Args:
+        node: The affected node.
+
+    Returns:
+        The balance change.
+    """
+    value: Optional[Decimal] = None
+    if node.get("NewFields") is not None:
+        assert node["NewFields"] is not None
+        if node["NewFields"].get("Balance"):
```
JST5000, line 38: In response to your comment above in the conversation.

As far as I understand, the issue you're running into here is that while `.get` does prove the entity exists, mypy doesn't know that `node.get("NewFields")` AND it finding a value is the same as `node["NewFields"]`.

I think you can avoid the assert if you do something like:
```
newFields = node.get("NewFields")
if(newFields is not None):
    if(newFields.get("Balance")):
...
```

That way mypy knows you've just proven that `NewFields` can't be `None`

LimpidCrypto: This is way better :)

xrpl/utils/txn_parser/get_balance_changes.py
```diff
@@ -0,0 +1,73 @@
...
+            if previous_fields_balance is not None and final_fields_balance is not None:
+                value = _get_value(final_fields_balance) - _get_value(
+                    previous_fields_balance
+                )
+    if value is None or value == Decimal(0):
+        return None
+    return value
+
+
+def get_balance_changes(metadata: TransactionMetadata) -> List[BalanceChanges]:
```
mvadari, line 57: nit: reading this file would be a little less confusing if this method was at the top and the helper methods were under it

xrpl/utils/txn_parser/get_balance_changes.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Returns:
+        All balance changes caused by a transaction.
+        The balance changes are grouped by the affected account addresses.
+    """
+    quantities = [
+        get_quantities(node, _compute_balance_change)
+        for node in normalize_nodes(metadata)
+    ]
+    flattened_quantities = list(flatten(quantities))
```
mvadari, line 72: I don't think you need to re-convert this to a list - the [documentation](https://pydash.readthedocs.io/en/latest/api.html?highlight=flatten#pydash.arrays.flatten) says that it returns a list.

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,161 @@
...
+
+from xrpl.utils.txn_parser.utils.nodes import NormalizedNode
+from xrpl.utils.txn_parser.utils.types import Balance, BalanceChange, BalanceChanges
+from xrpl.utils.xrp_conversions import drops_to_xrp
+
+
+def _get_xrp_quantity(
+    node: NormalizedNode,
+    value_parser: Callable[[NormalizedNode], Optional[Decimal]],
+) -> Optional[BalanceChange]:
```
mvadari, line 16: Is this method of this form (where it takes a method as a parameter) because you also need it for other transaction parser methods?

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,161 @@
...
+    value = value_parser(node)
+    if value is None:
+        return None
+    value_is_negative = value < Decimal(0)
```
mvadari, line 20: (This)[https://docs.python.org/3/library/decimal.html#decimal.Decimal.is_signed] might be a better way of doing that

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,161 @@
...
+    value = drops_to_xrp(str(value.copy_abs()))
+    if value_is_negative:
+        value = Decimal(f"-{value}")
+    final_fields = node.get("FinalFields")
+    new_fields = node.get("NewFields")
```
mvadari, line 25: nit: this should be right above line 36

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,161 @@
...
+        account=issuer,
+        balance=Balance(
+            currency=balance_change["balance"]["currency"],
+            issuer=balance_change["account"],
+            value=f"{negated_balance.normalize():f}",
+        ),
+    )
+
+
+def _get_trustline_quantity(
```
mvadari, line 65: It looks like it would be easier both in this function and in `get_quantities` for this method to return `[]` instead of `None` - then `get_quantities` doesn't have to have an extra check for if the result is `None`

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,161 @@
...
+    node: NormalizedNode,
+    value_parser: Callable[[NormalizedNode], Optional[Decimal]],
+) -> Optional[BalanceChange]:
+    value = value_parser(node)
+    if value is None:
+        return None
+    value_is_negative = value < Decimal(0)
+    value = drops_to_xrp(str(value.copy_abs()))
+    if value_is_negative:
+        value = Decimal(f"-{value}")
```
mvadari, line 23: What's the objective of this?

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,161 @@
...
+    grouped = group_by(balance_changes, lambda change: change["account"])
+    result = []
+    for account, balances in grouped.items():
+        balance_changes_object = BalanceChanges(
+            account=account,
+            balances=[],
+        )
+        for balance in balances:
+            balance.pop("account")
+            balance_changes_object["balances"].append(list(balance.values())[0])
```
mvadari, line 159:
```suggestion
            balance_changes_object["balances"].append(balance["balance"])
```

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+    return None
+
+
+def _flip_trustline_perspective(balance_change: BalanceChange) -> BalanceChange:
+    negated_balance = Decimal(balance_change["balance"]["value"]).copy_negate()
+    balance = balance_change.get("balance")
+    assert balance is not None
+    issuer = balance.get("issuer")
+    assert issuer is not None
```
mvadari, line 54:
```suggestion
    balance = balance_change["balance"]
    issuer = balance["issuer"]
```
The call will automatically error if it doesn't exist

xrpl/utils/txn_parser/utils/balance_parser.py
LimpidCrypto: The `drops_to_xrp` function only accepts positive values. This is supposed to negate the value again. I guess I made it a little bit more readable now.

xrpl/utils/txn_parser/utils/balance_parser.py
LimpidCrypto: Yes, `value_parser` can be the parser for balance changes or for final balances :)

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,159 @@
...
+    grouped = group_by(balance_changes, lambda change: change["account"])
+    result = []
+    for account, balances in grouped.items():
+        balance_changes_object = BalanceChanges(
+            account=account,
+            balances=[],
+        )
+        for balance in balances:
+            balance_changes_object["balances"].append(balance["balance"])
+        result.append(balance_changes_object)
```
mvadari, line 158:
```suggestion
    for account, account_balances in grouped.items():
        balances = []
        for balance in account_balances:
            balances.append(account_balances["balance"])
        balance_changes_object = BalanceChanges(
            account=account,
            balances=balances,
        )
        result.append(balance_changes_object)
```

xrpl/utils/txn_parser/get_balance_changes.py
```diff
@@ -0,0 +1,73 @@
...
+        The balance change.
+    """
+    value: Optional[Decimal] = None
+    new_fields = node.get("NewFields")
+    if new_fields is not None:
+        balance = new_fields.get("Balance")
+        if balance is not None:
+            value = _get_value(balance)
+    elif node.get("PreviousFields") is not None and node.get("FinalFields") is not None:
+        assert node["PreviousFields"] is not None and node["FinalFields"] is not None
```
JST5000, line 61: You can get rid of this assert by setting `node.get("PreviousFields")` and `node.get("FinalFields")` to variables before this condition instead of inside of it. Even if it's a case where there's only new_fields, I don't think the look up affects performance.

xrpl/utils/txn_parser/utils/balance_parser.py
JST5000: Small typo: On line 154 of this proposed change, `balances.append(account_balances["balance"]` should be `balances.append(balance["balance"])`

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,159 @@
...
+    node: NormalizedNode,
+    value_parser: Callable[[NormalizedNode], Optional[Decimal]],
+) -> List[BalanceChange]:
+    """
+    Computes the complete list of every balance that changed in the ledger
+    as a result of the given transaction.
+
+    Args:
+        node: The affected node.
+        value_parser: The needed value parser.
```
JST5000, line 75:
```suggestion
        value_parser: A function to extract a decimal value from the given node.
```

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,185 @@
...
+    result = []
+    for account, account_balances in grouped.items():
+        balances: List[Balance] = []
+        for balance in account_balances:
+            balances.append(balance["balance"])
+        balance_changes_object = BalanceChanges(
+            account=account,
+            balances=balances,
+        )
+        result.append(balance_changes_object)
```
JST5000, line 184:
```suggestion
        result.append(BalanceChanges(
            account=account,
            balances=balances,
        ))
```
Since this is a simple constructor, there's no need to create it in a separate line, plus the name `balance_changes_object` doesn't add much more info than just using the constructor.

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,185 @@
...
+        if value.is_signed()
+        else drops_to_xrp(str(absolute_value))
+    )
+    final_fields = node.get("FinalFields")
+    if final_fields is not None:
+        account = final_fields.get("Account")
+        if account is not None:
+            return BalanceChange(
+                account=account,
+                balance=Balance(  # type: ignore
```
JST5000, line 30: Why does this need a # type: ignore?

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,185 @@
...
+                    value=f"{xrp_value.normalize():f}",
+                ),
+            )
+    new_fields = node.get("NewFields")
+    if new_fields is not None:
+        account = new_fields.get("Account")
+        if account is not None:
+            return BalanceChange(
+                account=account,
+                balance=Balance(  # type: ignore
```
JST5000, line 41: Same question - Why does this need a type ignore?

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,185 @@
...
+            )
+    return None
+
+
+def _flip_trustline_perspective(balance_change: BalanceChange) -> BalanceChange:
+    balance = balance_change["balance"]
+    negated_value = Decimal(balance["value"]).copy_negate()
+    issuer = balance["issuer"]
+    return BalanceChange(
+        account=issuer,  # type: ignore
```
JST5000, line 54: Same question

xrpl/utils/txn_parser/utils/balance_parser.py
LimpidCrypto: mypy raises: Missing key "issuer" for TypedDict "Balance"

xrpl/utils/txn_parser/utils/balance_parser.py
LimpidCrypto: mypy raises: Missing key "issuer" for TypedDict "Balance"

xrpl/utils/txn_parser/utils/balance_parser.py
LimpidCrypto: `issuer` is str or None. I added `cast` and removed `# type: ignore`.

tests/unit/utils/txn_parser/test_get_balance_changes.py
```diff
@@ -0,0 +1,465 @@
...
+with open(path_to_json + "trustline_create.json", "r") as infile:
+    trustline_create = json.load(infile)
+with open(path_to_json + "trustline_delete.json", "r") as infile:
+    trustline_delete = json.load(infile)
+with open(path_to_json + "trustline_set_limit_zero.json", "r") as infile:
+    trustline_set_limit_zero = json.load(infile)
+with open(path_to_json + "trustline_set_limit.json", "r") as infile:
+    trustline_set_limit = json.load(infile)
+with open(path_to_json + "trustline_set_limit2.json", "r") as infile:
+    trustline_set_limit2 = json.load(infile)
```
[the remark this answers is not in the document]
mvadari: If it's just another similar test case, then leaving it as `trustline_set_limit2` is fine

xrpl/utils/txn_parser/utils/balance_parser.py
mvadari: It looks like the way that `TypedDict` works, you should omit the optional variables instead of marking them as `Optional`.

mvadari: I'd recommend just leaving a comment in the `Balance` `TypedDict` that there's an optional key, `issuer`

LimpidCrypto: The `Balance` `TypedDict` now looks like this:
```python
class Balance(TypedDict):
    """A account's balance model."""

    currency: str
    """The currency."""
    issuer: str
    """The issuer of the currency. This value is optional."""
    value: str
    """The amount of the currency."""
```
Now I do not have to use `cast` anymore to pass the `issuer` in the `Balance` `TypedDict`. But the two problems above still remain. mypy still raises: Missing key "issuer" for TypedDict "Balance"

mvadari: If you make it look like this:
```py
class Balance(TypedDict):
    """A account's balance model."""

    currency: str
    """The currency."""
    value: str
    """The amount of the currency."""
    # issuer: str
    # """The issuer of the currency. This value is optional."""
```
I think that should work

LimpidCrypto: I found a way that seems to work fine:
```python
Balance = TypedDict(
    "Balance", {"currency": str, "issuer": str, "value": str}, total=False
)
```

LimpidCrypto: This is the best solution I could come up with:
```python
class IssuerField(TypedDict, total=False):
    """The optional `issuer` field for an account's balance."""

    issuer: str
    """The issuer of the currency. This value is optional."""


class Balance(IssuerField):
    """A account's balance model."""

    currency: str
    """The currency code."""
    value: str
    """The amount of the currency."""
```
It works perfectly fine. Maybe `IssuerField` might not be the best name for the `TypedDict`. Please give me a short feedback :)

JST5000: One last idea is to use `NotRequired[...]` from `typing_extensions` instead of `Optional[...]` (See answer with it in: https://stackoverflow.com/questions/68814891/typeddict-does-not-allow-optional-keys)

Note: You have to use TypedDict from `typing_extensions` since this is a very new feature which isn't in normal `typing` until Python 3.11, and we support back to Python 3.7. 

If that doesn't work, I think the solution you put here works with two small changes:
1. I would add a one-line comment above the `IssuerField` explaining why it's split into two
2. Maybe `OptionalIssuer` would be a better name than `IssuerField`?

LimpidCrypto: I already tried `NotRequired`. Unfortunately it has no affect on mypy. It still raises: Missing key "issuer"...

JST5000: Ah ok, well then I think your solution is solid :)

LimpidCrypto: Do you think it would make sense to do the same for all other optional values in the TypedDicts (e.g. `NormalizedNode`, `TransactionMetadata`, `ModifiedNodeFields`)? Or do you think this is not necessary?

xrpl/utils/txn_parser/get_balance_changes.py
```diff
@@ -0,0 +1,60 @@
...
+    Args:
+        node: The affected node.
+
+    Returns:
+        The balance change.
+    """
+    value: Optional[Decimal] = None
+    new_fields = node.get("NewFields")
+    previous_fields = node.get("PreviousFields")
+    final_fields = node.get("FinalFields")
```
mvadari, line 48: If I'm understanding the flow of your code correctly, these don't need `.get` because you already used `.get` in `normalize_nodes`

xrpl/utils/txn_parser/utils/balance_parser.py
LimpidCrypto: > If you can figure out how to remove that additional `type: ignore`, that would be great

I found a way to remove the last `type: ignore`, but I find it a bit messy. Maybe you can come up with an easier way, or simplify mine.

LimpidCrypto:
Tests failed because of:
"AssertionError: XRPLRequestFailureException not raised"

xrpl/utils/txn_parser/utils/nodes.py
```diff
@@ -0,0 +1,92 @@
...
+    elif diff_type == "ModifiedNode":
+        modified_node = cast(ModifiedNode, affected_node)["ModifiedNode"]
+    else:
+        deleted_node = cast(DeletedNode, affected_node)["DeletedNode"]
+    node = (
+        created_node
+        if created_node is not None
+        else modified_node
+        if modified_node is not None
+        else deleted_node
```
JST5000, line 61: I think this could be cleaner using if/else
```suggestion
if (created_node is not None):
    node = created_node
elif (modified_node is not None):
    node = modified_node
else: 
    node = deleted_node
```

xrpl/utils/txn_parser/utils/nodes.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+    ledger_entry_type = node["LedgerEntryType"]
+    ledger_index = node["LedgerIndex"]
+    new_fields = cast(Optional[Fields], node.get("NewFields"))
```
JST5000, line 65: Does this syntax work? (If not this seems fine)
```suggestion
    new_fields: Optional[Fields] = node.get("NewFields")
```

xrpl/utils/txn_parser/utils/nodes.py
JST5000: Actually, it's probably easier to just set `node` directly instead of setting `created_node`/`modified_node`/`deleted_node` as intermediary values.

JST5000:
> Tests failed because of: "AssertionError: XRPLRequestFailureException not raised"

The tests passed once re-ran. I copied the error down and will keep an eye out for the root cause. (Seems to be an intermittent error though, not something related to your changes)

xrpl/utils/txn_parser/utils/nodes.py
LimpidCrypto: I would have done so, but unfortunately mypy raises: 'Incompatible types in assignment (expression has type "ModifiedNodeFields", variable has type "CreatedNodeFields")'

LimpidCrypto: Your suggestion is also not valid. Same error.

xrpl/utils/txn_parser/utils/nodes.py
LimpidCrypto: Unfortunately this syntax does not work.

xrpl/utils/txn_parser/utils/nodes.py
JST5000: Ah sorry about that. Since node's type isn't inferrable from it's first declaration we have to specify the type explicitly. I tried locally and this version works:
```
    if diff_type == "CreatedNode":
        node: Union[CreatedNodeFields, ModifiedNodeFields, DeletedNodeFields] = cast(
            CreatedNode, affected_node
        )["CreatedNode"]
    elif diff_type == "ModifiedNode":
        node = cast(ModifiedNode, affected_node)["ModifiedNode"]
    else:
        node = cast(DeletedNode, affected_node)["DeletedNode"]
```

(Note that this replaces the `diff_type` if statements AND the `node = ...` statement)

xrpl/models/transactions/metadata.py
```diff
@@ -0,0 +1,73 @@
...
+
+from typing_extensions import Literal, TypedDict
+
+from xrpl.models.amounts.amount import Amount
+
+
+class Fields(TypedDict):
+    """Model for possible fields."""
+
+    Account: Optional[str]
```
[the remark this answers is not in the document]
mvadari: Is there any reason you're using `TypedDict` instead of objects here (for everything except the final return, which I agree should be a dictionary)? It might be easier to type everything.

xrpl/utils/txn_parser/utils/balance_parser.py
```diff
@@ -0,0 +1,184 @@
...
+
+    Returns:
+        The currency amount's value.
+    """
+    if isinstance(balance, str):
+        return Decimal(balance)
+    return Decimal(balance["value"])
+
+
+def get_quantities(
```
mvadari, line 135: A more descriptive name for this method might be `get_balance_changes` or `get_node_balance_changes`

xrpl/models/transactions/metadata.py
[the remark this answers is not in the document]
LimpidCrypto: So you mean making `CreatedNode`, `ModifiedNode` and `DeletedNode` a `dataclass` but keep `TransactionMetadata` as a `TypedDict`?

xrpl/utils/txn_parser/utils/nodes.py
JST5000: I don't have a preference either way. I think the TypedDicts make sense, and that using classes would also make sense. So I'm leaning towards what's easier.

xrpl/models/transactions/metadata.py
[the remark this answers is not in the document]
JST5000: I think that since there's no added functions tied to the data, it makes sense to keep them as TypedDicts.

JST5000 (approved):
Looks good to me!

Issue #353, opened 2022-02-03
Better error handling for wrong URL for client


mvadari:
If you put in the wrong URL of a node for a client, the error that it returns is super opaque (it's a JSONDecodeError). We should return a better error/error message, so that it's clear that the user entered the wrong URL for the node.

khancode:
@mvadari Can you share code to repro this?

mvadari:
Here's one example:
```
>>> client = JsonRpcClient("https://s2.ripple.com:51233")
>>> client.request(ServerInfo())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mvadari/Documents/xrpl-py/xrpl/clients/sync_client.py", line 28, in request
    return asyncio.run(self.request_impl(request))
  File "/Users/mvadari/.pyenv/versions/3.9.1/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/mvadari/.pyenv/versions/3.9.1/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/Users/mvadari/Documents/xrpl-py/xrpl/asyncio/clients/json_rpc_base.py", line 39, in request_impl
    return json_to_response(response.json())
  File "/Users/mvadari/Documents/xrpl-py/.venv/lib/python3.9/site-packages/httpx/_models.py", line 1415, in json
    return jsonlib.loads(self.text, **kwargs)
  File "/Users/mvadari/.pyenv/versions/3.9.1/lib/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/Users/mvadari/.pyenv/versions/3.9.1/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Users/mvadari/.pyenv/versions/3.9.1/lib/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

mvadari:
Resolved by #388

PR #388, opened 2022-05-11
add better error handling for invalid client URLs

khancode:

Add better error handling for invalid client URLs.

### Context of Change

If you put in the wrong URL of a node for a client, the error that it returns is super opaque (it's a `JSONDecodeError`). We should return a better error/error message, so that it's clear that the user entered the wrong URL for the node.

This change now raises an `XRPLRequestFailureException` if the response can't be JSON decoded (`JSONDecodeError`).

https://github.com/XRPLF/xrpl-py/issues/353

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added unit tests.

### DIFF
--- a/tests/unit/clients/test_json_rpc_client.py
+++ b/tests/unit/clients/test_json_rpc_client.py
@@ -2,8 +2,8 @@
 from __future__ import annotations
 
 from unittest import TestCase
-from xrpl.asyncio.clients.exceptions import XRPLRequestFailureException
 
+from xrpl.asyncio.clients.exceptions import XRPLRequestFailureException
 from xrpl.clients import JsonRpcClient
 from xrpl.models.requests import ServerInfo
 
@@ -23,4 +23,3 @@ class TestJsonRpcClient(TestCase):
         with self.assertRaises(XRPLRequestFailureException):
             client = JsonRpcClient(JSON_RPC_URL)
             client.request(ServerInfo())
-        
--- a/tests/unit/clients/test_json_rpc_client.py
+++ b/tests/unit/clients/test_json_rpc_client.py
@@ -1,0 +1,26 @@
+"""Test the json_rpc_client."""
+from __future__ import annotations
+
+from unittest import TestCase
+from xrpl.asyncio.clients.exceptions import XRPLRequestFailureException
+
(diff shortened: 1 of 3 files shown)


xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -36,4 +37,9 @@ async def request_impl(self: JsonRpcBase, request: Request) -> Response:
                 self.url,
                 json=request_to_json_rpc(request),
             )
+            if response.status_code != 200:
```
JST5000, line 40: Should we allow any `200` response code? - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses

ckniffen: Maybe 204 too.  That is very common.  It could be with the RPC api only 200s are sent.

mvadari: That might be a question for the rippled team

mvadari (approved):

xrpl/asyncio/clients/json_rpc_base.py
khancode: From Elliot's suggestion, I changed this to raise `XRPLRequestFailureException` when the response can't be JSON decoded.

mvadari (approved):

tests/unit/clients/test_json_rpc_client.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class TestJsonRpcClient(TestCase):
+    """Test json_rpc_client."""
+
+    def test_json_rpc_client_valid_url(self: TestJsonRpcClient) -> None:
+        # Valid URL
+        JSON_RPC_URL = "https://s.altnet.rippletest.net:51234/"
+        client = JsonRpcClient(JSON_RPC_URL)
+        client.request(ServerInfo())
```
JST5000, line 18: This test case seems odd to me, because if testnet goes down our CI will fail. I would probably just include the invalid URL case

JST5000 (approved):
LGTM

PR #391, opened 2022-05-18
fix bug in GenericRequest.to_dict()

khancode:

Fix bug in `GenericRequest.to_dict()`

### Context of Change

There's an initialization bug in GenericRequest where `command` param doesn't get set to `method`:
```
>>> GenericRequest(command='validator_list_sites').to_dict()
{'method': 'zzgeneric_request', 'command': 'validator_list_sites'}
```

Correct state:
```
>>> GenericRequest(command='validator_list_sites').to_dict()
{'method': 'validator_list_sites'}
```

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added unit test to check GenericRequest state after initialization.

### DIFF
--- a/tests/unit/models/requests/test_requests.py
+++ b/tests/unit/models/requests/test_requests.py
@@ -1,6 +1,6 @@
 from unittest import TestCase
 
-from xrpl.models.requests import Fee
+from xrpl.models.requests import Fee, GenericRequest
 
 
 class TestRequest(TestCase):
@@ -8,3 +8,8 @@ class TestRequest(TestCase):
         tx = Fee()
         value = tx.to_dict()["method"]
         self.assertEqual(type(value), str)
+
+    def test_generic_request_to_dict_sets_command_as_method(self):
+        command = "validator_list_sites"
+        tx = GenericRequest(command=command).to_dict()
+        self.assertDictEqual(tx, {"method": command})
--- a/xrpl/models/requests/generic_request.py
+++ b/xrpl/models/requests/generic_request.py
@@ -33,9 +33,13 @@ class GenericRequest(Request):
             kwargs: All the arguments for the request.
         """
         # initialize all the dataclass stuff
+        method = RequestMethod.GENERIC_REQUEST
+        if "command" in kwargs:
+            method = kwargs["command"]
+            del kwargs["command"]
         super().__init__(
             id=(cast(Union[str, int, None], kwargs["id"]) if "id" in kwargs else None),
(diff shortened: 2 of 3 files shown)


ckniffen (approved):

tests/unit/models/requests/test_requests.py
```diff
@@ -1,10 +1,15 @@
...
+from xrpl.models.requests import Fee, GenericRequest
 
 
 class TestRequest(TestCase):
     def test_to_dict_includes_method_as_string(self):
         tx = Fee()
         value = tx.to_dict()["method"]
         self.assertEqual(type(value), str)
+
+    def test_generic_request_to_dict_sets_command_as_method(self):
```
mvadari, line 12: I believe there's a separate file for testing `GenericRequest` - this test should be there

xrpl/models/requests/generic_request.py
```diff
@@ -33,9 +33,13 @@ def __init__(self: GenericRequest, **kwargs: Any) -> None:
             kwargs: All the arguments for the request.
         """
         # initialize all the dataclass stuff
+        method = RequestMethod.GENERIC_REQUEST
+        if "command" in kwargs:
+            method = kwargs["command"]
```
mvadari, line 38: Does this not fail typing? `command` isn't going to be a `RequestMethod`, and if `method` is stated then that's not used at all. 
I believe my original attempt to solve this problem was in `to_dict`, not in the initialization.

mvadari (changes requested):
see comments

tests/unit/models/requests/test_requests.py
khancode: @mvadari That separate file `test_generic_request` is for integration tests. I believe my fix falls under unit tests since I'm just testing model initialization. What do you think?

mvadari: That's fair. I ran them as integration tests to make sure they'd be in a rippled-accepted form, but they could in theory be moved to unit tests

xrpl/models/requests/generic_request.py
khancode: Ah I see. Do you mean in `from_dict`? I believe `to_dict` needs this logic too in order to solve this bug.

mvadari: No, I mean `to_dict` - since that's what's used by `to_xrpl` to get the JSON submitted to rippled

mvadari (approved):

PR #396, opened 2022-06-02
refactor: redo binary codec types to simplify

mvadari:

This PR:
* Fixes a typing issue with the binary parser that was caused by a circular dependency (which we now know how to resolve)
* Renames `SerializedDict` -> `STObject` and `SerializedList` -> `STArray` (to match rippled names)
* Refactors the list of types in `FieldInstance` to pull directly from `xrpl.core.binarycodec.types`, so that doesn't need to be edited every time a new type is added

There are no feature changes here; this is just an internal refactor.

### Context of Change

I was adding some new types to the binary codec and thought that needing to add them in 2 places was unnecessary. I fixed a few other things to make the binary codecs more analogous as well.

### Type of Change

## Test Plan

CI passes. Tests that test the internal objects were edited to use the new type names (the methods exposed externally have not changed at all).

### DIFF
--- a/tests/unit/core/binarycodec/fixtures/data/data-driven-tests.json
+++ b/tests/unit/core/binarycodec/fixtures/data/data-driven-tests.json
@@ -33,11 +33,11 @@
       "ordinal": 8
     },
     {
-      "name": "SerializedDict",
+      "name": "STObject",
       "ordinal": 14
     },
     {
-      "name": "SerializedList",
+      "name": "STArray",
       "ordinal": 15
     },
     {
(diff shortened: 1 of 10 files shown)


JST5000 (approved):
LGTM

ckniffen:
@mvadari Would you consider minor version bump?

ckniffen (approved):

mvadari:
> @mvadari Would you consider minor version bump?

There's literally no feature change - this PR is just refactors. No one should be using this library at the depth where I changed things.

We'll have to do a minor version bump for #397 anyways, so it can be rolled into that.

PR #402, opened 2022-06-10
Remove flaky JsonRpcClient test

JST5000:

This test has been failing intermittently, and doesn't seem very important, so I'm removing it.

### Context of Change

This test is failing intermittently and it's unclear why - running this locally a hundred times didn't cause the error to appear at all, but it has popped up in the CI a couple times. My guess is that it's due to s2 changes/status, rather than the JsonRpcClient, so I'm removing the test. 

### Type of Change


intelliot (approved):
it would be nice to add an alternative way to exercise the same code path. but sometimes a flaky test is worse than no test

khancode (approved):

mvadari:
This test is getting flakier, so probably better to remove.

PR #406, opened 2022-06-15
add ExpandedSignerList amendment support

khancode:

Adds [ExpandedSignerList](https://xrpl.org/known-amendments.html#expandedsignerlist) amendment support.

### Context of Change

- Expands the maximum signer list size to 32 entries.
- Adds an optional field `wallet_locator` to `SignerEntry`.
- [rippled PR](https://github.com/ripple/rippled/pull/4097)

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Adds unit & integration tests.

## Future Tasks
Merge this PR once the amendment goes live.


ckniffen (approved):

tests/unit/models/transactions/test_signer_list_set.py
```diff
@@ -177,3 +177,106 @@ def test_signer_entries_signer_quorum_valid(self):
             signer_entries=_SIGNER_ENTRIES_VALID,
         )
         self.assertTrue(tx.is_valid())
+
```
JST5000, line 180: I don't see a test case for more than 16 signers, but less than 32. Maybe I missed it? Otherwise, could you add it? (Should pass, just want to make sure we're testing the new functionality as well :) )

JST5000 (approved):
LGTM after adding that test case :)

tests/unit/models/transactions/test_signer_list_set.py
khancode: Ah good catch, yes I'll add this.

intelliot (approved):
- Expected error if someone tries to set too many signers (before amendment is enabled): temDISABLED
- Some test networks could activate the amendment before Mainnet
- Should be fine to merge and release at any time (so it's ready if/when the amendment activates)

tests/unit/models/transactions/test_signer_list_set.py
khancode: Just added it for above 8 and below 32 signers.

PR #398, opened 2022-06-02
Add final balances parser

LimpidCrypto:
## High Level Overview of Change

- Add  and export `get_final_balances`
- Add `_compute_final_balance`
- Export `NormalizedNode`
- Change `BalanceChange` to `ComputedBalance`
- Change `BalanceChanges` to `ComputedBalances`
- Generally renamed some terms (function names and args) that indicate to only deal with balance changes. Now indicating to deal with computed balances.

### Context of Change

The https://github.com/XRPLF/xrpl-py/pull/342 PR was splitted. For each exported function (get_balance_changes (already merged #383), get_final_balances, get_previous_balances and get_order_book_changes) there will now be a separate PR to make it easier to review. This PR adds the functionality to parse the final balances from a transaction's metadata (get_final_balances).

### Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)

## Test Plan

I added tests that are testing the same transactions as `get_balance_changes`.

PR #407, opened 2022-06-16
Get order book changes

LimpidCrypto:

- Add `get_order_book_changes`
- Add models for offer changes
- separate `get_value` and `group_by_account` to parser.py
- Add fields to `Fields`

### Context of Change

The https://github.com/XRPLF/xrpl-py/pull/342 PR was splitted. For each exported function (get_balance_changes (already merged https://github.com/XRPLF/xrpl-py/pull/383), get_final_balances (already merged https://github.com/XRPLF/xrpl-py/pull/398), get_previous_balances and get_order_book_changes) there will now be a separate PR to make it easier to review. This PR adds the functionality to parse the order book changes from a transaction's metadata (get_order_book_changes).

Reference is: https://github.com/ripple/ripple-lib-extensions/blob/master/transactionparser/src/orderbookchanges.js
The return value is structured as the `get_balance_changes` and `get_final_balances` return values.

### Type of Change

## Test Plan

- `test_offer_created`
- `test_offer_partially_filled_and_filled`
- `test_offer_cancelled`
- `test_offer_with_expiration`

## For Discussion
- [x] When calculating the delta (`_calculate_delta`) the absolute value is returned. IMO it might be easier to understand if it is returned as a signed float ("value": "60" -> "value": "-60"), even though the values only can get smaller.
- [ ] Maybe there is a better name for arg `account_objects` in `group_by_account`.

## TODO
- [x] fix linter


JST5000:
> * [ ]  When calculating the delta (`_calculate_delta`) the absolute value is returned. IMO it might be easier to understand if it is returned as a signed float ("value": "60" -> "value": "-60"), even though the values just can get smaller.
> *

What do you mean by "the values just can get smaller"? (Would it always be negative in that case?)

LimpidCrypto:
> > * [ ]   When calculating the delta (`_calculate_delta`) the absolute value is returned. IMO it might be easier to understand if it is returned as a signed float ("value": "60" -> "value": "-60"), even though the values just can get smaller.
> 
> What do you mean by "the values just can get smaller"? (Would it always be negative in that case?)

Correct. An offer offer object can only be consumed or cancelled. Basically there are four statuses:
`created` = The transaction has created a new offer object.
`partially-filled` = The transaction has consumed an offer object, for example by 50 %.
`filled` = The transaction has consumed an offer object by 100 %.
`cancelled` = The transaction has cancelled an offer with one of the statuses above.
To make it short, the taker amounts can not get raised by the owner, they can only get consumed (get smaller). Despite that it would make sense to explicitly indicate that the `value` got smaller, instead of always returning the absolute `value`, as the reference does.

JST5000:
> To make it short, the taker amounts can not get raised by the owner, they can only get consumed (get smaller). Despite that it would make sense to explicitly indicate that the `value` got smaller, instead of always returning the absolute `value`, as the reference does.

That makes sense to me - It does strike me as odd that a `delta` function which takes a final value and previous value would be absolute valued. (So, I'm happy if it just does the signed calculation, even if it happens to always be negative in this case)

JST5000:
@LimpidCrypto Sorry for the slow review time - This week has been incredibly busy with travel, and I'm out on vacation next week - I'll make sure to prioritize this when I'm back a week from Tuesday though!

LimpidCrypto:
> @LimpidCrypto Sorry for the slow review time - This week has been incredibly busy with travel, and I'm out on vacation next week - I'll make sure to prioritize this when I'm back a week from Tuesday though!

No rush. Enjoy your well earned vacation :)

tests/unit/utils/txn_parser/test_get_order_book_changes.py
```diff
@@ -0,0 +1,162 @@
...
+                            "currency": "XRP",
+                            "value": "47.504858",
+                        },
+                        "total_price": {
+                            "currency": "XDX",
+                            "issuer": "rMJAXYsbNzhwp7FfYnAsYP5ty3R9XnurPo",
+                            "value": "82335.52909",
+                        },
+                        "sequence": 70922543,
+                        "status": "cancelled",
```
JST5000, line 113: Just having a single status that's mandatory to understand the results is a bit risky since it's easy to overlook that field. (Which in this case would cause a user to interpret this as a new "Buy" offer for the amounts specified above, instead of treating this as a removal of a previous "Buy" offer)

Similar to the above comment, is there a way to convey the actual effect of a cancellation using the `quantity` and `total_price` fields?

tests/unit/utils/txn_parser/test_get_order_book_changes.py
```diff
@@ -0,0 +1,162 @@
...
+                        "direction": "buy",
+                        "quantity": {
+                            "currency": "XRP",
+                            "value": "47.504858",
+                        },
+                        "total_price": {
+                            "currency": "XDX",
+                            "issuer": "rMJAXYsbNzhwp7FfYnAsYP5ty3R9XnurPo",
+                            "value": "82335.52909",
+                        },
```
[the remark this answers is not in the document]
JST5000: (Maybe I'm just getting confused by the names, and `quantity` and `total_price` are meant to be tied directly to the Offer object instead of the net result to the order book?)

xrpl/models/transactions/metadata.py
```diff
@@ -14,6 +14,12 @@ class Fields(TypedDict, total=False):
     LowLimit: Optional[Dict[str, str]]
     HighLimit: Optional[Dict[str, str]]
     Balance: Optional[Union[Dict[str, str], str]]
+    TakerGets: Optional[Union[Dict[str, str], str]]
+    TakerPays: Optional[Union[Dict[str, str], str]]
+    Flags: int
+    Sequence: int
+    BookDirectory: str
+    Expiration: int
```
JST5000, line 22: It seems like these two (BookDirectory and Expiration) are optional from the test data provided.

xrpl/utils/txn_parser/utils/order_book_parser.py
```diff
@@ -0,0 +1,217 @@
...
+) -> str:
+    final_value = get_value(final_amount)
+    previous_value = get_value(previous_amount)
+    delta = final_value - previous_value
+    return str(delta)
+
+
+def _get_change_amount(
+    node: NormalizedNode,
+    status: Literal["created", "partially-filled", "filled", "cancelled"],
```
JST5000, line 65: You don't actually need the `status` parameter here since we know that the status is "cancelled" if `final_fields` exists (which we check anyway), and we know that the status is "created" if `new_fields` exists (which we also check). Then the other two possible statuses are never checked. 

By doing that, you can bring the if-statements one level down :)

xrpl/utils/txn_parser/utils/order_book_parser.py
```diff
@@ -0,0 +1,217 @@
...
+        previous_fields_amount = previous_fields.get(side)
+        if final_fields_amount is not None and previous_fields_amount is not None:
+            final_amount = _derive_currency_amount(final_fields_amount)
+            previous_amount = _derive_currency_amount(previous_fields_amount)
+            value = _calculate_delta(final_amount, previous_amount)
+            # final_amount is being reused because it makes
+            # it easier to return the changed amount.
+            final_amount["value"] = value
+            # From now on you could consider final_amount as changed_amount.
+            return final_amount
```
JST5000, line 93: Instead of this you can return a new object with just that field changed using the spread operator :) 

`return {**final_amount, "value": value}`

Source: https://how.wtf/spread-operator-in-python.html#object-literals

xrpl/utils/txn_parser/utils/order_book_parser.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def _get_quality(
+    taker_gets: CurrencyAmount,
+    taker_pays: CurrencyAmount,
+    book_directory: str,
+) -> str:
+    quality_hex = book_directory[-16:]  # last 16 characters of the BookDirectory
+    taker_gets_currency = taker_gets["currency"]
+    taker_pays_currency = taker_pays["currency"]
```
JST5000, line 104: nit: By moving these down to line 109, you can have them be initialized right before they are used which makes it easier to read (There's less to remember as you go through the function :) )

xrpl/utils/txn_parser/utils/order_book_parser.py
```diff
@@ -0,0 +1,217 @@
...
+    mantissa = float.fromhex(quality_hex[2:])
+    offset = int(quality_hex[:2], 16) - 100
```
JST5000, line 106: Why is this -100?

xrpl/utils/txn_parser/utils/order_book_parser.py
```diff
@@ -0,0 +1,217 @@
...
+    scientific_quality = "{:.16f}e{}".format(mantissa, offset)
+    quality = Decimal(scientific_quality)
+    if taker_gets_currency == "XRP" or taker_pays_currency == "XRP":
+        quality = quality * 1000000
+    return str(quality.normalize())
+
+
+def _get_optional_fields(
```
JST5000, line 114: It's not obvious to me from the name of this function why `ModifiedFields` would not be checked - would it be possible to add an explaining comment diving into more detail for what this function should be used for? (Or maybe update the name?)

xrpl/utils/txn_parser/utils/order_book_parser.py
```diff
@@ -0,0 +1,217 @@
...
+    # if required fields are None return None
+    if (
+        taker_gets is None
+        or taker_pays is None
+        or account is None
+        or sequence is None
+        or book_directory is None
+    ):
+        return None
+    expiration_time = _get_fields(node, "Expiration")
```
JST5000, line 156: nit: There should probably be an extra line between this `return None` and `expiration time...` since they are logically distinct. (This is a more general rule I'd apply throughout - if the code within a function can be separated without impacting each other, using a line break to indicate that can help make it more readable. Kinda like paragraphs :) )

xrpl/utils/txn_parser/utils/order_book_parser.py
```diff
@@ -0,0 +1,217 @@
...
+    offer_change = OfferChange(
+        direction=direction,
+        quantity=quantity,
+        total_price=total_price,
+        sequence=sequence,
+        status=status,
+        maker_exchange_rate=quality,
+    )
+    if expiration_time is not None:
+        offer_change["expiration_time"] = ripple_time_to_posix(expiration_time)
```
JST5000, line 173: I would avoid making this conversion from ripple time to posix time since it's not obvious that it's happening behind the scenes. (I expected it to just be a copy of the expiration time from the ripple data when reading the test results)

xrpl/utils/txn_parser/utils/types.py
```diff
@@ -39,3 +39,45 @@ class AccountBalances(TypedDict):
 
     account: str
     balances: List[Balance]
+
+
+class CurrencyAmount(Balance):
+    """A currency amunt model. Has the same fields as `Balance`"""
```
JST5000, line 45:
```suggestion
    """A currency amount model. Has the same fields as `Balance`"""
```

xrpl/utils/txn_parser/utils/parser.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Returns:
+        The grouped computed objects.
+    """
+    grouped_objects: Dict[str, Any] = {}
+    for object in account_objects:
+        account = object["account"]
+        if account not in grouped_objects:
+            grouped_objects[account] = []
+        grouped_objects[account].append(object)
```
JST5000, line 45: I just learned a better way to do this :) 

Source with more detail and other built-in ways to do this: https://stackoverflow.com/questions/12905999/how-to-create-key-or-append-an-element-to-key

```suggestion
grouped_objects.setdefault(object["account"], []).append(object)
```

tests/unit/utils/txn_parser/test_get_order_book_changes.py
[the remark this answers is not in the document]
LimpidCrypto: My reference called it `quantity` and `total_price`. When I think about it, it might be easier and less confusing when the amounts are just named `TakerGets` and `TakerPays`. Also I find the `direction` field a bit confusing. IMO it would make more sense to just return the `flags` field as it is. What do you think?
You're right the amount should always be negative if the `status` is not "created".

xrpl/utils/txn_parser/utils/order_book_parser.py
LimpidCrypto: Unfortunately your solution has typing problems because `final_amount` is a `TypedDict` (Argument 1 to "update" of "dict" has incompatible type "CurrencyAmount"; expected "Mapping[str, str]").
Alternatively I could do:
```python
change_amount = final_amount
change_amount["value"] = value
return change_amount
```

tests/unit/utils/txn_parser/test_get_order_book_changes.py
LimpidCrypto: > …. You're right the amount should always be negative if the `status` is not "created".

This should be solved with my comment above. Then you would always have the actual amount that got changed.

xrpl/utils/txn_parser/utils/order_book_parser.py
LimpidCrypto: I would like to discuss the `_get_quality` function anyways. Because when the offer gets consumed the quality can slightly differ from the quality it initially had. IMO it would be better to always calculate the quality again (`TakerPays` / `TakerGets`) instead of deriving it from the `BookDirectory`. Otherwise you would have a false exchange rate.

tests/unit/utils/txn_parser/test_get_order_book_changes.py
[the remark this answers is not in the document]
LimpidCrypto: Since the function is intended to parse order book CHANGES it might be even better to neither return `direction` or `flags` because the fields cannot change.

xrpl/utils/txn_parser/utils/order_book_parser.py
JST5000: I prefer that suggestion, although I'd recommend the name `changed_amount` instead of `change_amount`

JST5000:
I'll review this tomorrow :)

.vscode/settings.json
```diff
@@ -6,4 +6,13 @@
...
     "mypy.runUsingActiveInterpreter": true,
+    "python.testing.unittestArgs": [
+        "-v",
+        "-s",
+        "./tests",
+        "-p",
+        "test_*.py"
+    ],
+    "python.testing.pytestEnabled": false,
+    "python.testing.unittestEnabled": true,
```
JST5000, line 17: What do these settings do? (I tried googling but couldn't quite find what -v, -s, and -p do in this case)

tests/unit/utils/txn_parser/test_get_order_book_changes.py
```diff
@@ -0,0 +1,137 @@
...
+                "account": "rJHbqhp9Sea4f43RoUanrDE1gW9MymTLp9",
+                "offer_changes": [
+                    {
+                        "flags": 131072,
+                        "taker_gets": {"currency": "XRP", "value": "44.930000"},
+                        "taker_pays": {
+                            "currency": "USD",
+                            "issuer": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B",
+                            "value": "14.524821",
+                        },
```
JST5000, line 33: So to clarify, in this case, this means that the account `rJH...` is receiving 44.93 XRP in exchange for 14.5... USD?

JST5000: (Is the "taker" in this case the account in question?)

.vscode/settings.json
LimpidCrypto: Ah sorry, sometimes (can't even tell you when) they get generated automatically when I run pre-commit, but seams to be a with my configs. I will remove the lines.

tests/unit/utils/txn_parser/test_get_order_book_changes.py
LimpidCrypto: The account `rJH...` which created the offer is offering 44.93 XRP in exchange for 14.5... USD. The taker is always the one consuming an offer. In other words `taker_gets` amount is always the amount the creator is offering and the `taker_pays` amount is always the amount the creator is receiving.

JST5000: Ah, nevermind, I think taker is the person who accepts this offer, and this account is then the maker of this offer since this has the "created" status

tests/unit/utils/txn_parser/test_get_order_book_changes.py
```diff
@@ -0,0 +1,137 @@
...
+                            "value": "-62.4730283770749",
+                        },
+                        "sequence": 5931,
+                        "status": "filled",
+                        "maker_exchange_rate": "0.979999999999960",
+                    }
+                ],
+            },
+            {
+                "account": "rPu2feBaViWGmWJhvaF5yLocTVD8FUxd2A",
```
JST5000, line 69: This account is always the maker since these changes are recording only changes to the ledger object right? (If so it might make sense to call this field `maker_account`)

LimpidCrypto: Correct `account` is always the owner of the  account. `maker_account` would be a way clearer name 👍

xrpl/utils/txn_parser/utils/order_book_parser.py
```diff
@@ -0,0 +1,191 @@
...
+    taker_gets: CurrencyAmount,
+    taker_pays: CurrencyAmount,
+) -> str:
+    taker_gets_value = Decimal(taker_gets["value"])
+    taker_pays_value = Decimal(taker_pays["value"])
+    quality = taker_pays_value / taker_gets_value
+    normalized_quality = str(quality.normalize())
+    integer_digit, decimal_places = normalized_quality.split(".")
+    if len(decimal_places) > 15:
+        decimal_places = decimal_places[:15]
```
JST5000, line 102: Why is this 15 digits of precision? (I couldn't find a reference which explained the precision to use in this situation)

.vscode/settings.json
JST5000: Hahaha, vscode settings are mysterious to me as well :P - I only notice when they don't work

xrpl/utils/txn_parser/utils/order_book_parser.py
LimpidCrypto: I misinterpreted a comment on Discord. The amount of decimal places are determined based on the [TickSize](https://xrpl.org/ticksize.html) configured by the issuer. If trading an IOU against an IOU the lower TickSize applies.
Considering that it does not make sense to truncate the quality.

JST5000 (approved):
LGTM (Asked for a second reviewer to look at this next week :) )

tests/unit/utils/txn_parser/transaction_jsons/offer_partially-filled_and_filled.json
```diff
@@ -0,0 +1,391 @@
+{
```
khancode, line 1: Can you remove the hyphen from the filename and rename it to `offer_partially_filled_and_filled.json` so the casing stays consistent?

khancode (approved):
LGTM. Just left a comment related to filename casing. Great work! 🤠

LimpidCrypto:
How do I resolve the linting error 'poetry run flake8 xrpl tests --darglint-ignore-regex="^_(.*)"'?

JST5000:
> How do I resolve the linting error 'poetry run flake8 xrpl tests --darglint-ignore-regex="^_(.*)"'?

@LimpidCrypto Generally you want to run the following three commands:
1. `poetry run black .` - This will format the code according to https://pypi.org/project/black/
2. `poetry run isort .` - This sorts your imports so they're consistent
3. `poetry run flake8 .` - This reveals any other linting/style errors you may have (Each error should be searchable, and have a straightforward fix)

(The `.` in these is just saying to do it on all files, these should be run from the top level directory I believe)

LimpidCrypto:
> poetry run black . - This will format the code according to https://pypi.org/project/black/
poetry run isort . - This sorts your imports so they're consistent
poetry run flake8 . - This reveals any other linting/style errors you may have (Each error should be searchable, and have a straightforward fix)
(The . in these is just saying to do it on all files, these should be run from the top level directory I believe)
(If you open up the details of the error you mentioned above, it'll show that almost all the errors are "BLK would make changes" which means poetry run black . should fix things :) )

I tried that but still the same errors. It wonders me anyways because I did not even touch the files that cause these errors.

JST5000:
> I tried that but still the same errors. It wonders me anyways because I did not even touch the files that cause these errors.

Ok, I know the reason, but don't quite know the fix yet - It seems that the pre-commit hook is changing things back to a state that Black doesn't like automatically. (When you run `poetry run black .` it updates all the files correctly, but then once you try to commit that the `A**B` becomes `A ** B` which makes black unhappy again) 

Currently looking into why this is happening :)

JST5000:
@LimpidCrypto so seems that there's an update to either Black or flake8-black, so if you do `poetry install` it'll update those and then you can try `poetry run black .` and commit. I was able to get that working locally, so if all else fails I can push my local version, but hopefully that fixes the issue for you on your side :)

LimpidCrypto:
> @LimpidCrypto so seems that there's an update to either Black or flake8-black, so if you do `poetry install` it'll update those and then you can try `poetry run black .` and commit. I was able to get that working locally, so if all else fails I can push my local version, but hopefully that fixes the issue for you on your side :)

Great, thanks for your help @JST5000  :)

PR #408, opened 2022-06-16
fix: fix typing of `from_dict` and other classmethods

mvadari:

This PR uses `TypeVar` to make the typing of `from_dict` and other similar constructor class methods work as expected.

### Context of Change

I was getting really annoyed by it - it required a lot of unnecessary `cast`s.

### Type of Change

## Test Plan

mypy passes.

### DIFF
--- a/xrpl/models/transactions/transaction.py
+++ b/xrpl/models/transactions/transaction.py
@@ -396,7 +396,7 @@ class Transaction(BaseModel):
                     "Transaction does not include transaction_type."
                 )
             correct_type = cls.get_transaction_type(value["transaction_type"])
-            return correct_type.from_dict(value)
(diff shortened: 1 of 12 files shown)


xrpl/models/transactions/transaction.py
```diff
@@ -235,6 +235,9 @@ def to_dict(self: Signer) -> Dict[str, Any]:
         return {"signer": super().to_dict()}
 
 
+T = TypeVar("T", bound="Transaction")
```
JST5000, line 238: Should this have the same comment? (`# any type inherited from Transaction`)

JST5000 (approved):
LGTM

khancode (approved):

PR #409, opened 2022-06-16
fix: properly encode transactions in signing requests

mvadari:

This PR replaces the use of `to_dict` with `to_xrpl` in the `Sign`, `SignFor`, and `SignAndSubmit` requests (the requests that make rippled do the signing of the transaction for you). This makes the requests and the transactions actually valid.

### Context of Change

I ran into this issue when working on sidechain tooling.

### Type of Change

## Test Plan

CI passes. Works when I use it.


JST5000 (approved):
LGTM

ckniffen (approved):

PR #412, opened 2022-06-30
Update Definitions (Hooks)

dangell7:

Add Hook Definitions
Remove Duplicate Definitions
Reorder Definitions - Matches Rippled.

Because I reordered them this PR could me a nightmare to verify...

https://github.com/XRPL-Labs/xrpld-hooks/blob/e89475e76c9f633e424711bef31b48c7b6d0e51c/src/ripple/protocol/impl/SField.cpp#L148

### Context of Change

Refactor. It's now easier to visually check the definitions against rippled. This might help for the future when more amendments are added with more fields

### Type of Change

- [ X] Refactor (non-breaking change that only restructures code)

## Test Plan

If you have a definitions test then this should pass.... maybe I can write one if not?

### DIFF
--- a/xrpl/core/binarycodec/definitions/definitions.json
+++ b/xrpl/core/binarycodec/definitions/definitions.json
@@ -46,23 +46,103 @@
   },
   "FIELDS": [
     [
-      "Generic",
+      "LedgerEntry",
       {
-        "nth": 0,
+        "nth": 1,
         "isVLEncoded": false,
         "isSerialized": false,
         "isSigningField": false,
-        "type": "Unknown"
+        "type": "LedgerEntry"
       }
     ],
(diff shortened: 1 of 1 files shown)


JST5000:
@dangell7 Sorry, I'm a bit confused why you closed this PR, is it handled elsewhere?

dangell7:
You can initialize the entire definitions file from the repo above. It had issues so I had to clean those up.

After that the definitions test didn't run with success so I closed this so that it wasn't cluttering up you PR's.

I can add it back in once the tests pass.

mvadari:
@dangell7 this should solve the re-ordering: https://github.com/XRPLF/xrpl-py/pull/416

PR #415, opened 2022-07-06
feat: add support for ED25519 seeds that don't use the `sEd` prefix

mvadari:

The library currently assumes that all `ed25519` seeds use the `sEd` prefix. However, this is not necessarily true - they can also just have the `s` prefix, like the `secp256k1` seeds.

This PR adds support for those `ed25519` seeds.

### Context of Change

needed for sidechains work

### Type of Change

## Test Plan

CI passes. Added tests to test the new functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [[Unreleased]]
 ### Added:
 - Function to parse the final account balances from a transaction's metadata
+- Support for ED25519 seeds that don't use the `sEd` prefix
 
 ### Fixed:
 - Typing for factory classmethods on models
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [[Unreleased]]
 ### Added:
 - Function to parse the final account balances from a transaction's metadata
-- Support for ED25519 seeds that don't use the `sEd` prefix
+- Support for Ed25519 seeds that don't use the `sEd` prefix
 
 ### Fixed:
 - Typing for factory classmethods on models
--- a/tests/unit/core/keypairs/test_main.py
+++ b/tests/unit/core/keypairs/test_main.py
@@ -49,6 +49,19 @@ class TestMain(TestCase):
         with self.assertRaises(XRPLKeypairsException):
             keypairs.derive_keypair("sEdSKaCy2JT7JaM7v95H9SxkhP9wS2r", validator=True)
 
+    def test_derive_keypair_ed25519_different_prefix(self):
+        public, private = keypairs.derive_keypair(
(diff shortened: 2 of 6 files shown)


ckniffen (approved):

tests/unit/core/keypairs/test_main.py
```diff
@@ -49,6 +49,19 @@ def test_derive_keypair_ed25519_validator(self):
...
+        )
+        self.assertEqual(
+            public,
+            "ED6BBFC23A490D021B87D25563C15DA953A7F0F1A493DAA3767FB27F82E2F80C3D",
+        )
+        self.assertEqual(
+            private,
+            "ED644E705250E4D736875E85DD3E5FBABA4E12E004549202010228E17D3D574576",
+        )
+
```
JST5000, line 64: Can you add a test case for when a normal ED25519 key is given with the CryptoAlgorithm set to SECP256K1? (That way the exception is thrown for an invalid decoding)

xrpl/wallet/main.py
```diff
@@ -16,21 +16,29 @@ class Wallet:
     details.
     """
 
-    def __init__(self: Wallet, seed: str, sequence: int) -> None:
+    def __init__(
+        self: Wallet,
+        seed: str,
+        sequence: int,
+        *,
```
JST5000, line 23: What does this `*` mean here?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [[Unreleased]]
 ### Added:
 - Function to parse the final account balances from a transaction's metadata
+- Support for ED25519 seeds that don't use the `sEd` prefix
```
intelliot, line 11:
```suggestion
- Support for Ed25519 seeds that don't use the `sEd` prefix
```
I usually see the `d` in lowercase - presumably because it is an abbreviation for [Edwards](https://en.wikipedia.org/wiki/Edwards_curve)

xrpl/wallet/main.py
mvadari: It means that you have to specify the keyword as a part of your argument, and you can't just add another param to the list

xrpl/core/addresscodec/codec.py
```diff
@@ -84,25 +86,39 @@ def encode_seed(entropy: bytes, encoding_type: CryptoAlgorithm) -> str:
...
-        prefix = _ALGORITHM_TO_PREFIX_MAP[algorithm]
+    if algorithm is not None:
+        # check all algorithm prefixes
+        for prefix in _ALGORITHM_TO_PREFIX_MAP[algorithm]:
+            try:
+                decoded_result = _decode(seed, bytes(prefix))
+                return decoded_result, algorithm
+            except XRPLAddressCodecException:
+                # prefix is incorrect, wrong prefix
+                continue
```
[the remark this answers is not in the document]
mvadari: I looked into this, and it either requires a bunch of repeated code or makes the code way messier type-wise

JST5000 (approved):
LGTM

PR #421, opened 2022-08-02
refactor: create new abstract model for nested dict objects

mvadari:

Some objects are of the form:
```
{
  "Memo": {
    ...
  }
}
```

xrpl-py abstracts this away and creates the outer dictionary for you. There are a few objects that do this (`Memo`, `Signer`, and `SignerList`). There is a lot of copy-pasted code between these 3 objects. This PR seeks to simplify that by creating a parent object that inherits from BaseModel that handles this extra code.

### Context of Change

I have a draft PR that has a few more of these objects, and it avoids having copy-pasted code.

### Type of Change

## Test Plan

CI passes. No change in functionality.

### DIFF
--- a/xrpl/models/nested_model.py
+++ b/xrpl/models/nested_model.py
@@ -61,4 +61,6 @@ class NestedModel(BaseModel):
         Returns:
             The dictionary representation of a NestedModel.
         """
-        return {self.nested_name: super().to_dict()}
+        super_dict = super().to_dict()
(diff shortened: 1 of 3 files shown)


JST5000 (approved):
LGTM

khancode (approved):
Nice refactor!

ckniffen (approved):

PR #431, opened 2022-09-14
Add AuthAccount and refactor special models check

khancode:

Add final AMM changes for beta launch.

### Context of Change

Changes included:
- Adds `AuthAccount` model to be used inside `AMMBid`
- Refactors `_value_to_tx_json()` such that `AuthAccount` model gets correctly converted to XRPL json format. My proposed fix adds a private class property `_special_model` in `PathStep` such that `AuthAccount` isn't identified as `PathStep` model anymore.
- Refactored `Path` model by extracting `PathStep` and putting it inside its own file.
- Updated unit tests for `test_to_xrpl_paths()` to use PathStep constructor for initializing paths

### Type of Change

- [x] Release

## Test Plan

Adds unit tests for AuthAccount updates .


mvadari:
I don't like this - `_special_models` will show up in `to_dict`, which would be confusing for users

PR #443, opened 2022-09-23
docs: add xrpl.js code snippets over to xrpl-py

connorjchen:

Ported xrpl.js code snippets over to xrpl-py

### Context of Change

Code snippets are a great way for developers in the community to learn from

### Type of Change

- [x] Documentation Updates

## Test Plan

Ran each file individually **(claim_payment_channel fails because missing hashing function now)**


mvadari (commented):
I would write these with the `JsonRpcClient` (unless you want to write a snippet with each client, which seems like too much).

snippets/get_transaction.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    Async snippet that walks us through getting a transaction.
+
+    Args:
+        client: The async network client to use to send the request.
+
+    Raises:
+        Exception: if meta not included in the transaction response.
+    """
+    await client.open()
```
mvadari, line 16: If you do want to use the websocket clients, you should use `with ... as ...` syntax - the websocket client documentation explains details.

snippets/get_transaction.py
```diff
@@ -0,0 +1,49 @@
...
+    Async snippet that walks us through getting a transaction.
+
+    Args:
+        client: The async network client to use to send the request.
+
+    Raises:
+        Exception: if meta not included in the transaction response.
+    """
+    async with AsyncWebsocketClient("wss://s.altnet.rippletest.net:51233") as client:
+        await client.open()
```
mvadari, line 17: You don't need to open and close the client when you use `with ... as ...` syntax - that does it automatically for you.

snippets/get_transaction.py
```diff
@@ -1,49 +1,47 @@
 """A snippet that walks us through getting a transaction."""
-from xrpl.asyncio.clients import AsyncWebsocketClient
+from xrpl.clients import JsonRpcClient
 from xrpl.models.requests import Ledger, Tx
 
 
-async def async_get_transaction() -> None:
+def get_transaction(client: JsonRpcClient) -> None:
```
mvadari, line 6: Why did you move the client to outside the function?

mvadari: Actually, I don't know if these have to be functions at all - I think these could just be scripts.

connorjchen: If I don't make it a function, I think it tries running it when I make docs

connorjchen: Also, I can move client into function. No real reason I put it outside.

connorjchen: Actually, I think I can make it not run on make docs

docs/source/snippets.rst
```diff
@@ -0,0 +1,14 @@
...
+
+.. toctree::
+   :maxdepth: 1
+
+   get_transaction
+   partial_payment
+   paths
+   reliable_transaction_submission
+   send_escrow
+   set_regular_key
```
JST5000, line 14: Order matters here :) - We want simpler cases to the top in my opinion
```suggestion
   get_transaction
   reliable_transaction_submission
   set_regular_key
   send_escrow
   paths
   partial_payment
```

JST5000, line 14: Also, is it possible to give these more human readable names?

Ex. 
* Look up a transaction on ledger
* Send a transaction and see it get validated
* Change your password by setting a regular key
* Set up an Escrow
* Find the best path to trade with
* Handle partial payments

snippets/get_transaction.py
```diff
@@ -0,0 +1,31 @@
+"""A snippet that walks us through getting a transaction."""
```
JST5000, line 1:
```suggestion
"""How to see a validated transaction"""
```
(Or something else more active + about the lesson rather than the implementation)

snippets/get_transaction.py
```diff
@@ -0,0 +1,31 @@
...
+from xrpl.clients import JsonRpcClient
+from xrpl.models.requests import Ledger, Tx
+
+client = JsonRpcClient("https://s.altnet.rippletest.net:51234/")
```
JST5000, line 5: Can you add comments as we go explaining why we're doing things? 

Ex here: # First we connect to the test network

snippets/get_transaction.py
```diff
@@ -0,0 +1,31 @@
...
+
+ledger_request = Ledger(transactions=True, ledger_index="validated")
+ledger_response = client.request(ledger_request)
+print(ledger_response)
+
+transactions = ledger_response.result["ledger"]["transactions"]
+
+if transactions:
```
JST5000, line 13: What happens if there are no transactions? (Ex. Will often close a ledger with no transactions)
https://devnet.xrpl.org/

snippets/get_transaction.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # delivered_amount is the amount actually received by the destination account.
+    # Use this field to determine how much was delivered,
+    # regardless of whether the transaction is a partial payment.
+    # https://xrpl.org/transaction-metadata.html#delivered_amount
+    if type(tx_response.result["meta"] != "string"):
+        if "delivered_amount" in tx_response.result["meta"]:
+            print("delivered_amount:", tx_response.result["meta"]["delivered_amount"])
+        else:
+            print("delivered_amount: undefined")
```
JST5000, line 31: This all seems like a lot of extra complexity that's not required for a `get_transaction` snippet. I'd probably just remove this.

snippets/partial_payment.py
```diff
@@ -0,0 +1,88 @@
...
+from xrpl.models.transactions import Payment, PaymentFlag, TrustSet
+from xrpl.transaction import (
+    safe_sign_and_autofill_transaction,
+    send_reliable_submission,
+)
+from xrpl.wallet import generate_faucet_wallet
+
+client = JsonRpcClient("https://s.altnet.rippletest.net:51234/")
+
+# creating wallets as prerequisite
```
JST5000, line 14:
```suggestion
# Creating two wallets to send money between
```

Style nit: Comments should start with a capital letter (Ex. creating -> Creating)

docs/source/snippets.rst
```diff
@@ -0,0 +1,14 @@
+XRPL Code Snippets
+===================
+
+Code snippets to demonstrate basic usage of the xrpl-py library.
+
+.. toctree::
+   :maxdepth: 1
+
+   get_transaction
```
mvadari, line 9: Should there be links to each snippet, or should all the snippets just be on this page?

snippets/get_transaction.py
```diff
@@ -0,0 +1,29 @@
...
+# References
+# - https://xrpl.org/look-up-transaction-results.html
+# - https://xrpl.org/parallel-networks.html#parallel-networks
+# - https://xrpl.org/tx.html
+
+# Create a client to connect to the test network
+client = JsonRpcClient("https://s.altnet.rippletest.net:51234/")
+
+# Create a Ledger request and have the client call it
+ledger_request = Ledger(transactions=True, ledger_index="validated")
```
mvadari, line 14: super nit: IMO the params in this object should be flipped in order

snippets/get_transaction.py
mvadari: This specific snippet should probably be on mainnet, then - otherwise you'll mostly get nothing

mvadari (commented):
A couple of notes:
* Please add a multisign snippet (this could be in a separate PR if you'd prefer)
* Please add these as tests (this would probably best done by importing the file in a test, since the file will run when it's imported)

docs/source/snippets.rst
connorjchen: Hm yeah I think all snippets on one page might be bit better and still have tabs on left bar for it

docs/source/snippets.rst
```diff
@@ -0,0 +1,37 @@
...
+-------------------------------------
+
+.. literalinclude:: ../../snippets/get_transaction.py
+
+Send a Transaction and See if It Gets Validated
+-----------------------------------------------
+
+.. literalinclude:: ../../snippets/reliable_transaction_submission.py
+
+Change Your Password by Setting a Regular Key
```
mvadari, line 19: I don't really like the phrase "password" here - can we leave the title at `Set a Regular Key`?

snippets/set_regular_key.py
```diff
@@ -0,0 +1,57 @@
...
+)
+from xrpl.wallet import generate_faucet_wallet
+
+# References
+# - https://xrpl.org/assign-a-regular-key-pair.html#assign-a-regular-key-pair
+# - https://xrpl.org/setregularkey.html#setregularkey
+# - https://xrpl.org/change-or-remove-a-regular-key-pair.html#change-or-remove-a-regular-key-pair
+
+# Create a client to connect to the test network
+client = JsonRpcClient("https://s.altnet.rippletest.net:51234/")
```
mvadari, line 17: nit: leave off the `/` at the end

mvadari (approved):

jonathanlei (approved):

Issue #205, opened 2021-03-25
Models for ledger state objects


mDuo13:
We should have models for ledger data, including the [ledger header](https://xrpl.org/ledger-header.html) and all [types of ledger objects](https://xrpl.org/ledger-object-types.html) that can be in the ledger's state data.

This would aid reading/exploring and processing ledger data as well as transaction metadata.

intelliot:
More important than other enhancements, but not truly a bug fix

intelliot:
More important than adding response objects

intelliot:
May make sense to prioritize if/when devs request it

PR #444, opened 2022-09-26
Add Ledger Objects and `Metadata`

LimpidCrypto:

This PR adds [`Ledger Objects`](https://xrpl.org/ledger-object-types.html#ledger-object-types) to have a more pythonic, object orientated way of structuring data (the objects themselves, transactions, ...). It will also add a `Metadata` model representing a transaction's metadata. The models can be derived directly from a dictionary.
There is already a `TransactionMetadata` model (which is a bit messy), that basically also represents a transaction's metadata. The issue with that model is that it has no real checks if the fields got derived correctly, if there were required fields missing, etc. It also only predefines ledger object fields currently needed for this repository, which by far aren't all possible fields.
Having actual models for Ledger Objects and Metadata also delivers better typing of the code.

There are separate models for Ledger Objects when represented in a transaction's metadata (`MD...Fields`). The reason behind this is that nodes can have `FinalFields` or `PreviousFields` where some of the normally `REQUIRED` fields are not included. Like if an `Offer` object gets changed in the `PreviousFields` are only `TakerGets` and `TakerPays` defined, all other fields are missing there, because they are defined in `FinalFields` or a level higher. That's also why in those models all fields are `Optional`

The models are intendedly not exported to the `models/__init__.py` because models like `DepositPreauth` would collide with the transaction type.

### Context of Change

#205

### Type of Change

## Test Plan

- [x] Test `from_dict`
- [x] Test `from_xrpl`
- [x] Test `to_dict`
- [ ] Test errors

## Tasks

- [x] Add [ledger header](https://xrpl.org/ledger-header.html) models
- [ ] Remove old metadata representation as soon as there is confidence that this PR has the chance to get merged. The transaction parsers currently use the old models, those have to be replaced.
- [x] Put the json test transactions into a separate directory


xrpl/models/ledger_objects/account_root.py
```diff
@@ -0,0 +1,86 @@
...
+    wallet_locator: Optional[str] = None
+    wallet_size: Optional[int] = None
+    ledger_entry_type: LedgerEntryType = field(
+        default=LedgerEntryType.ACCOUNT_ROOT, init=False
+    )
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True)
+class MDAccountRootFields(LedgerObject):
```
mvadari, line 47: IMO the metadata version of these objects doesn't need to be included - we don't include it in the TypeScript models in xrpl.js, the metadata fields are just strings.

LimpidCrypto: So you would prefer just to have let's say a ˋAccountRootˋ model with all fields to be optional to also use them in the ˋMetadataˋ?

mvadari: The opposite, actually - see: https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/models/transactions/metadata.ts

LimpidCrypto: Seems responsible :)

mvadari (commented):
The metadata and ledger objects should be in separate PRs.

JST5000:
Sorry, somehow I completely missed this PR, I'll aim to review it by Monday :)

LimpidCrypto:
> Sorry, somehow I completely missed this PR, I'll aim to review it by Monday :)

All good, I need to find time to continue this PR first anyway :D

mvadari (commented):
Perhaps this PR should be split into two: one for the ledger objects and one for the metadata. I suspect the metadata part will involve a lot more review.

LimpidCrypto:
> Perhaps this PR should be split into two: one for the ledger objects and one for the metadata. I suspect the metadata part will involve a lot more review.

I agree 👍🏼 I will close this PR for now and hope I get to open a PR for the ledger objects this weekend :)

PR #445, opened 2022-09-27
feat: poetry script for easier test case running

connorjchen:

Enabled developers to use `poetry run test FILE PATH`,  `poetry run test_unit`,  `poetry run test_integration` and `poetry run test_coverage` for testing now

### Context of Change

Currently it’s hard to remember how to run tests, it basically requires looking up the instructions in the CONTRIBUTING.md

Instead, it’d be better to write some short poetry scripts that allow you to just do something like poetry run test and poetry run test_integration for unit and integration tests respectively.

### Type of Change

## Test Plan

Ran both commands


tests/poetry_scripts.py
```diff
@@ -0,0 +1,19 @@
+import subprocess
+
+def run_unit_tests():   
+    """
+    Run all unittests. Equivalent to:
+    `poetry run python -u -m unittest discover tests/unit`
+    """
+    subprocess.run(
```
mvadari, line 8: Is there an easier way to do this without subprocess?

mvadari: Looks like we could use a plugin: https://github.com/nat-n/poethepoet#configuring-the-plugin (there might be other options as well, this was just the one linked in the PR that you linked).
I think that this would be really useful, so a plugin would be worth it.

connorjchen: I installed it as a dev dependency and seems to work just by calling `poe test`

mvadari (commented):
Another script to add: running one specific test/test file (you just leave off the `discover`)

I recommend updating the CI scripts to use poe as well.

mvadari:
I just remembered that I added `coverage` support to xrpl-py - it would be even better if those could be incorporated into the scripts. You'll just need to modify the scripts to be something like `poetry run coverage run -m unittest discover tests/integration && poetry run coverage report -m`

Perhaps a script should run both sets of tests just for coverage purposes?

I recommend using these settings:
```
[tool.coverage.run]
branch = true
source = ["xrpl"] # this might need to be more fine-tuned depending on which set of tests

[tool.coverage.report]
show_missing = true
fail_under = 85
precision = 2
```

connorjchen:
Did not call `poe test_unit` in CI files (unit_test.yml and integration_test.yml) because must install poethepoet and seemed overkill to run every time for checks

mvadari:
> Did not call `poe test_unit` in CI files (unit_test.yml and integration_test.yml) because must install poethepoet in yml file and seemed overkill to run every time for checks

`poetry run poe test_unit` should work (the CONTRIBUTING instructions should read like that as well - the alternative won't work unless you install `poe` locally, which we don't need to do). `poetry` essentially controls its own virtual Python environment, so everything is installed in that environment.

pyproject.toml
```diff
@@ -71,3 +72,19 @@ build-backend = "poetry.core.masonry.api"
 [tool.coverage.run]
 branch = true
 source = ["xrpl"]
+
+[tool.coverage.report]
+show_missing = true
+fail_under = 85
```
mvadari, line 78: This should be left out - if you run integration tests this way, it'll fail

pyproject.toml
```diff
@@ -71,3 +72,19 @@ build-backend = "poetry.core.masonry.api"
...
+precision = 2
+
+[tool.poe.tasks]
+test_unit = "coverage run -m unittest discover tests/unit"
+test_integration = "coverage run -m unittest discover tests/integration"
+
+[tool.poe.tasks.test]
+cmd  = "python3 -m unittest ${FILE_PATHS}"
+args = [{ name = "FILE_PATHS", positional = true, multiple = true}]
```
mvadari, line 87: nice

pyproject.toml
```diff
@@ -71,3 +72,19 @@ build-backend = "poetry.core.masonry.api"
 [tool.coverage.run]
 branch = true
 source = ["xrpl"]
+
+[tool.coverage.report]
+show_missing = true
```
mvadari, line 77: I recommend adding `skip_empty = true` and `skip-covered = true` to make the report a little easier to read.

CONTRIBUTING.md
```diff
@@ -68,10 +68,18 @@ poetry run flake8 ./xrpl
...
+# Works for single or multiple unit/integration tests
+# Ex: poe test tests/unit/clients/test_json_rpc_client.py tests/integration/transactions/test_account_delete.py
+poe test FILE_PATHS
+```
+
 #### Unit Tests
 
 ```bash
-poetry run python3 -m unittest discover tests/unit
+poe test_unit
```
mvadari, line 82:
```suggestion
poetry run poe test_unit
```
Poetry is a Python package manager - it handles the package installs itself. Contributors don't need to install it themselves.

ditto with below

.github/workflows/integration_test.yml
```diff
@@ -52,5 +52,5 @@ jobs:
 
       - name: Integration test
         run: |
-          poetry run coverage run -m unittest discover tests.integration
-          poetry run coverage report -m --fail-under=70
+          poetry run poe test_integration
+          poetry run coverage report --fail-under=70
```
mvadari, line 56: Do you not need the `-m`?

CONTRIBUTING.md
```diff
@@ -68,10 +68,18 @@ poetry run flake8 ./xrpl
 
 ### Running Tests
 
+#### Individual Tests
+
+```bash
+# Works for single or multiple unit/integration tests
+# Ex: poetry run poe test tests/unit/clients/test_json_rpc_client.py tests/integration/transactions/test_account_delete.py
```
mvadari, line 75: super nit: the client tests might be moving to integration tests later, I'd pick a different test to use as your example

CONTRIBUTING.md
```diff
@@ -84,7 +92,19 @@ docker run -p 5005:5005 -p 6006:6006 -it natenichols/rippled-standalone:latest
...
+
+To see manually code coverage after running unit tests or integration tests:
+```bash
+poetry run coverage report
+```
+
+To run both unit and integration tests and see code coverage:
+```bash
+poetry run poe test_coverage
 ```
```
mvadari, line 108: nit: I'd switch the order of these two paragraphs

JST5000 (approved):
LGTM

.github/workflows/integration_test.yml
connorjchen: -m is used for show_missing = true but I believe our pyproject.toml defaults to that already. If you check the CI tests in this PR run, the missing column still exists 

`[tool.coverage.report]
show_missing = true
skip_covered = true
skip_empty = true
precision = 2`

mvadari (approved):

PR #449, opened 2022-10-05
test: Change testnet to standalone to increase speed on integration tests

connorjchen:

Changed some tests involving reliable submission to run off standalone node instead of testnet because we can close standalone ledgers faster than that of devnet. Decreased test runtimes by 50% or more.

### Context of Change

The xrpl-py integration tests currently take about 5 minutes to run. This time can be improved, while at the same time simplifying the tests.

There are currently two types of tests. 

One set runs against standalone rippled, pulled from Docker. This is most of the tests, and can run really quickly because we can close ledgers whenever necessary instead of needing to wait 4 seconds for each ledger to close, as with Testnet or Devnet.

Another set runs against Devnet. These are testing things like reliable submission, where we need ledgers to close regularly. This isn’t quite as easy to do with Python compared to JS, since it’s a little more complicated to do something like setTimeout for closing a ledger asynchronously, while the relsub method is waiting. But it’s certainly nowhere near impossible.

Getting rid of the use of Devnet in the tests would make the tests run much faster, and would also allow us to remove all the logic in the test code that is dedicated to Devnet management (e.g. calling the faucet and waiting to generate a couple of wallets).

This task is a good task to get to know the integration testing system for xrpl-py, which may be a bit confusing.

### Type of Change

## Test Plan

Ran tests, **why won't integrations tests run coverage report and close after running OK?**


tests/integration/it_utils.py
```diff
@@ -60,6 +65,20 @@
...
+    def __init__(self, timeout, client):
+        self._timeout = timeout
+        self._client = client
+        self._task = asyncio.ensure_future(self._job())
+
+    async def _job(self):
+        await asyncio.sleep(self._timeout)
+        await self._client.request(LEDGER_ACCEPT_REQUEST)
+
+    def cancel(self):
```
mvadari, line 78: Is the task supposed to be canceled at some point?

mvadari (commented):
We can also stop creating the testnet wallets now

tests/integration/it_utils.py
```diff
@@ -60,6 +65,20 @@
 LEDGER_ACCEPT_REQUEST = GenericRequest(method="ledger_accept")
 
 
+class AsyncTestTimer:
```
mvadari, line 68: Why can't the normal timer be used in both situations?

mvadari (commented):
Something else that would save some time (and not really have any functional change) is to do a relsub check every second instead of every 4, like we're doing right now.

tests/integration/it_utils.py
```diff
@@ -124,6 +143,22 @@ async def sign_and_reliable_submission_async(
     return response
 
 
+def send_timed_reliable_submission(
```
JST5000, line 146: Can you add docstrings explaining what this function does + what the args are for?

tests/integration/it_utils.py
```diff
@@ -124,6 +143,22 @@ async def sign_and_reliable_submission_async(
...
+    client = _choose_client(use_json_client)
+    SyncTestTimer(1, client.request, (LEDGER_ACCEPT_REQUEST,)).start()
+    return send_reliable_submission_sync(transaction, client)
+
+
+async def send_timed_reliable_submission_async(
+    transaction: Transaction, use_json_client: bool = True
+) -> Response:
+    client = _choose_client_async(use_json_client)
+    AsyncTestTimer(1, client)
```
JST5000, line 158: You never call start on this one, is that intentional?

connorjchen: Yes, the AsyncTestTimer starts on initialization

tests/integration/it_utils.py
connorjchen: From my understanding, the task automatically cancels whenever the task completes. Should I keep the cancel function if I don't intend on using it? Unsure if it'll be useful in the future.

mvadari: That's fair, you can leave it in then

tests/integration/it_utils.py
connorjchen: AsyncTestTimer works for Async callback functions - sync functions don't work because asyncio.ensure_future (event loop error stuff)

I originally thought SyncTestTimer would work for everything (and it kinda did?). It worked in that it passed all test cases. However, I encountered a weird bug where on AsyncWebsocketClient (only), the client.request call never finished so the poetry poe test didn't exit process (which is why report coverage never called after and the CI tests ran indefinitely). As a result, I made a new test timer implementation for async tests.

tests/integration/it_utils.py
```diff
@@ -63,17 +63,19 @@
 FUNDING_AMOUNT = "1200000000"
 
 LEDGER_ACCEPT_REQUEST = GenericRequest(method="ledger_accept")
+LEDGER_ACCEPT_TIME = 1
 
 
 class AsyncTestTimer:
-    def __init__(self, timeout, client):
+    def __init__(self, timeout, client, request):
```
mvadari, line 70:
```suggestion
    def __init__(self, client, timeout=LEDGER_ACCEPT_TIME,  request=LEDGER_ACCEPT_REQUEST):
```
Since that's going to be the usecase most of the time (if not all)

tests/integration/it_utils.py
```diff
@@ -148,20 +150,20 @@ def send_timed_reliable_submission(
...
             `last_ledger_sequence` param.
-        client: the network client used to submit the transaction to a rippled node.
+        use_json_client: boolean for choosing json or websocket client.
 
     Returns:
         The response from the server, as a Response object.
     """
     client = _choose_client(use_json_client)
-    SyncTestTimer(1, client.request, (LEDGER_ACCEPT_REQUEST,)).start()
+    SyncTestTimer(LEDGER_ACCEPT_TIME, client.request, (LEDGER_ACCEPT_REQUEST,)).start()
```
mvadari, line 166: I recommend writing a wrapper class around `threading.Timer`, so the interface is the same for `AsyncTestTimer` and `SyncTestTimer`

tests/integration/it_utils.py
```diff
@@ -124,6 +145,74 @@ async def sign_and_reliable_submission_async(
...
+async def accept_ledger_async(use_json_client: bool = True) -> None:
+    """
+    Allows integration tests for async clients to send a `ledger_accept` request
+    after a set period of time.
+
+    Arguments:
+        use_json_client: boolean for choosing json or websocket client.
+    """
+    client = _choose_client_async(use_json_client)
+    AsyncTestTimer(LEDGER_ACCEPT_TIME, client, LEDGER_ACCEPT_REQUEST)
```
mvadari, line 213: These methods don't seem to ever be used.

tests/integration/it_utils.py
```diff
@@ -124,6 +145,74 @@ async def sign_and_reliable_submission_async(
...
+
+    Returns:
+        The response from the server, as a Response object.
+    """
+    client = _choose_client_async(use_json_client)
+    AsyncTestTimer(LEDGER_ACCEPT_TIME, client, LEDGER_ACCEPT_REQUEST)
+    return await send_reliable_submission_async(transaction, client)
+
+
+def accept_ledger(use_json_client: bool = True) -> None:
```
mvadari, line 192: I think this would be better if instead of `send_timed_reliable_submission_async`, you used the `accept_ledger` methods in the tests.

mvadari, line 192: This method should take another optional arg for the time until ledger_accept (with the default equal to `LEDGER_ACCEPT_TIME`

tests/integration/reqs/test_subscribe.py
```diff
@@ -1,17 +1,24 @@
...
 
 
 class TestSubscribe(IntegrationTestCase):
-    @test_async_and_sync(globals(), websockets_only=True, use_testnet=True)
+    @test_async_and_sync(globals(), websockets_only=True)
     async def test_ledger_subscription(self, client):
         await client.send(Subscribe(streams=[StreamParameter.LEDGER]))
         count = 0
         async for message in client:
+            await accept_ledger_async()
```
mvadari, line 21: You'll probably need an `accept_ledger_async` before this loop, because otherwise it won't get any initial message to enter the for loop

mvadari: ditto for every case here

connorjchen: The client seems to always get some sort of initial message, unsure if it is from the client.send(Subscribe) thing.

tests/integration/it_utils.py
```diff
@@ -58,6 +58,40 @@
 FUNDING_AMOUNT = "1200000000"
 
 LEDGER_ACCEPT_REQUEST = GenericRequest(method="ledger_accept")
+LEDGER_ACCEPT_TIME = 0.1
+
+
+class AsyncTestTimer:
+    def __init__(
+        self,
+        client: SyncClient,
```
mvadari, line 67:
```suggestion
        client: AsyncClient,
```

mvadari (approved):

mvadari:
Note: you should probably edit the name of this PR so it doesn't specifically mention relsub tests (since this PR now covers more than just that)

tests/integration/it_utils.py
```diff
@@ -124,6 +158,36 @@ async def sign_and_reliable_submission_async(
...
+def accept_ledger(
+    use_json_client: bool = True, timeout: float = LEDGER_ACCEPT_TIME
+) -> None:
+    """
+    Allows integration tests for sync clients to send a `ledger_accept` request
+    after a set period of time.
+
+    Arguments:
+        use_json_client: boolean for choosing json or websocket client.
+        timeout: float for how long to wait before accepting ledger.
```
JST5000, line 170: You should add the units here :) 
```suggestion
        timeout: float for how many seconds to wait before accepting ledger.
```

JST5000: (Same down below)

tests/integration/it_utils.py
```diff
@@ -124,6 +158,36 @@ async def sign_and_reliable_submission_async(
...
+    """
+    client = _choose_client(use_json_client)
+    SyncTestTimer(client, timeout)
+
+
+async def accept_ledger_async(
+    use_json_client: bool = True, timeout: float = LEDGER_ACCEPT_TIME
```
JST5000, line 177: I would rename `timeout` to `delay` since it's being used to stall the ledger accept command rather than giving a failure if the ledger accept command takes too long (which is how we use the word `timeout` elsewhere in the tests)

tests/integration/reqs/test_subscribe.py
```diff
@@ -1,17 +1,24 @@
...
 _MESSAGE_LIMIT = 3
 
 
 class TestSubscribe(IntegrationTestCase):
-    @test_async_and_sync(globals(), websockets_only=True, use_testnet=True)
+    @test_async_and_sync(globals(), websockets_only=True)
     async def test_ledger_subscription(self, client):
         await client.send(Subscribe(streams=[StreamParameter.LEDGER]))
         count = 0
         async for message in client:
```
JST5000, line 20: Might be worth a comment explaining that a default message comes from the client - otherwise someone reading the test later on might get confused and think that there's an issue with this line of these tests

JST5000 (approved):
LGTM

PR #458, opened 2022-10-27
fix: adjust `check_fee` logic to be more reasonable

mvadari:

This PR makes the behavior of `check_fee` much more reasonable. It will object when a user types in a fee that is over 10 drops and is signing offline, which is really annoying.

### Context of Change

Previously, if you had a fee that was greater than 10 drops for any normal transaction, the method would object (unless you turned off `check_fee`, and it's hard to remember that you can do that).  There have been many previous complaints about this, but I can't find them right now (they don't appear to be in the repo's issues). Users should be able to enter in a higher fee if they want.

Relevant issue (and the reason why this feature exists): https://github.com/XRPLF/xrpl-py/issues/213

### Type of Change

### DIFF
--- a/xrpl/asyncio/transaction/main.py
+++ b/xrpl/asyncio/transaction/main.py
@@ -269,7 +269,10 @@ async def _check_fee(transaction: Transaction, client: Optional[Client] = None) 
     Raises:
         XRPLException: if the transaction fee is higher than the expected fee.
     """
-    expected_fee = xrp_to_drops(0.1)  # a fee that is obviously too high
+    expected_fee = max(
+        xrp_to_drops(0.1),  # a fee that is obviously too high
+        await _calculate_fee_per_transaction_type(transaction, client),
+    )
 
     if transaction.fee and int(transaction.fee) > int(expected_fee):
         raise XRPLException(
--- a/xrpl/asyncio/transaction/main.py
+++ b/xrpl/asyncio/transaction/main.py
@@ -33,7 +33,7 @@ async def safe_sign_and_submit_transaction(
     wallet: Wallet,
     client: Client,
     autofill: bool = True,
-    check_fee: bool = False,
+    check_fee: bool = True,
 ) -> Response:
     """
     Signs a transaction (locally, without trusting external rippled nodes) and submits
@@ -62,7 +62,7 @@ async def safe_sign_and_submit_transaction(
 async def safe_sign_transaction(
     transaction: Transaction,
     wallet: Wallet,
-    check_fee: bool = False,
+    check_fee: bool = True,
 ) -> Transaction:
     """
     Signs a transaction locally, without trusting external rippled nodes.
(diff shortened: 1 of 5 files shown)


xrpl/asyncio/transaction/main.py
```diff
@@ -269,18 +269,13 @@ async def _check_fee(transaction: Transaction, client: Optional[Client] = None)
     Raises:
         XRPLException: if the transaction fee is higher than the expected fee.
     """
-    # Calculate the expected fee from the network load and transaction type
-    expected_fee = await _calculate_fee_per_transaction_type(transaction, client)
```
JST5000, line 273: Should `_calculate_fee_per_transaction_type` be removed?

xrpl/asyncio/transaction/main.py
```diff
@@ -269,18 +269,13 @@ async def _check_fee(transaction: Transaction, client: Optional[Client] = None)
...
+    expected_fee = xrp_to_drops(0.1)  # a fee that is obviously too high
```
JST5000, line 272: Also doesn't AccountDelete have a much higher fee than this by default?

xrpl/asyncio/transaction/main.py
mvadari: No, it's used for the autofill fee calculation

xrpl/asyncio/transaction/main.py
mvadari: Ok, I'll add it back in as a max

JST5000 (approved):
LGTM

xrpl/asyncio/transaction/main.py
```diff
@@ -269,18 +269,16 @@ async def _check_fee(transaction: Transaction, client: Optional[Client] = None)
...
-            "Fee value: "
-            + str(drops_to_xrp(transaction.fee))
-            + " XRP exceeds the "
-            + str(drops_to_xrp(expected_fee))
-            + " maximum XRP fee limit for "
-            + transaction.transaction_type
-            + " transaction"
+            f"Fee value: {str(drops_to_xrp(transaction.fee))} XRP is likely entered "
+            "incorrectly, since it is much larger than the "
+            f"{str(drops_to_xrp(expected_fee))} maximum XRP fee limit."
```
mDuo13, line 281: Saying it's "much larger than the maximum XRP fee limit" feels wrong here (especially since "the" is referring to a configurable setting in the client). 

```suggestion
            f"{str(drops_to_xrp(expected_fee))} typical XRP transaction cost."
```

The error message _could_ maybe also say something along the lines of, "If you meant to burn this much XRP, change (some setting) to bypass this check."

mDuo13 (approved):
Seems like a reasonable change. I suggested an improvement to the error message, but otherwise :+1: from me.

mDuo13 (approved):

PR #459, opened 2022-10-31
feat: Add multisign snippet, multisign functionality to autofill, and multisign helper

connorjchen:

Add code snippet for multisigning. Make autofill more compatible with multisigning process + make it more similar to xrpl.js

### Context of Change

Add quality of life code snippets for multisigning.
Helper function to autofill for multisigning compatability.

### Type of Change

- [x] Documentation Updates


snippets/multisigning.py
```diff
@@ -0,0 +1,102 @@
...
+    send_reliable_submission,
+)
+from xrpl.utils import str_to_hex
+from xrpl.wallet import generate_faucet_wallet
+
+client = JsonRpcClient("https://s.altnet.rippletest.net:51234/")
+
+wallet1 = generate_faucet_wallet(client, debug=True)
+wallet2 = generate_faucet_wallet(client, debug=True)
+wallet_master = generate_faucet_wallet(client, debug=True)
```
JST5000, line 19: nit: I'd rename `wallet1`/`wallet2` to indicate they're the signer wallets so later on the reader doesn't accidentally think `wallet1` is a random wallet, or the main wallet. Also, I'd recommend instantiating `wallet_master` first since it's the main wallet we need to think about. 
```suggestion
wallet_master = generate_faucet_wallet(client, debug=True)
signer_wallet_1 = generate_faucet_wallet(client, debug=True)
signer_wallet_2 = generate_faucet_wallet(client, debug=True)
```

snippets/multisigning.py
```diff
@@ -0,0 +1,102 @@
...
+    signer_quorum=2,
+    signer_entries=signer_entries,
+)
+signed_signer_list_set_tx = safe_sign_and_autofill_transaction(
+    signer_list_set_tx, wallet_master, client
+)
+signed_list_set_tx_response = send_reliable_submission(
+    signed_signer_list_set_tx, client
+)
+print("SignerListSet constructed successfully")
```
JST5000, line 37: We did more than just construct the transaction, and we don't know if the response is successful yet.

```suggestion
print("SignerListSet constructed and submitted, here's the response:")
```

JST5000: It could be worth it to have an explicit error check here (Since after this point if the SignerListSet transaction didn't work, everything else will be weird

e.g. 
```
if(signed_list_set_tx_response.result.engineResult != "tesSuccess"):
    print("Failed to submit the SignerListSet transaction!")
```

JST5000: (Note I don't know if I have the right extension to see the `engineResult` :P - I guessed what the path to that would be)

snippets/multisigning.py
```diff
@@ -0,0 +1,102 @@
...
+print(signed_list_set_tx_response)
+
+account_set_tx = AccountSet(
+    account=wallet_master.classic_address, domain=str_to_hex("example.com")
+)
+autofilled_account_set_tx = autofill(account_set_tx, client, 2)
```
JST5000, line 43: I would add a comment here highlighting the point of this `2` - Maybe even putting it into a variable that explains it's purpose (e.g. `expected_signer_count = 2`)

snippets/multisigning.py
```diff
@@ -0,0 +1,102 @@
...
+signed_signer_list_set_tx = safe_sign_and_autofill_transaction(
+    signer_list_set_tx, wallet_master, client
+)
+signed_list_set_tx_response = send_reliable_submission(
+    signed_signer_list_set_tx, client
+)
+print("SignerListSet constructed successfully")
+print(signed_list_set_tx_response)
+
+account_set_tx = AccountSet(
```
JST5000, line 40: I'd recommend adding a comment here explaining what we're about to do + why 

e.g.
```
# Now that we've set up multisigning, let's try using it to submit an AccountSet transaction
```

snippets/multisigning.py
```diff
@@ -0,0 +1,102 @@
...
+autofilled_account_set_tx_json = autofilled_account_set_tx.to_xrpl()
+tx_blob1 = sign(
+    bytes.fromhex(
+        encode_for_multisigning(
+            autofilled_account_set_tx_json,
+            wallet1.classic_address,
+        )
+    ),
+    wallet1.private_key,
+)
```
JST5000, line 56: This definitely looks like it should be in a helper of some sort so people don't need to figure out that they need to do this themselves :) (Maybe as part of the Wallet refactor?)

snippets/multisigning.py
```diff
@@ -0,0 +1,102 @@
...
+        encode_for_multisigning(
+            autofilled_account_set_tx_json,
+            wallet2.classic_address,
+        )
+    ),
+    wallet2.private_key,
+)
+
+# Individually signing the transaction with each signer.
+# This would normally be handled by your application.
```
JST5000, line 68: Making the reasoning more explicit since there are a lot of types of apps :)
```suggestion
# This would normally be handled by reaching out to the signing account owners in your application
# since you would not normally be in control of all the keys.
```

snippets/multisigning.py
```diff
@@ -0,0 +1,102 @@
...
+
+account_set_tx = AccountSet(
+    account=wallet_master.classic_address, domain=str_to_hex("example.com")
+)
+autofilled_account_set_tx = autofill(account_set_tx, client, 2)
+print("AccountSet transaction is ready to be multisigned")
+print(autofilled_account_set_tx)
+
+autofilled_account_set_tx_json = autofilled_account_set_tx.to_xrpl()
+tx_blob1 = sign(
```
JST5000, line 48: Might also be worth a comment explaining what a blob is (A binary string representation of the transaction in hex notation - the reason we encode it into a blob is to make it smaller for rippled to process :) )

snippets/multisigning.py
```diff
@@ -0,0 +1,102 @@
...
+multisigned_tx_response = client.request(SubmitMultisigned(tx_json=multisigned_tx))
+
+if multisigned_tx_response.result["engine_result"] == "tesSUCCESS":
+    print("The multisigned transaction was accepted by the ledger:")
+    print(multisigned_tx_response)
+    if multisigned_tx_response.result["tx_json"]["Signers"]:
+        print(
+            "The transaction had {} signatures".format(
+                len(multisigned_tx_response.result["tx_json"]["Signers"])
+            )
```
JST5000, line 95: nit: You could also use f-strings here instead. (No preference, just an option!)

It would look like: 
```
f"The transaction had {len(multisigned_tx_response.result["tx_json"]["Signers"])} signatures"
```

snippets/multisigning.py
connorjchen: I think it errors out and stops program automatically because it is send_reliable_submission

xrpl/asyncio/transaction/main.py
```diff
@@ -295,6 +301,8 @@ async def _calculate_fee_per_transaction_type(
     Args:
         transaction: the Transaction to be submitted.
         client: the network client with which to submit the transaction.
+        signers_count: the expected number of signers for this transaction.
+            Only used for multisigned transactions.
```
mvadari, line 305: The `signers_count` needs to be added to all the methods that call `autofill` as well, like `safe_sign_and_autofill`

connorjchen: xrpl.js doesn't seem to include `signers_count` on all methods that call upon `autofill`. Also, I'm not sure if would ever call `safe_sign_and_autofill` if we are doing multisigning

mvadari: Good point

snippets/multisigning.py
```diff
@@ -0,0 +1,69 @@
...
+]
+signer_list_set_tx = SignerListSet(
+    account=master_wallet.classic_address,
+    signer_quorum=2,
+    signer_entries=signer_entries,
+)
+signed_signer_list_set_tx = safe_sign_and_autofill_transaction(
+    signer_list_set_tx, master_wallet, client
+)
+signed_list_set_tx_response = send_reliable_submission(
```
JST5000, line 32: I'd recommend adding a print statement each time right before you submit in order to let the user know that the code is still working. (Otherwise it just looks like it's stuck)

Ex. (If you change this, you'll also want to change the print statement below)
```suggestion
print('Constructed SignerListSet and submitting it to the ledger...')
signed_list_set_tx_response = send_reliable_submission(
```

xrpl/asyncio/transaction/main.py
```diff
@@ -142,6 +144,51 @@ async def submit_transaction(
...
+    tx_dict["signers"] = [
+        Signer(
+            account=signers[idx].classic_address,
+            txn_signature=tx_blobs[idx],
+            signing_pub_key=signers[idx].public_key,
+        )
+        for idx in range(len(signers))
+    ]
+
+    return AccountSet.from_dict(tx_dict)
```
JST5000, line 189: I think this helper is doing a bit too much - like the comment says, you normally don't have access to the private keys in order to sign the transactions for yourself. Instead I think this helper should take individually signed transactions and combine them into a single transaction with all signatures. (Otherwise it doesn't generalize well)

The rough idea is basically that the individually signed transactions each have a field where each account's signature will end up after they sign, and those can be combined into the eventual "multisigned transaction"'s `Signers` field. Happy to chat in more detail about this.

xrpl/asyncio/transaction/main.py
```diff
@@ -142,6 +147,76 @@ async def submit_transaction(
...
+        tx_blobs: a list of signed transactions (in blob form) to combine into
+            a single signed transaction.
+
+    Returns:
+        The multisigned transaction.
+    """
+    # Individually signing the transaction with each signer.
+    # This would normally be handled by reaching out to the signing
+    # account owners in your application since you would not normally
+    # be in control of all the keys.
```
JST5000, line 202: We don't need this part of the comment here since this is a utility method instead of a snippet doing something a bit unorthodox.

xrpl/asyncio/transaction/main.py
```diff
@@ -325,9 +408,8 @@ async def _calculate_fee_per_transaction_type(
 
     # Multi-signed Transaction
     # 10 drops × (1 + Number of Signatures Provided)
-    if transaction.signers and len(transaction.signers) > 0:
-        base_fee = net_fee * (1 + len(transaction.signers)) + base_fee
-
+    if signers_count is not None and signers_count > 0:
+        base_fee += net_fee * (1 + signers_count)
```
JST5000, line 412: Why do we need the `signers_count` parameter? (It seems like the same value is calculated automatically without passing that in)

(And if it's not needed in this function, can it also be removed from the function that calls this?)

xrpl/asyncio/transaction/main.py
```diff
@@ -142,6 +147,76 @@ async def submit_transaction(
     raise XRPLRequestFailureException(response.result)
 
 
+async def safe_sign_transaction_with_multisign(
+    transaction: Transaction,
+    wallet: Wallet,
+) -> str:
+    """
+    Signs a transaction with multisign locally, without trusting external rippled nodes.
```
JST5000, line 155: nit: Rewording
```suggestion
    Signs a transaction to be used for multisigning later.
```

xrpl/transaction/main.py
```diff
@@ -139,5 +144,49 @@ def autofill(transaction: Transaction, client: SyncClient) -> Transaction:
...
+        )
+    )
+
+
+def safe_sign_transaction_with_multisign(
+    transaction: Transaction,
+    wallet: Wallet,
+) -> str:
+    """
+    Signs a transaction with multisign locally, without trusting external rippled nodes.
```
JST5000, line 157: nit: Same change as above
```suggestion
    Signs a transaction to be used for multisigning.
```

mvadari (commented):
IMO this snippet should wait until after the multisign helper method is written.

xrpl/asyncio/transaction/main.py
connorjchen: signers_count is the number of signers that a multisigned transaction expects to have. When this function is called in the multisign snippet, it hasn't yet been signed, so transaction.signers isn't correct. This autofills the transaction before the transaction is individually signed and multisigned. Also, this parameter isn't used in functions that call autofill (this function) because signers_count will only be used for multisigning and the other functions that call this function will never be used for multisigning.

connorjchen:
@mvadari The multisign helper method is a part of this PR

xrpl/asyncio/transaction/main.py
```diff
@@ -142,6 +147,72 @@ async def submit_transaction(
     raise XRPLRequestFailureException(response.result)
 
 
+async def safe_sign_transaction_with_multisign(
```
mvadari, line 150: This file is already massive - let's move these methods to a separate file in this folder (perhaps called `multisign`).

xrpl/asyncio/transaction/main.py
```diff
@@ -142,6 +147,72 @@ async def submit_transaction(
...
+    Signs a transaction with to be used for multisigning.
+
+    Args:
+        transaction: the transaction to be signed.
+        wallet: the wallet with which to sign the transaction.
+
+    Returns:
+        The signed transaction blob.
+    """
+    signature = sign(
```
mvadari, line 164: a sidenote: we should add support for `sign` to accept hex or bytes, and convert between them

snippets/multisigning.py
```diff
@@ -0,0 +1,75 @@
+"""Example of how we can multisign a transaction"""
```
mvadari, line 1: super nit: I'd call this file `multisign` instead of `multisigning`

snippets/multisigning.py
```diff
@@ -0,0 +1,75 @@
...
+    autofill,
+    multisign,
+    safe_sign_and_autofill_transaction,
+    send_reliable_submission,
+    sign_for_multisign,
+)
+from xrpl.utils import str_to_hex
+from xrpl.wallet import generate_faucet_wallet
+
+client = JsonRpcClient("https://s.altnet.rippletest.net:51234/")
```
mvadari, line 15: nit:
```suggestion
client = JsonRpcClient("https://s.altnet.rippletest.net:51234")
```

snippets/multisigning.py
```diff
@@ -0,0 +1,75 @@
...
+
+master_wallet = generate_faucet_wallet(client, debug=True)
```
mvadari, line 17: Add a comment describing what the debug does

xrpl/asyncio/transaction/multisign.py
```diff
@@ -0,0 +1,73 @@
+"""Multisign transaction methods with XRPL transactions."""
+from typing import List
+
+from xrpl.core.binarycodec import decode, encode, encode_for_multisigning
+from xrpl.core.keypairs.main import sign
+from xrpl.models.transactions.transaction import Signer, Transaction
+from xrpl.wallet.main import Wallet
+
+
+async def sign_for_multisign(
```
mvadari, line 10: I actually preferred the way that we did it in xrpl.js - when doing a multisign, the normal `sign` method has an extra param.

JST5000, line 10: Why is `sign_for_multisign` async?

xrpl/asyncio/transaction/multisign.py
```diff
@@ -0,0 +1,73 @@
...
+            account=wallet.classic_address,
+            txn_signature=signature,
+            signing_pub_key=wallet.public_key,
+        )
+    ]
+
+    return encode(Transaction.from_dict(tx_dict).to_xrpl())
+
+
+async def multisign(transaction: Transaction, tx_blobs: List[str]) -> Transaction:
```
JST5000, line 46: Same question here - does this function need to be async?

connorjchen: If the function doesn't require async methods, then should it be removed from asyncio folder entirely? So if someone was using Async client, they would just use sync version of it?

JST5000: That seems reasonable

snippets/multisign.py
```diff
@@ -0,0 +1,72 @@
...
+# transaction.
+account_set_tx = AccountSet(
+    account=master_wallet.classic_address, domain=str_to_hex("example.com")
+)
+autofilled_account_set_tx = autofill(account_set_tx, client, len(signer_entries))
+print("AccountSet transaction is ready to be multisigned")
+print(autofilled_account_set_tx)
+
+tx_1 = sign(autofilled_account_set_tx, signer_wallet_1, multisign=True)
+tx_2 = sign(autofilled_account_set_tx, signer_wallet_2, multisign=True)
```
JST5000, line 50: nit: I'd add a comment above here explaining that this is likely something they'd want to change. 

Ex. 
```
# Since we created both signer keys, we can sign locally, but if you were to build an app
# That allows multisigning, you would need to request signatures from the key holders.
```

tests/integration/reqs/test_submit_multisigned.py
```diff
@@ -1,115 +1,59 @@
...
-                    account=SECOND_SIGNER.classic_address,
-                    txn_signature=second_sig,
-                    signing_pub_key=SECOND_SIGNER.public_key,
-                ),
-            ],
         )
+        autofilled_tx = await autofill(tx, client, len(SIGNER_ENTRIES))
+        tx_1 = await sign(autofilled_tx, FIRST_SIGNER, multisign=True)
+        tx_2 = await sign(autofilled_tx, SECOND_SIGNER, multisign=True)
+        multisigned_tx = multisign(autofilled_tx, [tx_1, tx_2])
```
JST5000, line 56: Two recommendations:
1. I'd recommend using a fixed FIRST_SIGNER / SECOND_SIGNER, and then doing a check that `multisign(autofilled_tx, [tx_1, tx_2])` matches a pre-computed value of what that should be. (You can use `xrpl.js`'s multisign for those fixed signer values as a reference to see what the correct output for `multisign` should be) - The reason for this is it acts as a unit test going forward ensuring that `multisign` gives an accurate value for a given wallet - not just one that rippled receives successfully. 

2.  Since this is the defacto integration test for multisign, I think it should go through the proper steps of setting up signers and checking that rippled actually allows the multisigned transaction to take effect. (Basically doing the same steps you have in `multisign.py` but replacing the print statements with asserts :) - Otherwise I'm worried that the test could pass while multisigned transactions are being rejected by rippled.

snippets/multisign.py
```diff
@@ -0,0 +1,72 @@
...
+signer_entries = [
+    SignerEntry(account=signer_wallet_1.classic_address, signer_weight=1),
+    SignerEntry(account=signer_wallet_2.classic_address, signer_weight=1),
+]
+signer_list_set_tx = SignerListSet(
+    account=master_wallet.classic_address,
+    signer_quorum=2,
+    signer_entries=signer_entries,
+)
+signed_signer_list_set_tx = autofill_and_sign(signer_list_set_tx, master_wallet, client)
```
mvadari, line 32: nit: add a newline after this line

tests/integration/reqs/test_submit_multisigned.py
```diff
@@ -1,120 +1,127 @@
...
 from xrpl.wallet import Wallet
 
-FEE = get_fee(JSON_RPC_CLIENT)
-
-#
 # Set up signer list
-FIRST_SIGNER = Wallet.create()
-SECOND_SIGNER = Wallet.create()
+# Warning: Order of these seeds matter
+# Since SignerList will be sorted by codec
```
JST5000, line 14: What was the error you got when these were out of order? (If this was hard to debug, we should probably fix that because it's a very easy thing to do) - Maybe open a ticket with details?

JST5000 (approved):
LGTM

tests/integration/reqs/test_submit_multisigned.py
mvadari: Yeah, the order really shouldn't matter...

xrpl/asyncio/transaction/main.py
```diff
@@ -75,18 +76,39 @@ async def sign(
...
     """
-    if check_fee:
-        await _check_fee(transaction)
-    transaction_json = _prepare_transaction(transaction, wallet)
-    serialized_for_signing = encode_for_signing(transaction_json)
-    serialized_bytes = bytes.fromhex(serialized_for_signing)
-    signature = keypairs_sign(serialized_bytes, wallet.private_key)
-    transaction_json["TxnSignature"] = signature
-    return Transaction.from_xrpl(transaction_json)
+    if not multisign:
```
mvadari, line 84: This if statement should be flipped - the logic becomes minorly complicated to understand when the `else` is called (`else if not not multisign`), so it's better to just have an `if multisign` block and flip the if and else blocks.

xrpl/asyncio/transaction/main.py
```diff
@@ -335,9 +365,8 @@ async def _calculate_fee_per_transaction_type(
 
     # Multi-signed Transaction
     # 10 drops × (1 + Number of Signatures Provided)
-    if transaction.signers and len(transaction.signers) > 0:
-        base_fee = net_fee * (1 + len(transaction.signers)) + base_fee
-
+    if signers_count is not None and signers_count > 0:
+        base_fee += net_fee * (1 + signers_count)
```
mvadari, line 369: This feels like a breaking change - wouldn't we want to check both?

xrpl/transaction/multisign.py
```diff
@@ -0,0 +1,32 @@
...
+    tx_dict["signers"] = [
+        Signer(
+            account=decoded_tx_signer["Account"],
+            txn_signature=decoded_tx_signer["TxnSignature"],
+            signing_pub_key=decoded_tx_signer["SigningPubKey"],
+        )
+        for decoded_tx_signer in decoded_tx_signers
+    ]
+
+    return Transaction.from_dict(tx_dict)
```
connorjchen, line 32: note: fix naming convention for safe_sign_and_autofill usage in sign_and_submit

tests/integration/reqs/test_submit_multisigned.py
connorjchen: The SignerList got sorted by the codec when signing or something. This made it so `multisigned_tx.to_xrpl()` wasn't exactly what the ledger used for response data fields. Since it wasn't the same, `"tx_blob": encode(multisigned_tx.to_xrpl())` and `"hash": multisigned_tx.get_hash()` didn't match.

xrpl/asyncio/transaction/main.py
connorjchen: I see how this would be a breaking change. However, I am unsure how someone is supposed to use `autofill` for `multisign` with `transaction.signers`. If I knew how, I wouldn't have added in `signers_count`.

For `multisign`, you must make a transaction, then have people sign this, and then submit it. To my knowledge, you must specify `Fee` in the original transaction because people must sign this specific transaction. If you do it this way, `transaction.signers = 0` at the time of creating the transaction, which is bad. So, you must explicitly pass in `signers_count` (which is how xrpl.js uses it).

I am unsure if I am missing something. The original `test_submit_multisigned.py` file doesn't even use autofill, they fill it with `FEE = get_fee(JSON_RPC_CLIENT)`. I'm not sure if logic with `transactions.signers` is tested/used anywhere.

JST5000: I think this is just a bug fix - I don't think it was calculating correctly for multisigning previously.

tests/integration/reqs/test_submit_multisigned.py
mvadari: I'm a little confused, the tx_blob and the hash shouldn't match because they have different data - the tx_blob is the encoded transaction and the hash is the hash of the transaction.

xrpl/asyncio/transaction/main.py
```diff
@@ -75,18 +74,39 @@ async def sign(
...
+        tx_dict = transaction.to_dict()
+        tx_dict["signers"] = [
+            Signer(
+                account=wallet.classic_address,
+                txn_signature=signature,
+                signing_pub_key=wallet.public_key,
+            )
+        ]
+        return Transaction.from_dict(tx_dict)
+    else:
```
mvadari, line 101: This `else` isn't necessary, because the `if` returns something.

tests/integration/reqs/test_submit_multisigned.py
connorjchen: I meant my expected tx_blob didn't match the actual tx_blob and the expected hash didn't match the actual hash. 

This was because when it submitted, it sorted the signer list and then filled in tx_blob and hash. But on my end, the transaction I submitted, was never sorted. So my calculated of expected values was not correct.

mvadari: You don't need to check that level of detail in the response - just make sure that it was successful and that it actually did change the domain. rippled does the checking for us.

connorjchen: Oh, I thought it'd be better to have little more robust checking. I adapted bit of logic from xrpl.js submitMultisigned test. Do you suggest I change it to check less?
```
const expectedResponse: SubmitMultisignedResponse = {
      id: submitResponse.id,
      type: 'response',
      result: {
        engine_result: 'tesSUCCESS',
        engine_result_code: 0,
        engine_result_message:
          'The transaction was applied. Only final in a validated ledger.',
        tx_blob: multisigned,
        tx_json: {
          ...(decode(multisigned) as unknown as Transaction),
          hash: hashSignedTx(multisigned),
        },
      },
    }

    assert.deepEqual(submitResponse, expectedResponse)
   ```

Comment (mvadari) [reply tests/integration/reqs/test_submit_multisigned.py:14]: Yes. xrpl.js does more checking because it has response objects - xrpl-py does not, so it's unnecessary.

Comment (connorjchen) [reply tests/integration/reqs/test_submit_multisigned.py]: Ok, removed lot of asserts and just checked domain and successful at end

File: tests/integration/reqs/test_submit_multisigned.py [hunk 31]
```diff
@@ -1,120 +1,60 @@
...
+        multisigned_tx = multisign(autofilled_tx, [tx_1, tx_2])
+
         # submit tx
         response = await client.request(
             SubmitMultisigned(
                 tx_json=multisigned_tx,
             )
         )
         self.assertTrue(response.is_successful())
+        self.assertEqual(response.result["tx_json"]["Domain"], "6578616D706C652E636F6D")
```
mvadari, line 60: this should be a constant, to make it more future-proof

intelliot (approved):

pdp2121 (approved):

PR #465, opened 2022-11-03
feat: new ways of Wallet generation

connorjchen:

Added new ways to create a wallet: from_seed, from_public_private_key, from_entropy

### Context of Change

In xrpl.js we have quite a few different ways to instantiate a Wallet 

From Public/Private keys
From secret
From entropy

It’d be great to port these over to make the Wallet an easier class to work with.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Wrote new tests


xrpl/wallet/main.py
```diff
@@ -16,23 +17,53 @@ class Wallet:
...
+        """
+        Alias for wallet.classic_address.
+
+        Returns:
+            The address that publicly identifies this wallet, as a base58 string.
+        """
+        return self.classic_address
+
+    @address.setter
+    def address(self: Wallet, value: str) -> None:
```
mvadari, line 31: Why should this be set-able?

xrpl/wallet/main.py
```diff
@@ -16,23 +17,53 @@ class Wallet:
...
         algorithm: Optional[CryptoAlgorithm] = None,
+        master_address: Optional[str] = None,
     ) -> None:
         """
         Generate a new Wallet.
 
         Args:
             seed: The seed from which the public and private keys are derived.
-            sequence: The next sequence number for the account.
+                Generates a new seed if not included.
```
mvadari, line 53: If this is happening now, then `.create()` should be deprecated in favor of this.

xrpl/wallet/main.py
```diff
@@ -16,23 +17,53 @@ class Wallet:
...
+            sequence: The next sequence number for the account. Defaulted to 0 if not
+                included.
             algorithm: The algorithm used to encode the keys. Inferred from the seed if
-                not included.
+                not included. The default is None.
+            master_address: Include if a Wallet uses a Regular Key Pair. It must be
```
mvadari, line 58: super nit: this should probably go before `algorithm`

xrpl/wallet/main.py
```diff
@@ -16,23 +17,53 @@ class Wallet:
...
+                the master address of the account. The default is `None`.
         """
-        self.seed = seed
+        if seed is None:
+            if algorithm is None:
+                seed = generate_seed()
+            else:
+                seed = generate_seed(algorithm=algorithm)
+        self.seed: Optional[str] = seed
```
mvadari, line 66:
```suggestion
        self.seed: str = seed
```

mvadari: Actually, the type should be inferrable

xrpl/utils/ensure_classic_address.py
```diff
@@ -0,0 +1,35 @@
+"""Classic address helper utils."""
+
+from xrpl.constants import XRPLException
+from xrpl.core.addresscodec import is_valid_xaddress, xaddress_to_classic_address
+
+
+def ensure_classic_address(account: str) -> str:
```
mvadari, line 7: If this is being added as a helper method, it should really be in the address codec

xrpl/wallet/main.py
connorjchen: If someone wants to do `wallet.address = "something"`, we would need this setter

xrpl/wallet/main.py
```diff
@@ -77,8 +112,88 @@ def create(
...
+        wallet = cls()
+        wallet.seed = None
+        wallet.public_key = public_key
+        wallet.private_key = private_key
+        wallet.classic_address = (
+            ensure_classic_address(master_address)
+            if master_address is not None
+            else derive_classic_address(public_key)
+        )
+        return wallet
```
mvadari, line 145: Instead of doing this (which is kind of hacky), what do you think about essentially an alternate constructor?
```py
def __init__(
        self: Wallet,
        seed: Optional[str] = None,
        sequence: Optional[int] = 0,
        *,
        public_key: Optional[str] = None,
        private_key: Optional[str] = None,
        algorithm: Optional[CryptoAlgorithm] = None,
        master_address: Optional[str] = None,
    ) -> None
    # it's invalid for both `seed` and `public_key`/`private_key` to be passed in - throw an error if that's the case
```

xrpl/wallet/main.py
```diff
@@ -77,8 +112,88 @@ def create(
...
+
+    @classmethod
+    def from_seed(
```
mvadari, line 148: This feels unnecessary given the constructor

xrpl/wallet/main.py
```diff
@@ -16,23 +17,53 @@ class Wallet:
...
+            value: New value for address/classic_address.
+        """
+        self.classic_address = value
+
     def __init__(
         self: Wallet,
-        seed: str,
-        sequence: int,
+        seed: Optional[str] = None,
+        sequence: Optional[int] = 0,
```
mvadari, line 43:
```suggestion
        sequence: int = 0,
```
`None` should not be a valid value for this

xrpl/wallet/main.py
connorjchen: Makes sense

xrpl/wallet/main.py
mvadari: We should discourage that, especially now that there's regular key support.

xrpl/wallet/main.py
connorjchen: Put master address before algorithm? What's the reasoning?

mvadari: It's more likely to be used. Everything after the `*` must use a keyword, so order doesn't matter.

Like I said, super nit - if you don't want to, that's fine.

xrpl/wallet/main.py
connorjchen: Yeah I agree. Wanted to make it similar to xrpl.js but not sure if it matters too much. Should I remove from_seed and from_secret as a whole?

xrpl/wallet/main.py
connorjchen: All of this logic would go in the same constructor for the wallet? Not 2 separate constructors?

mvadari: Python doesn't really support alternate constructor in the normal/Java sense.

xrpl/utils/ensure_classic_address.py
connorjchen: It exists in https://github.com/XRPLF/xrpl.js/blob/develop/packages/xrpl/src/sugar/utils.ts for xrpl.js. 

Question:
Is the address codec code the same between xrpl-py and xrpl.js? Are there people working on both?

xrpl/wallet/main.py
connorjchen: I think this method might be a bit confusing for developers

xrpl/wallet/main.py
connorjchen: Hm, if I don't put the optional type, an err is thrown in my `from_public_private_key` method when I try to do `seed = None`. Unsure what best approach is.

xrpl/utils/ensure_classic_address.py
mvadari: We can move it for xrpl-py - I think that's a better place for it to live. All the other code that involves classic vs x-addresses is there.

mvadari: The two libraries don't need to be _exactly_ the same.

xrpl/wallet/main.py
JST5000: I think the reverse should be true - you shouldn't be able to set either `wallet.classic_address` or `wallet.address` after initialization.

connorjchen:
PR on hold - evaluating necessity for breaking changes to Wallet for xrpl-py 2.0

PR #468, opened 2022-11-15
refactor: Alias function for submit_transaction called submit

connorjchen:

Added alias function for `submit_transaction` called `submit`

### Context of Change

`submit_transaction` is too long and unnecessary -- potentially deprecate/remove in xrpl-py 2.0

### Type of Change

- [ ] Documentation Updates
- [ ] Release


connorjchen:
One thing I wasn’t sure of is if I should change the integration test it_util functions submit_transaction and submit_transaction_async to follow similar naming conventions.

Also, I tried to find a cleaner way to copy docstring over without copy paste, but doesn't seem like many good methods while also using the @deprecated decorator in the way I did.

xrpl/asyncio/transaction/main.py
```diff
@@ -142,6 +143,30 @@ async def submit_transaction(
     raise XRPLRequestFailureException(response.result)
 
 
+@deprecated(
+    reason="Use `submit` instead - it is a more concise, descriptive name.",
+    version="1.8.0",
+)
+async def submit_transaction(
```
mvadari, line 150: I think you can just do `submit_transaction = submit`

xrpl/asyncio/transaction/main.py
```diff
@@ -142,6 +143,30 @@ async def submit_transaction(
     raise XRPLRequestFailureException(response.result)
 
 
+@deprecated(
```
mvadari, line 146: This doesn't need to be deprecated until v2.0 (if then). It can stay as a supported alias.

JST5000:
> One thing I wasn’t sure of is if I should change the integration test it_util functions `submit_transaction` and `submit_transaction_async` to follow similar naming conventions.

I think it makes sense to rename those to similar naming conventions, although I don't think it's a big deal. (When testing being verbose is fine)

JST5000:
You should probably add a test case for `submit` which matches a test case for `submit_transaction` - they should work the same, but it's such a core feature that I'd prefer to have that be guaranteed by our CI

tests/integration/sugar/test_transaction.py
```diff
@@ -236,6 +237,26 @@ async def test_payment_high_fee_authorized(self, client):
         self.assertTrue(response.is_successful())
         WALLET.sequence += 1
 
+    @test_async_and_sync(globals())
+    async def test_payment_high_fee_authorized_submit_xalias(self, client):
+        # GIVEN a new Payment transaction
+        response = await submit_transaction_with_alias_async(
```
mvadari, line 243: Instead of creating a helper method for this one test, you should instead use `submit_transaction` and add it to the list of dependencies that need to be imported.

jonathanlei (approved):

tests/integration/sugar/test_transaction.py
```diff
@@ -236,6 +239,35 @@ async def test_payment_high_fee_authorized(self, client):
...
+            "xrpl.transaction.submit_transaction",
+        ],
+    )
+    async def test_payment_high_fee_authorized_with_submit_alias(self, client):
+        signed_and_autofilled = await safe_sign_and_autofill_transaction(
+            Payment(
+                account=WALLET.classic_address,
+                sequence=WALLET.sequence,
+                amount="1",
+                # WITH the fee higher than 2 XRP
```
JST5000, line 255: I think you can remove these comments in this function. I don't think they add much. 

(Or otherwise the GIVEN comment needs to be when the Payment value is defined)

JST5000 (approved):
Logic looks good :) (Just a small comment about comments)

intelliot (approved):

PR #469, opened 2022-11-15
fix packaging and `md` syntax issues

wiseaidev:

- Made the package [pep-561](https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages) compatible by Including the root `py.typed` file in the package. 
- Removed redundant `py.typed` files.
- Updated `poetry` and `poetry-core`.
- Fixed markdown syntax.
- Fixed a `flake8` issue.
- Removed an unnecessary `MANIFEST.in` file.

### Type of Change

- [X] Documentation Updates
- [ ] Release

## Test Plan

I haven't tested these changes yet. I will be working on it later, or you can trigger the CI.

## Context

Note that the new version of poetry introduced new arguments like `--group`. For example, you can run the following commands to:

- Install the main dependencies only:

```sh
poetry install --only main
``` 

- Install the `dev` dependencies:

```sh
poetry install --only dev
``` 

- Add a new dependency to `dev`:

```sh
poetry add pytest --group dev
```

- Add a new dependency to `main`:

```sh
poetry add requests
```

You can refer to [`python-poetry.org/docs/cli`](https://python-poetry.org/docs/cli/) for more info.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,3 +1,32 @@
 [mypy]
+mypy_path = xrpl/
 exclude = dist
+ignore_errors = False
 ignore_missing_imports = True
+warn_unused_configs = True
+warn_redundant_casts = True
+warn_unused_ignores = True
+no_implicit_optional = True
+strict_equality = True
+strict_concatenate = True
+
+# Strongly recommend enabling this
+check_untyped_defs = True
+
+disallow_subclassing_any = False
+disallow_untyped_decorators = True
+disallow_any_generics = True
+
+# forcing use of type annotations
+disallow_untyped_calls = True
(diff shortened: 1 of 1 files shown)


mvadari:
Will removing all the individual `py.typed` in individual folders support the syntax of `from xrpl.models import ...`?

poetry.lock
```diff
@@ -19,7 +19,7 @@ python-versions = "*"
 
 [[package]]
 name = "anyio"
-version = "3.6.1"
```
mvadari, line 22: Please avoid running `poetry update` in this PR so that we don't have issues with dependencies.

PR #473, opened 2022-11-16
refactor: Deprecate xrpl-py sugar that is just a wrapper on a request

connorjchen:

Deprecated ___

### Context of Change

Some functions in xrpl-py are identical to a rippled request, but with less features since the return types are simplified. We should deprecate these, and instead recommend that people just directly do client.request(...) with the appropriate request type. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/xrpl/account/transaction_history.py
+++ b/xrpl/account/transaction_history.py
@@ -27,7 +27,7 @@ def get_latest_transaction(account: str, client: SyncClient) -> Response:
 
 
 @deprecated(
-    reason="Sending an AccountTx request directly allows you to page through all"
+    reason="Sending an AccountTx request directly allows you to page through all "
     "results and is just as easy to use.",
     version="1.6.0",
 )
@@ -64,7 +64,7 @@ def get_account_payment_transactions(
         client: the network client used to make network calls.
 
     Returns:
-        The most recent payment transaction history for the address. For the full
+        The first page of payment transaction history for the address. For the full
         history, page through the :class:`AccountTx` request directly.
     """
     return asyncio.run(
--- a/xrpl/asyncio/account/transaction_history.py
+++ b/xrpl/asyncio/account/transaction_history.py
@@ -35,7 +35,7 @@ async def get_latest_transaction(account: str, client: Client) -> Response:
 
 
 @deprecated(
-    reason="Sending an AccountTx request directly allows you to page through all"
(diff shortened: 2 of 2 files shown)


xrpl/account/transaction_history.py
```diff
@@ -27,7 +27,7 @@ def get_latest_transaction(account: str, client: SyncClient) -> Response:
 
 
 @deprecated(
-    reason="Sending an AccountTx request directly allows you to page through all"
+    reason="Sending an AccountTx request directly allows you to page through all "
```
JST5000, line 30:
```suggestion
    reason="Sending an AccountTx request directly allows you to page through all."
```

xrpl/asyncio/account/main.py
```diff
@@ -97,11 +99,30 @@ async def get_account_root(
 
     Returns:
         The AccountRoot dictionary for the address.
+
+    Raises:
+        XRPLRequestFailureException: if the rippled API call fails.
     """
-    account_info = await get_account_info(address, client, ledger_index)
+    if is_valid_xaddress(address):
+        address, _, _ = xaddress_to_classic_address(address)
```
JST5000, line 107: It's bad practice to overwrite a parameter because it can be very surprising to have the parameter change (Especially if it's an object, in which case the change could cause side effects)

Instead I'd recommend creating a variable inside the function and setting/changing it instead of the parameter.

xrpl/account/transaction_history.py
JST5000: Nevermind, oops!

JST5000 (approved):
LGTM after one small change

khancode (approved):

xrpl/asyncio/account/main.py
```diff
@@ -97,11 +99,32 @@ async def get_account_root(
...
+
+    Raises:
+        XRPLRequestFailureException: if the rippled API call fails.
     """
-    account_info = await get_account_info(address, client, ledger_index)
+    classic_address = address
+
+    if is_valid_xaddress(address):
+        classic_address, _, _ = xaddress_to_classic_address(address)
+
```
mvadari, line 110: Isn't there a helper method that does this conversion for you?

connorjchen: I don't believe so, it seems that other logic in this file does it this way

PR #475, opened 2022-11-21
refactor: simplify `TypedDict` types

mvadari:

This PR uses `NotRequired` (equivalent to TypeScript `?` in interfaces) to simplify the metadata `TypedDict` types, making the code both simpler and easier to read.

### Context of Change

I wanted to try out a new feature that was added in Python 3.11 and backported to previous Python versions.

The `TypedDict` metadata types were messy because `NotRequired` wasn't supported previously. But it has since been added to `typing_extensions`.

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/xrpl/utils/txn_parser/utils/nodes.py
+++ b/xrpl/utils/txn_parser/utils/nodes.py
@@ -16,27 +16,17 @@ from xrpl.models.transactions.metadata import (
 )
 
 
-class OptionalFieldNames(TypedDict, total=False):
-    """The optional fields of `NormalizedNode`."""
-
-    """
-    The fields are separated from `NormalizedNode` to make them optional,
-    while keeping `NodeType`, `LedgerEntryType` and `LedgerIndex` required.
-    """
-
-    NewFields: Optional[Fields]
-    FinalFields: Optional[Fields]
-    PreviousFields: Optional[Fields]
-    PreviousTxnID: Optional[str]
-    PreviousTxnLgrSeq: Optional[int]
-
-
-class NormalizedNode(OptionalFieldNames):
+class NormalizedNode(TypedDict):
     """A model representing an affected node in a standard format."""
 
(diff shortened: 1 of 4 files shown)


ckniffen (approved):

mvadari:
cc @LimpidCrypto

xrpl/utils/txn_parser/utils/nodes.py
```diff
@@ -16,29 +16,19 @@
...
+    """A model representing an affected node in a standard format."""
 
+    NodeType: Literal["CreatedNode", "ModifiedNode", "DeletedNode"]
+    LedgerEntryType: str
+    LedgerIndex: str
     NewFields: Optional[Fields]
     FinalFields: Optional[Fields]
     PreviousFields: Optional[Fields]
     PreviousTxnID: Optional[str]
     PreviousTxnLgrSeq: Optional[int]
```
JST5000, line 29: Should these 5 also be `NotRequired` instead of `Optional`?

mvadari: No, those are actually `None`. If you'd rather they not be there instead of having the value of `None`, I can do that.

LimpidCrypto:
> cc @LimpidCrypto

Nice, thanks for tagging me. Would you prefer using `TypedDict` with `NotRequired` for the objects over the way I did in #444 ?

xrpl/utils/txn_parser/utils/nodes.py
JST5000: Ah that makes sense - I got a bit confused about the difference between the two :)

mvadari:
> > cc @LimpidCrypto
> 
> Nice, thanks for tagging me. Would you prefer using `TypedDict` with `NotRequired` for the objects over the way I did in #444 ? Would be nice to know, before I separate the PR :)

I think the objects are fine 🙂

JST5000 (approved):
LGTM

PR #477, opened 2022-11-22
breaking: Refactored wallet class and removed wallet sequence

connorjchen:

- Complete revamped wallet class to be more alike to xrpl.js (breaking changes)
- Some (breaking) changes to key formatting and seed generation with given entropy

### Context of Change

xrpl-py wallet class was outdated and lacked many functions that xrpl.js offers
https://docs.google.com/document/d/1G_5H40eD1gKaue6XgdecP5SYifrljkCjtGsZnVZDLyo/edit#

### Type of Change

- [x] Documentation Updates
- [ ] Release

## Test Plan

wrote entire test suite for wallet class


CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.0.0]
+### Added:
+- Wallet support for regular key compatibility
+
+### Changed:
+- Updated params for Wallet class constructor and added new ways of wallet generation: from_seed, from_secret, from_entropy
```
JST5000, line 13: I would move the second half of this line to `Added`

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
...
+- Added address alias to Wallet.classic_address (both now also readonly)
```
JST5000, line 14: 1st half of this one should be in Added - second half could be in changed. (The second half should be turned into a full sentence though)

xrpl/wallet/main.py
```diff
@@ -16,69 +17,131 @@ class Wallet:
...
-        submission.
+        The core value that is used to derive all other information about
+        this wallet. MUST be kept secret!
         """
 
     @classmethod
-    def create(
-        cls: Type[Wallet], crypto_algorithm: CryptoAlgorithm = CryptoAlgorithm.ED25519
+    def generate(
+        cls: Type[Wallet], algorithm: CryptoAlgorithm = CryptoAlgorithm.ED25519
```
JST5000, line 78: I think this doesn't need to be renamed - `create` seems like just as good a name as `generate` in my opinion.

tests/integration/reqs/test_submit_multisigned.py
```diff
@@ -20,16 +22,16 @@
 from xrpl.wallet import Wallet
 
 FEE = get_fee(JSON_RPC_CLIENT)
+SEQ = get_next_valid_seq_number_sync(WALLET.classic_address, JSON_RPC_CLIENT)
```
JST5000, line 25: This shouldn't be a global because the effectiveness of the sequence is impacted by the test case - so if multiple tests were added to this file, they wouldn't be able to re-use the SEQ value. (Similarly the values that depend on SEQ should be defined within the test case in my opinion) 

I think it's also a bit more readable to have all the moving parts in the test case, rather than assuming people read the top-level code before diving into the specific test they care about.

xrpl/wallet/main.py
```diff
@@ -16,69 +17,131 @@ class Wallet:
     details.
     """
 
+    @property
+    def classic_address(self: Wallet) -> str:
+        """
+        The address that publicly identifies this wallet,
+        as a base58 string.
+        """  # noqa: DAR201
```
JST5000, line 25: Is this noqa still necessary?

xrpl/wallet/main.py
```diff
@@ -16,69 +17,131 @@ class Wallet:
...
-                not included.
-        """
-        self.seed = seed
-        """
-        The core value that is used to derive all other information about
-        this wallet. MUST be kept secret!
+            public_key: The public key for the account.
+            private_key: The private key used for signing transactions for the account.
+            master_address: Include if a Wallet uses a Regular Key Pair. It must be
+                the master address of the account. The default is `None`.
```
JST5000, line 48: "It" is a bit ambiguous (Could be talking about the Wallet/Regular Key Pair/ or master_address)

```suggestion
            master_address: Include if a Wallet uses a Regular Key Pair. This sets the address
                    that this wallet corresponds to. The default is `None`.
```

xrpl/wallet/main.py
```diff
@@ -16,69 +17,131 @@ class Wallet:
...
+        cls: Type[Wallet], algorithm: CryptoAlgorithm = CryptoAlgorithm.ED25519
     ) -> Wallet:
         """
         Generates a new seed and Wallet.
 
         Args:
-            crypto_algorithm: The key-generation algorithm to use when generating the
-                seed. The default is Ed25519.
+            algorithm: The digital signature algorithm to generate an address for.
+                The default is Ed25519.
```
JST5000, line 85: I think the previous explanation is a bit more accurate (While the seed can be used to derive an address, the main point of this is to get a seed or public/private keys)

xrpl/wallet/main.py
```diff
@@ -16,69 +17,131 @@ class Wallet:
...
-        seed = generate_seed(algorithm=crypto_algorithm)
-        return cls(seed, sequence=0, algorithm=crypto_algorithm)
+        seed = generate_seed(algorithm=algorithm)
+        return Wallet.from_seed(seed, algorithm=algorithm)
+
+    @classmethod
+    def from_seed(
+        cls: Type[Wallet],
+        seed: str,
+        *,
```
JST5000, line 97: Why does this have a * parameter?

JST5000: (Similar question for the functions below)

xrpl/wallet/main.py
connorjchen: Ok, was weighing between keeping similar to xrpl.js vs keeping as similar to old xrpl-py

xrpl/wallet/main.py
connorjchen: Because it is like a function, it wanted the docstring to have a return parameter explanation, but I wanted the docstring to appear alike to other properties (like self.seed in the constructor).

xrpl/wallet/main.py
connorjchen: * in Python is used to tell program that params after * need to be named when called like from_seed(seed, algorithm=algorithm) instead of from_seed(seed, None, algorithm)

I added * because params like master_address and algorithm are like options passed into the method. Instead of passing in opts param, I used *. I think it's clearer than having no * since the order of parameters would matter then.

xrpl/wallet/main.py
JST5000: My thought process is that since the goal is to make xrpl-py better we want to make a good design that is easy to transition from wherever possible. (So since this is a neutral change in terms of usability, sticking with xrpl-py convention is preferred)

xrpl/wallet/main.py
JST5000: Might be worth adding to the docstring for the functions that use it because it seems like a rare enough feature that we wouldn't expect new coders to know it before encountering our library.

connorjchen: I think it might be okay because if they write `Wallet.from_seed(seed, algorithm)`, program will error with something like "TypeError: from_seed() takes 1 positional arguments but 2 were given" which is easily debuggable + googleable.

xrpl/wallet/main.py
```diff
@@ -16,69 +17,141 @@ class Wallet:
...
+            seed: The seed (secret) used to derive the account keys.
+            master_address: Include if a Wallet uses a Regular Key Pair. This sets the
+                address that this wallet corresponds to. The default is `None`.
+            algorithm: The key-generation algorithm to use when generating the seed.
+                The default is `ED25519`.
+
+        Returns:
+            The wallet that is generated from the given secret.
+        """
+        public_key, private_key = derive_keypair(seed, algorithm=algorithm)
```
JST5000, line 114: I don't know if we should be setting a default algorithm. Decoding based on the seed prefix seems to make sense to me as better behavior that users can override if they have a problem with it. (Rather than requiring every secp256k1 user to override)

connorjchen:
Made `tests/integration/transactions/test_check_create.py` test case have its own wallet to avoid INSUFFICIENT_RESERVE failure. Unsure why that error was happening in first place, but probably won't be issue in future if each test has its own wallet.

tests/integration/reqs/test_submit_multisigned.py
```diff
@@ -59,7 +65,7 @@ async def test_basic_functionality(self, client):
         issuer = Wallet.create()
         tx = TrustSet(
             account=WALLET.classic_address,
-            sequence=WALLET.sequence,
+            sequence=await get_next_valid_seq_number(WALLET.classic_address, client),
```
mvadari, line 68: Does the wallet even need to be here, if it's filled in further down?

tests/integration/reqs/test_submit_multisigned.py
```diff
@@ -21,15 +23,19 @@
...
         account=WALLET.classic_address,
-        sequence=WALLET.sequence,
-        last_ledger_sequence=WALLET.sequence + 10,
+        sequence=get_next_valid_seq_number_sync(
+            WALLET.classic_address, JSON_RPC_CLIENT
+        ),
+        last_ledger_sequence=get_next_valid_seq_number_sync(
+            WALLET.classic_address, JSON_RPC_CLIENT
+        )
+        + 10,
```
mvadari, line 38: This should be fetching the last ledger that passed, not using the sequence number of the wallet.

tests/integration/reqs/test_submit_only.py
```diff
@@ -1,29 +1,39 @@
...
+        ],
+    )
     async def test_basic_functionality(self, client):
+        TX = OfferCreate(
+            account=WALLET.classic_address,
+            sequence=await get_next_valid_seq_number(WALLET.classic_address, client),
+            last_ledger_sequence=await get_next_valid_seq_number(
+                WALLET.classic_address, client
+            )
+            + 10,
```
mvadari, line 27: These are filled in with `autofill`

tests/integration/reusable_values.py
```diff
@@ -18,7 +23,9 @@ async def _set_up_reusable_values():
     OFFER = await sign_and_reliable_submission_async(
         OfferCreate(
             account=WALLET.classic_address,
-            sequence=WALLET.sequence,
+            sequence=await get_next_valid_seq_number(
+                WALLET.classic_address, ASYNC_JSON_RPC_CLIENT
+            ),
```
mvadari, line 28: Isn't this autofilled anyways? So the sequence doesn't need to be filled in separately.
ditto for every other test file

xrpl/core/keypairs/ed25519.py
```diff
@@ -96,4 +96,6 @@ def _public_key_to_str(cls: Type[ED25519], key: ECPublicKey) -> str:
 
     @classmethod
     def _format_key(cls: Type[ED25519], keystr: str) -> str:
+        if len(keystr) < 64:
+            keystr = keystr.zfill(64)
```
mvadari, line 100: Is this safe to do security-wise? Or does it lessen the randomness of entropy?

xrpl/utils/ensure_classic_address.py
```diff
@@ -0,0 +1,35 @@
+"""Classic address helper utils."""
+
+from xrpl.constants import XRPLException
+from xrpl.core.addresscodec import is_valid_xaddress, xaddress_to_classic_address
+
+
+def ensure_classic_address(account: str) -> str:
```
mvadari, line 7: This should be in the address codec

xrpl/wallet/main.py
mvadari: We can't decode based on the seed prefix if they don't use a `sEd` key - an `s` key could be either algorithm.

tests/unit/asyn/wallet/test_main.py
```diff
@@ -0,0 +1,358 @@
...
+        self.assertIsInstance(wallet.private_key, str)
+        self.assertIsInstance(wallet.classic_address, str)
+        self.assertIsInstance(wallet.address, str)
+        self.assertIsInstance(wallet.seed, str)
+        self.assertTrue(wallet.public_key.startswith(ed25519_key_prefix))
+        self.assertTrue(wallet.private_key.startswith(ed25519_key_prefix))
+        self.assertTrue(wallet.classic_address.startswith(classic_address_prefix))
+
+    def test_generate_using_algorithm_ed25519(self):
+        wallet = Wallet.create()
```
JST5000, line 94: This should define ed25519 explicitly (otherwise it's identical to above)

tests/unit/asyn/wallet/test_main.py
```diff
@@ -0,0 +1,358 @@
...
+        self.assertIsInstance(wallet.public_key, str)
+        self.assertIsInstance(wallet.private_key, str)
+        self.assertIsInstance(wallet.classic_address, str)
+        self.assertIsInstance(wallet.address, str)
+        self.assertIsInstance(wallet.seed, str)
+        self.assertTrue(wallet.private_key.startswith(secp256k1_private_key_prefix))
+        self.assertTrue(wallet.classic_address.startswith(classic_address_prefix))
+
+    def test_from_seed_using_default_algorithm(self):
+        wallet = Wallet.from_seed(seed_constants["seed"])
```
JST5000, line 117: This should also have the right seed

JST5000: Ditto to below variations

tests/unit/asyn/wallet/test_main.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def test_from_entropy_using_algorithm_ecdsa_secp256k1(self):
+        wallet = Wallet.from_entropy(
+            entropy_constants["entropy"], algorithm=CryptoAlgorithm.SECP256K1
+        )
+
+        self.assertEqual(wallet.public_key, entropy_constants["secp256k1_public_key"])
+        self.assertEqual(wallet.private_key, entropy_constants["secp256k1_private_key"])
+
+    def test_from_entropy_using_algorithm_ed25519(self):
```
JST5000, line 292: You can probably simplify these tests by splitting out a helper function which does the checks with parameters like `algorithm` + `constants` 

(Instead of `entropy_constants["secp256k1_public_key"]` you could organize it like `entropy_constants["secp256k1"]["public_key"]` which would make it easy to scalably dereference with those two parameters :) )

JST5000: General pattern: If you're specifying two values in a name, may be better to have those both be indices

tests/unit/asyn/wallet/test_main.py
```diff
@@ -0,0 +1,358 @@
...
+    def test_from_entropy_using_regular_key_pair_using_default_algorithm(self):
+        wallet = Wallet.from_entropy(
+            entropy_constants["entropy"],
+            master_address=regular_key_pair_constants["master_address"],
+        )
+
+        self.assertEqual(wallet.public_key, entropy_constants["ed25519_public_key"])
+        self.assertEqual(wallet.private_key, entropy_constants["ed25519_private_key"])
+
+    def test_from_entropy_using_regular_key_pai_using_algorithm_ed25519(self):
```
JST5000, line 309:
```suggestion
    def test_from_entropy_using_regular_key_pair_using_algorithm_ed25519(self):
```

JST5000: Small typo (missing an `r` in `key_pair`)

xrpl/core/keypairs/ed25519.py
JST5000: I would recommend just throwing an error if they give less entropy than desired. (Or using a securely random number generator to fill the rest, but I'm not sure which random generators are considered secure)

xrpl/wallet/main.py
connorjchen: @mvadari How does decode_seed work right now if it is trying to figure out which algorithm to use? From the code, it seems like it just tries prefixes and see if it comes out to a valid outcome. For an `s` key, does this usually default to a ED25519 algorithm based on order of algorithms it tries?

tests/integration/reqs/test_submit_multisigned.py
connorjchen: Not filled later down. Also, this test case is being changed drastically in multisign PR.

xrpl/wallet/main.py
mvadari: We misunderstood how seeds work when it was originally written - we thought that ED25519 keys had to start with `sEd` and secp256k1 keys just started with `s`. It turns out that this is not correct - ED25519 keys can be encoded with `sEd`, but they can also be encoded with just `s`. The algorithm is only really used to get the keys from the seed; the seed doesn't need to be encoded in any specific way. 

For example, when you call `wallet_propose` on rippled and request an ED25519 key, this is what you get:
{'account_id': 'r92k4fsv4ehYk3GzFRL1gigRk2aie8ZRkR',
 'key_type': 'ed25519',
 'master_key': 'BAIT RULE LIMB WARN IONS BAIL GAUR CORK COMB VOID SIT MARS',
 'master_seed': 'shEdVvJUB7a3eq7TTFVxfLpjEwKb7',
 'master_seed_hex': '6FC9A37AA969ED89A1ECD97BBE821A51',
 'public_key': 'aKNq5Z4GNoKc9gpxUiJkiuNjBjRf7PvhcQ6weAwCtcjZQodjviXT',
 'public_key_hex': 'ED09B678EFB1B02B8C3372FC1B31E0562508983DAB9F4E53464A51F47E93D5C4F8'}

So the somewhathacky way that it's written right now (to maintain backwards compatibility) is basically this:
```py
if algorithm is not None:
  if key.startswith(`sEd`) and algorithm == 'secp256k1':
    raise Exception()
  # encode with the algorithm provided
if key.startswith(`sEd`):
  # use ed25519
else:
  # use secp256k1
```

tests/integration/reqs/test_submit_multisigned.py
mvadari: Whoops, I meant sequence here, not wallet

xrpl/wallet/main.py
connorjchen: @mvadari So for future, are you fine with just defaulting everything to ed25519 unless otherwise specified?

mvadari: Yes, I think that would be reasonable - switch the logic to default to ed25519 unless otherwise specified. But this has to be really well-documented, because it'll be difficult to debug.

xrpl/core/keypairs/ed25519.py
connorjchen: I added this because one of my from_entropy test cases was giving back a private key of `EDB6CBAC838DFE7F47EA1BD0DF00EC282FDF45510C92161072CCFB84035390C4D` while xrpl.js gives back `ED0B6CBAC838DFE7F47EA1BD0DF00EC282FDF45510C92161072CCFB84035390C4D` (note the leading zero after `ED` prefix. My thought was that xrpl-py was cutting off the leading zero somewhere so it should be filled. I do not know if it possible to generate keys with less than 63 characters, so I don't think randomness is concern.

tests/integration/reqs/test_submit_multisigned.py
connorjchen: Sequence is not filled later on in this test. Before multisign PR gets merged, autofill isn't quite compatible with multisign.

xrpl/core/keypairs/ed25519.py
JST5000: Ah, that makes sense - good catch :)

tests/unit/asyn/wallet/test_main.py
```diff
@@ -0,0 +1,320 @@
...
+
+    if method == "regular_key_pair" or is_entropy_with_regular_key_pair:
+        self.assertEqual(
+            wallet.classic_address, constants["regular_key_pair"]["master_address"]
+        )
+    else:
+        self.assertEqual(wallet.classic_address, algorithm_constants["classic_address"])
+
+
+def _test_wallet_types(self, wallet, algorithm):
```
JST5000, line 86: This seems very similar to the other test format name. Can we be a bit more specific?

Ex.
```suggestion
def _test_wallet_encoding_algorithm(self, wallet, algorithm):
```

JST5000 (approved):
LGTM :)

tests/unit/asyn/wallet/test_main.py
connorjchen: `_test_wallet_fields` is for checking the actual values of the wallet fields

`_test_wallet_types` if for checking the actual type of the wallet fields

I don't think `_test_wallet_encoding_algorithm` makes sense here. Would it be better to change `_test_wallet_fields` to `_test_wallet_values`?

JST5000: Yep that makes sense to me

tests/integration/reqs/test_submit_only.py
```diff
@@ -7,23 +7,21 @@
...
     async def test_basic_functionality(self, client):
+        TX = OfferCreate(
+            account=WALLET.classic_address,
+            taker_gets="13100000",
+            taker_pays=IssuedCurrencyAmount(
+                currency="USD",
+                issuer=WALLET.classic_address,
+                value="10",
+            ),
+        )
```
mvadari, line 22: This transaction can still be outside the class - it'll just be autofilled and signed inside the test.

mvadari: Sidenote that doesn't need to be handled right now: this test should really use a simpler transaction with no side effects, like an AccountSet that changes a flag or something

tests/integration/sugar/test_transaction.py
```diff
@@ -415,16 +381,15 @@ async def test_reliable_submission_last_ledger_expiration(self, client):
...
+            "xrpl.ledger.get_latest_validated_ledger_sequence",
         ],
     )
     async def test_reliable_submission_bad_transaction(self, client):
-        WALLET.sequence = await get_next_valid_seq_number(ACCOUNT, client)
         payment_dict = {
             "account": ACCOUNT,
-            "last_ledger_sequence": WALLET.sequence + 20,
             "fee": "10",
+            "last_ledger_sequence": await get_latest_validated_ledger_sequence(client)
```
mvadari, line 391: Does autofill not fill in a `last_ledger_sequence`? I thought it did....

xrpl/wallet/main.py
```diff
@@ -16,69 +16,141 @@ class Wallet:
...
-        wallet.
-        Must be updated by the user. Increments on the ledger with every successful
-        transaction submission, and stays the same with every failed transaction
-        submission.
+        The core value that is used to derive all other information about
+        this wallet. MUST be kept secret!
         """
 
     @classmethod
     def create(
```
mvadari, line 76: Should this be renamed to `generate` to mimic xrpl.js? 

I think it's fine to leave this as `create`, so that it's the same as previous.

xrpl/wallet/main.py
```diff
@@ -16,69 +16,141 @@ class Wallet:
...
+
+        if entropy is not None and len(entropy) != 32:
+            raise XRPLException(
+                "Entropy must be a 16-byte hexadecimal string of random numbers."
+            )
+
+        seed = generate_seed(parsed_entropy, algorithm)
+        return Wallet.from_seed(
+            seed, master_address=master_address, algorithm=algorithm
+        )
```
mvadari, line 153:
```suggestion
        if entropy is not None and len(entropy) != 32:
            raise XRPLException(
                "Entropy must be a 16-byte hexadecimal string of random numbers."
            )

        seed = generate_seed(entropy, algorithm)
        return Wallet.from_seed(
            seed, master_address=master_address, algorithm=algorithm
        )
```

xrpl/wallet/main.py
connorjchen: I mentioned it. Jackson said it was better to leave it as create for continuity.

tests/integration/sugar/test_transaction.py
connorjchen: It does. This specific transaction doesn't autofill - I think because it is expected to fail.

xrpl/wallet/main.py
connorjchen: Oops good catch, some leftover code from old testing.

mvadari (approved):

JST5000 (approved):
LGTM

PR #478, opened 2022-11-23
refactor: Make request_impl into a private function

connorjchen:

Changed `request_impl` to `_request_impl`

### Context of Change

It should not be used outside of the library - causes confusion

### Type of Change

- [ ] Documentation Updates
- [ ] Release


mvadari (approved):
Might be useful to leave this in the changelog - if someone _was_ incorrectly using `request_impl` somewhere, it'll help them identify the error.

intelliot (approved):
If someone was mistakenly using the private function, what (public) function should they use instead?

connorjchen:
> If someone was mistakenly using the private function, what (public) function should they use instead?

Users should be using `client.request` at all times. `request_impl` is what happens behind the scenes to implement specific clients.

PR #485, opened 2022-12-01
fix: fix Github Actions issues

mvadari:

This PR switches the Github Actions from running on `ubuntu-latest` (which recently switched to 22.04) to `ubuntu-20.04`, which fixes tests.

It also adds testing support for Python 3.11.

### Context of Change

Ubuntu 22.04 upgraded OpenSSL to version 3.0, which deprecated ripemd160. Github will not add support to the runner, because they want to run only the default version.

https://github.com/openssl/openssl/issues/16994
https://github.com/actions/runner-images/issues/6676

### Type of Change

## Test Plan

CI now passes.

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-20.04
 
     env:
-      PYTHON_VERSION: "3.10"
+      PYTHON_VERSION: "3.11"
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


.github/workflows/unit_test.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Install Python + Retrieve Poetry dependencies from cache
         uses: actions/setup-python@v4
         with:
-          python-version: '3.9'
+          python-version: ${{ env.PYTHON_VERSION }}
```
connorjchen, line 35: Why was it default to 3.9 before? / What is this doing?

connorjchen (approved):

.github/workflows/unit_test.yml
mvadari: If you look higher up in the script, you'll see that `PYTHON_VERSION` is set as an env var and is set to 3.10. I figured the version that linting was using was supposed to be using 3.10. Also, IMO it's better for the linter to use the most modern fully-supported version of Python, because it'll likely have the best performance.

JST5000 (approved):

PR #490, opened 2022-12-12
fix: Used RIPEMD160 algorithm from PyCrypto rather than hashlib

connorjchen:

Used RIPEMD160 algorithm from PyCrypto rather than hashlib since hashlib uses OpenSSL

### Context of Change

Apparently Ubuntu LTS 22.04 upgraded OpenSSL to version 3.0 which deprecated RIPEMD160. As a result, our tests fail since they require this algorithm. We have reverted tests back to Ubuntu 20.04 to make them work, but this doesn’t seem ideal. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         # we have to use 3.7.1 to get around openssl issues
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.7.1', '3.8', '3.9', '3.10', '3.11']
 
     services:
       rippled:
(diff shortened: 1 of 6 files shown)


mvadari (approved):

dangell7:
Please do not take this the wrong way. I would like to ask why the repo `pycryptodom` was chosen over a straight ripemd implementation?

https://github.com/XRPLF/xrpl-py/commit/494d6fdd867815725a10e73fdfa4f1cd850c7dbc

connorjchen:
@dangell7 That is a good point. Initially, I thought it would be better to use an well-trusted out-of-the box solution like pycryptodome. However, it might be worth just implementing it ourselves.

Would you want to make a PR for it? I found [this PR](https://github.com/3rdIteration/btcrecover/commit/63b20c9045405efa6c6105ac7d338ac365e12f3a#diff-b10f51cc528ae50553bd85f178e6df530a511b05d83b6ec09ebc1a15d2f1039d) as an example of a Python implementation

Issue #312, opened 2021-11-18
txnNotFound issue


aditya-stb:
When using send_reliable_submission , xrpl it sometimes returns txnNotFound error. This is due to the fact that it calls submit transaction and fetch result by hash in quick succession which fails to find that transaction sometimes ( https://github.com/ripple/rippled/issues/2828 )

JST5000:
Do you happen to have a stack trace of the error or code snippet which can reproduce the problem?

mvadari:
It doesn't actually call in quick succession - there is a 4-second delay between submitting the transaction and retrieving it for the first time.

JST5000:
Closing as unable to reproduce - if you manage to reproduce happy to reopen! :)

mvadari:
Reopening because this is a race condition issue, and thus hard to reproduce. Here is the PR that fixes it in xrpl.js: https://github.com/XRPLF/xrpl.js/pull/1883

brokeharvard:
Would be great if we could get a fix for this incorporated into xrpl.py.   I'm receiving the txnNotFound error for every OfferCreate transaction I submit notwithstanding them actually showing up on the ledger.

hammertoe:
I'm getting this same failure too. Code that reproduces it:

```from pprint import pprint
from xrpl.clients import JsonRpcClient
from xrpl.wallet.main import Wallet
from xrpl.models.requests import AccountNFTs
from xrpl.models.transactions import Memo, NFTokenCancelOffer, NFTokenMint
from xrpl.transaction import safe_sign_and_autofill_transaction
from xrpl.transaction import send_reliable_submission
from xrpl.utils import drops_to_xrp, hex_to_str, str_to_hex


client = JsonRpcClient("http://xls20-sandbox.rippletest.net:51234")

# rBYSPwS3zvB2wu1nb5rxvU5gSuh5qgYW8V                                                                                                                           
wallet = Wallet("sh69ae5M5k7JhYpMmCoPVAz27456p", 0)

mint_args = {
    "account": "rBYSPwS3zvB2wu1nb5rxvU5gSuh5qgYW8V",
    "flags": 8,
    "uri": "",
    "memos": [Memo.from_dict({"memo_data": str_to_hex("Minted by Dealer")})],
    "transfer_fee": 0,
    "token_taxon": 0,
}

mint = NFTokenMint.from_dict(mint_args)
my_tx_payment_signed = safe_sign_and_autofill_transaction(mint, wallet, client)
[...]

PR #500, opened 2023-01-06
add check for txnNotFound

jonathanlei:
solving txnNotFound issue https://github.com/XRPLF/xrpl-py/issues/312 due to racing condition - similar solution to JS  https://github.com/XRPLF/xrpl.js/pull/1883

### Context of Change
add a condition to check for exceptions - If the exception is  `txnNotFound`, rerun the function with sleeper activated to wait for mainnet queue to clear up. 

### Type of Change

- [X ] Bug fix (non-breaking change which fixes an issue)

## Test Plan
Integration and unit tests passed but haven't reproduced the spotty bug yet

### DIFF
--- a/xrpl/asyncio/transaction/reliable_submission.py
+++ b/xrpl/asyncio/transaction/reliable_submission.py
@@ -43,7 +43,8 @@ async def _wait_for_final_transaction_outcome(
         if e.error == "txnNotFound" and attempts < 4:
 
             """
-            For the case if a submitted transaction is still in queue and not processed on the ledger yet.
+            For the case if a submitted transaction is still
+            in queue and not processed on the ledger yet.
             Retry 4 times before raising an exception.
             """
             return await _wait_for_final_transaction_outcome(
--- a/xrpl/asyncio/transaction/reliable_submission.py
+++ b/xrpl/asyncio/transaction/reliable_submission.py
@@ -42,10 +42,10 @@ async def _wait_for_final_transaction_outcome(
     except XRPLRequestFailureException as e:
         if e.error == "txnNotFound" and attempts < 4:
 
-            """err code for if the txn is not found on the ledger due to
-            when the transaction is still in queue and tx command is not
-            able to find it reason being it is not processed and hence
-            doesn't reflect on rippled databases."""
+            """
+            For the case if a submitted transaction is still in queue and not processed on the ledger yet.
+            Retry 4 times before raising an exception.
+            """
             return await _wait_for_final_transaction_outcome(
                 transaction_hash, client, prelim_result, attempts + 1
             )
--- a/xrpl/asyncio/transaction/reliable_submission.py
+++ b/xrpl/asyncio/transaction/reliable_submission.py
@@ -23,7 +23,7 @@ class XRPLReliableSubmissionException(XRPLException):
 
 
 async def _wait_for_final_transaction_outcome(
-    transaction_hash: str, client: Client, prelim_result: str, attempts: int
+    transaction_hash: str, client: Client, prelim_result: str, attempts: int = 0
 ) -> Response:
     """
     The core logic of reliable submission.  Polls the ledger until the result of the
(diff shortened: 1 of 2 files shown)


xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -34,8 +35,18 @@ async def _wait_for_final_transaction_outcome(
...
     # query transaction by hash
-    transaction_response = await client._request_impl(Tx(transaction=transaction_hash))
-
+    try:
+        transaction_response = await client._request_impl(
+            Tx(transaction=transaction_hash)
+        )
+    except XRPLRequestFailureException as e:
+        if e.error == "txnNotFound":
+            # err code for if the txn is not found on the ledger due to racing condition
```
mvadari, line 44:
```suggestion
            # err code for if the txn is not found on the ledger due to race condition
```

mvadari (approved):
Please edit the changelog.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -34,8 +35,18 @@ async def _wait_for_final_transaction_outcome(
...
+    try:
+        transaction_response = await client._request_impl(
+            Tx(transaction=transaction_hash)
+        )
+    except XRPLRequestFailureException as e:
+        if e.error == "txnNotFound":
+            # err code for if the txn is not found on the ledger due to race condition
+            return await _wait_for_final_transaction_outcome(
+                transaction_hash, client, prelim_result
+            )
```
khancode, line 47: Can this result in infinite calls? What if the error message doesn't change and remains `txnNotFound`?

mvadari: The `sleep` restricts it to one call per second. It _should_ eventually change, but I guess it's hard to test the "stuck in the queue" scenario (and I don't know if that shows up on other nodes, but I _think_ it does).

jonathanlei: this could - we can add a check to stop after couple calls and return error

jonathanlei: I don't think the JS solutions covers the scenario

khancode: Yeah, I think it's best to have a maximum attempts check and error out if the status doesn't change.

jonathanlei: might have to add an additional arg to `_wait_for_final_transaction_outcome` to keep track of attempts, how many attempts do you think we should allow?

khancode: 3 or 4 sounds like a good number if restricting it to one call per second.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -22,7 +23,7 @@ class XRPLReliableSubmissionException(XRPLException):
 
 
 async def _wait_for_final_transaction_outcome(
-    transaction_hash: str, client: Client, prelim_result: str
+    transaction_hash: str, client: Client, prelim_result: str, attempts: int
```
JST5000, line 26: I would recommend defaulting `attempts` to 0 and then you can increment from there (Otherwise you have an undocumented requirement that users who call the function set it to 0)

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -34,8 +35,18 @@ async def _wait_for_final_transaction_outcome(
...
     # query transaction by hash
-    transaction_response = await client._request_impl(Tx(transaction=transaction_hash))
-
+    try:
+        transaction_response = await client._request_impl(
+            Tx(transaction=transaction_hash)
+        )
+    except XRPLRequestFailureException as e:
+        if e.error == "txnNotFound" and attempts < 4:
+            # err code for if the txn is not found on the ledger due to race condition
```
JST5000, line 44: What is the race condition? (Is it the ledger hasn't closed yet? That this validator doesn't know about our transaction?)

If you don't know that's fine, but if you do I would recommend including that context in this comment.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -34,8 +35,22 @@ async def _wait_for_final_transaction_outcome(
...
+        transaction_response = await client._request_impl(
+            Tx(transaction=transaction_hash)
+        )
+    except XRPLRequestFailureException as e:
+        if e.error == "txnNotFound" and attempts < 4:
+
+            """err code for if the txn is not found on the ledger due to
+            when the transaction is still in queue and tx command is not
+            able to find it reason being it is not processed and hence
+            doesn't reflect on rippled databases."""
```
khancode, line 48:
```suggestion
            """
            For the case if a submitted transaction is still in queue and not processed on the ledger yet.
            Retry 4 times before raising an exception.
            """
```

khancode (approved):
Left a comment, otherwise LGTM.

PR #518, opened 2023-02-15
add timeout to faucet

dangell7:

My faucet is slow and this PR adds a timeout to the faucet post request.

### Context of Change

I used a default timeout of 5 however there are some sub-settings we could add if we wanted to get more fine grained with the timeout.

connect: float | UnsetType | None = UNSET,
read: float | UnsetType | None = UNSET,
write: float | UnsetType | None = UNSET,

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/xrpl/asyncio/wallet/wallet_generation.py
+++ b/xrpl/asyncio/wallet/wallet_generation.py
@@ -140,7 +140,11 @@ async def _check_wallet_balance(address: str, client: Client) -> int:
 
 async def _request_funding(url: str, address: str) -> None:
     async with httpx.AsyncClient() as http_client:
-        response = await http_client.post(url=url, json={"destination": address})
+        response = await http_client.post(
+            json={"destination": address},
+            timeout=httpx.Timeout(5),
+            url=url,
+        )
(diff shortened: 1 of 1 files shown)


jonathanlei:
I am actually working on adding a memos params (with default values) to all faucet transactions in client libraries, this PR is highly related, would you mind if I base my PR on yours? thanks

dangell7:
> I am actually working on adding a memos params (with default values) to all faucet transactions in client libraries, this PR is highly related, would you mind if I base my PR on yours? thanks

Not at all :)

JST5000:
This didn't end up being added as part of the memos change - @dangell7 I think this would make sense as an optional parameter, but not as a default timeout value (because I'm not sure other users want a timeout, and picking a number that works for everyone with sidechain setups seems hard to do). 

If you don't want to add that, we can close this potentially.

mvadari:
@dangell7 FYI the Ripple faucet repo is public if you'd rather use that (it might be faster)
https://github.com/xpring-eng/testnet-faucet

PR #522, opened 2023-02-18
Add disallow_incoming flags

JST5000:
## High Level Overview of Change

Adding support for this amendment https://github.com/XRPLF/rippled/pull/4336

### Context of Change

Title says it all

### Type of Change

- [X] New feature (non-breaking change which adds functionality)

## Test Plan

CI Passes

PR #524, opened 2023-02-20
Fix a mix up with AccountSet flags

rikublock:

- rename AccountSetFlags -> AccountSetAsfFlags (this could be considered a breaking change)
- add AccountSetTfFlags
- update AccountSetFlagInterface to use AccountSetTfFlags
- update tests accordingly

### Context of Change

`AccountSet` has two types of flags.
- "AccountSet Flags" (prefixed asf), used in `SetFlag`/`ClearFlag` field, limited to one per transaction
- "AccountSet Transaction Flags" (prefixed tf), used in common `Flags` field, multiple possible (bit-flags)

Reference:
- https://xrpl.org/accountset.html#accountset-flags
- https://xrpl.org/transaction-common-fields.html

Similar to the implementation found in the xrpl.js framework, this PR creates an enum for each type of flag.

### Type of Change

- [ ] Bug fix (possibly breaking change which fixes an issue)

## Test Plan
- only ran `poetry run poe test_unit`
- needs better/more testing (and extensive review)


mvadari:
Why is this PR necessary?

ckniffen:
Nice catch!

I am in favor of  the tf flags and `TX_FLAGS` which currently has completely incorrect flags listed. That said I don't think `AccountSetFlags` needs to be renamed.

rikublock:
> Why is this PR necessary?

I think the current implementation is confusing and faulty.

Let me showcase this by example (code take from unit tests)
https://github.com/XRPLF/xrpl-py/blob/v1.7.0/tests/unit/models/transactions/test_better_transaction_flags.py#L20
```py
        actual = models.AccountSet(
            account=ACCOUNT,
            flags=models.AccountSetFlagInterface(
                asf_account_tx_id=True,
                asf_authorized_nftoken_minter=True,
                asf_default_ripple=True,
                asf_deposit_auth=True,
                asf_disable_master=True,
                asf_disallow_xrp=True,
                asf_global_freeze=True,
                asf_no_freeze=True,
                asf_require_auth=True,
                asf_require_dest=True,
            ),
        )
```

As outlined before there are two kinds of AccountSet flags, the `asf` and `tf` flags. The somewhat special `asf` flags can take integer values from 1 to 10 (these are just numbers and do represent bits). At most one `asf` flags can be set per transaction. 

The above code has several issues:
-  a `FlagInterface` essential represent bits;  doing bit operations on `asf` flags makes no sense
- the `flags` parameter accepts `tf` transaction flags, assigning the `asf` gibberish produced with the `AccountSetFlagInterface` makes even less sense

rikublock:
> Nice catch!
> 
> I am in favor of the tf flags and `TX_FLAGS` which currently has completely incorrect flags listed. That said I don't think `AccountSetFlags` needs to be renamed.

I think the `FlagInterface` changes are also very much necessary.

In regards to naming. Currently the pattern used throughout the library for naming flags seems to be tx name + 'Flag'. For example:
- `TrustSetFlag` for TrustSet
- `OfferCreateFlag` for OfferCreate

Following that would result in:
- `AccountSetFlag` for the `tf` transaction flags
- `AccountSetAsfFlag` for the  `asf` flags (that's the outlier, should probably also get the outlier name)

The only drawback/consideration would have be that we are "switching" the flags that were defined in `AccountSetFlag` (until now `asf`).

rikublock:
Assuming this PR moves forward eventually, I will have to make some changes:
- someone (@mvadari ?) has to make a decision regarding the enum names (see above)
- will need a rebase to resolve some minor conflicts introduced in #522  
- will need  minor linting fixes

justinr1234:
@rikublock would you be able to make those changes now?

@mvadari do you have an opinion on the enum names?

ckniffen:
> * `AccountSetFlag` for the `tf` transaction flags
> * `AccountSetAsfFlag` for the  `asf` flags (that's the outlier/special case, should probably also get the outlier name)

I like these names and agree with the logic.

rikublock:
> @rikublock would you be able to make those changes now?
I'll take a look at it this week.

xrpl/models/transactions/account_set.py
```diff
@@ -90,28 +92,57 @@ class AccountSetFlag(int, Enum):
...
+    The same as SetFlag: asfDisallowXRP.
+    """
+
+    TF_ALLOW_XRP = 0x00200000
+    """
+    The same as ClearFlag: asfDisallowXRP.
+    """
+
+
+class AccountSetFlagInterface(FlagInterface):
```
JST5000, line 134: Instead of replacing the `AccountSetFlagInterface` with tf flags, I believe we should have two objects. 

`AccountSetFlagAsfInterface` and `AccountSetFlagTfInterface` (open to other names as I know the acronym breaks down literally into `AccountSetFlagAccountSetFlagInterface` but the `asf` is distinctive so worth including in my opinion. 

They should both then allow for easy translation between the boolean `Interface` version and the `Flags` versions. 

I don't think we should have a generic `AccountSetFlagInterface` as I think that would be confusing. i'd prefer people explicitly choose `asf` or `tf`

JST5000: Oh sorry, I see the above conversation about naming. I agree with the defaults above.

JST5000:
As a rough plan of action, I think these changes should be included in xrpl-py 2.0, so I'll take it as an action item today/tomorrow to get the xrpl-py 2.0 branch up to date so we can point it at that branch without a lot of unrelated changes appearing in the changelog :)

rikublock:
I added the outstanding changes and fixes. I also split the commits to make each individual change easier to understand/review. 

> As a rough plan of action, I think these changes should be included in xrpl-py 2.0, so I'll take it as an action item today/tomorrow to get the xrpl-py 2.0 branch up to date so we can point it at that branch without a lot of unrelated changes appearing in the changelog :)

Sounds good. I agree, it might be best to include this in a major release as the PR contains somewhat breaking changes.

tests/unit/models/transactions/test_account_set.py
ckniffen, line 1: Should a test be added for the `AccountSetFlag`?  I get that doing it that way is deprecated but it is still used since it allows for multiple flags to be set in one tx.

JST5000: I think that's actually backwards. I think the `AccountSetFlag` options are the ones that support new features & can only be done one at a time, whereas the `AccountSetTfFlags` allow for multiple flags to be set in one tx (for a limited set of flags)

JST5000: (Source: https://xrpl.org/accountset.html#accountset-flags)

JST5000: Nevermind, I was confused because I thought `AccountSetFlags` -> asf, when we decided it would go to `tf`.

ckniffen (approved):

JST5000:
@rikublock Ah, didn't realize you were changing things at the moment - I just was rebasing to try to fix the merge conflicts. It seems I have included all the above commits in the final version. Are there any other changes you were hoping to make before we merge it?

JST5000:
I think I may have misinterpreted the Git history - if there are changes we need to make we can do a follow up PR :)

rikublock:
@JST5000 As far as I am concerned this PR included all the necessary changes.

PR #528, opened 2023-02-24
feat: add submit and wait for transaction submission

pdp2121:

Currently, user has to call 2 functions (`autofill_and_sign` and `send_reliable_submission`) in order to sign, autofill, submit and verifies that the transaction has been included in a validated ledger. This PR would simplify the workflow to a single call (`submit_and_wait`) similar to xrpl.js (the old workflow would still work so it is not a breaking change). Tests and snippets are also fixed to use the new method.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +114,32 @@ async def send_reliable_submission(
     return await _wait_for_final_transaction_outcome(
         transaction_hash, client, prelim_result, 0
     )
+
+
+async def submit_and_wait(
```
mvadari, line 119: The xrpl.js version of this transaction has support for signing and submitting or just signing - I'd rather this method do the same thing.

pdp2121: https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/sugar/submit.ts#L50-L93
The submitAndWait in xrpl.js will submit the transaction regardless, or am I missing something here? Probably just add the `opt` variable and handling?

mvadari: Sorry, I misspoke. `submitAndWait` in xrpl.js has two modes - just submit an already signed transaction, or autofill/sign/submit. This method should support both.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +116,83 @@ async def send_reliable_submission(
...
+        await _check_fee(transaction, client)
+
+    if autofill:
+        transaction = await _autofill(transaction, client)
+    return await sign(transaction, wallet, False)
+
+
+async def submit_and_wait(
+    transaction: Transaction,
+    wallet: Wallet,
```
JST5000, line 171: `wallet` should be an optional parameter. It's only needed if the given transaction is NOT signed already. (For transactions that are signed before being passed in we can just submit them without needing this function to know about the user's secret keys :) )

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +116,83 @@ async def send_reliable_submission(
...
+    if check_fee:
+        await _check_fee(transaction, client)
+
+    if autofill:
+        transaction = await _autofill(transaction, client)
+    return await sign(transaction, wallet, False)
+
+
+async def submit_and_wait(
+    transaction: Transaction,
```
JST5000, line 170: If a transaction is signed, it'll be a string, so this type should be `Union[Transaction, str]`. (Signed transactions are usually in binary string format, aka the "tx_blob")

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +116,83 @@ async def send_reliable_submission(
...
+    ) or (transaction.txn_signature is not None and len(transaction.txn_signature) > 0)
+
+
+async def _get_signed_tx(
+    transaction: Transaction,
+    wallet: Wallet,
+    client: Client,
+    check_fee: bool = True,
+    autofill: bool = True,
+) -> Transaction:
```
JST5000, line 142: This should return a string (the binary blob) unless this is using a different `sign` method than the one I'm thinking of.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +124,108 @@ async def send_reliable_submission(
...
+        autofill: an optional boolean indicating whether to autofill the
+            transaction. Defaults to True.
+
+    Returns:
+        The signed transaction.
+    """
+    if type(transaction) is str:
+        transaction = _decode_tx_blob(transaction)
+
+    transaction = cast(Transaction, transaction)
```
mvadari, line 183: I believe this cast is unnecessary if you replace the if with `if isinstance(transaction, str):` (which is also the more Pythonic way to check type)

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +124,108 @@ async def send_reliable_submission(
...
+
+    if _is_signed(transaction):
+        return transaction
+
+    if not wallet:
+        raise XRPLGetSignedTransactionException(
```
mvadari, line 189: This doesn't need to have its own exception - it can just be a standard `XRPLException`

snippets/multisign.py
```diff
@@ -29,12 +23,9 @@
     signer_quorum=2,
     signer_entries=signer_entries,
 )
-signed_signer_list_set_tx = autofill_and_sign(signer_list_set_tx, master_wallet, client)
 
-print("Constructed SignerListSet and submitting it to the ledger...")
-signed_list_set_tx_response = send_reliable_submission(
-    signed_signer_list_set_tx, client
-)
+print("Constructing SignerListSet and submitting it to the ledger...")
```
mvadari, line 27:
```suggestion
print("Constructing SignerListSet and autofilling, signing, and submitting it to the ledger...")
```

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -1,17 +1,22 @@
...
 import asyncio
+from typing import Optional, Union
 
 from typing_extensions import Final
 
 from xrpl.asyncio.clients import Client
 from xrpl.asyncio.ledger import get_latest_validated_ledger_sequence
-from xrpl.asyncio.transaction.main import submit
+from xrpl.asyncio.transaction.main import _check_fee
+from xrpl.asyncio.transaction.main import autofill as _autofill
```
mvadari, line 11: Why are these imported like this?

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +118,108 @@ async def send_reliable_submission(
...
+    if _is_signed(transaction):
+        return transaction
+
+    if not wallet:
+        raise XRPLException(
+            "Wallet must be provided when submitting an unsigned transaction"
+        )
+
+    if check_fee:
+        await _check_fee(transaction, client)
```
mvadari, line 187: Isn't `check_fee` already run in `autofill`?

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +118,109 @@ async def send_reliable_submission(
...
+    return await sign(transaction, wallet, False)
+
+
+async def submit_and_wait(
+    transaction: Union[Transaction, str],
+    client: Client,
+    wallet: Optional[Wallet] = None,
+    check_fee: bool = True,
+    autofill: bool = True,
+) -> Response:
```
mvadari, line 200: nit:
```suggestion
    *,
    check_fee: bool = True,
    autofill: bool = True,
) -> Response:
```
This forces the last two params to be keyword-only, which will make it easier to use IMO.

xrpl/asyncio/transaction/reliable_submission.py
pdp2121: For `_check_fee`, that is the original name of the function defined in transaction. For `_autofill`, that is because there is a boolean `autofill` so I added an underscore to distinguish.

xrpl/asyncio/transaction/reliable_submission.py
pdp2121: `autofill` only auto-fills in the `fee` if the `fee` field in `transaction` is not yet defined, while the check in `check_fee` only happens when the input transaction is already pre-filled.

See the following comment in `autofill_and_sign`:
`# We do the transaction fee check here as we have the Client available.
    # The fee check will be done if transaction.fee exists. Otherwise the fee
    # will be auto-filled in autofill()`

snippets/multisign.py
```diff
@@ -29,12 +23,9 @@
     signer_quorum=2,
     signer_entries=signer_entries,
 )
-signed_signer_list_set_tx = autofill_and_sign(signer_list_set_tx, master_wallet, client)
 
-print("Constructed SignerListSet and submitting it to the ledger...")
-signed_list_set_tx_response = send_reliable_submission(
-    signed_signer_list_set_tx, client
-)
+print("Constructing SignerListSet and autofilling, signing, and submitting it to the ledger...")
```
JST5000, line 27: nit:
```suggestion
print("Constructing SignerListSet then autofilling, signing, and submitting it to the ledger...")
```

tests/integration/sugar/test_transaction.py
```diff
@@ -455,3 +457,134 @@ async def test_reliable_submission_no_last_ledger_sequence(self, client):
...
+    @test_async_and_sync(
+        globals(),
+        [
+            "xrpl.transaction.submit_and_wait",
+            "xrpl.account.get_next_valid_seq_number",
+            "xrpl.ledger.get_fee",
+        ],
+    )
+    async def test_submit_and_wait_simple(self, client):
+        WALLET.sequence = await get_next_valid_seq_number(ACCOUNT, client)
```
JST5000, line 472: Why do we need to get the sequence in these tests? (Shouldn't that already be part of `autofill`?)

JST5000: (If this is to help test cases run in parallel, I'd recommend writing at least one test case where you create a fresh wallet and autofill the sequence number because that's an important part of the workflow)

tests/integration/sugar/test_transaction.py
```diff
@@ -455,3 +457,134 @@ async def test_reliable_submission_no_last_ledger_sequence(self, client):
...
+    )
+    async def test_submit_and_wait_payment(self, client):
+        WALLET.sequence = await get_next_valid_seq_number(ACCOUNT, client)
+        payment_dict = {
+            "account": ACCOUNT,
+            "sequence": WALLET.sequence,
+            "amount": "10",
+            "destination": DESTINATION,
+        }
+        payment_transaction = Payment.from_dict(payment_dict)
```
JST5000, line 502: Why do we need to do `from_dict` here? (Can we directly create a `Payment` object instead?)

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +118,110 @@ async def send_reliable_submission(
...
+    Args:
+        transaction: the signed/unsigned transaction (or transaction blob) to
+            be submitted.
+        client: the network client with which to submit the transaction.
+        wallet: the wallet with which to sign the transaction (optional, only needed
+        if the transaction is not signed).
+        check_fee: an optional bolean indicating whether to check if the fee is
+            higher than the expected transaction type fee. Defaults to True.
+        autofill: an optional boolean indicating whether to autofill the
+            transaction. Defaults to True.
```
JST5000, line 170: This function doesn't concern itself with submitting, so updating it's docs to clarify that it's only focused on signing.
```suggestion
        transaction: the transaction or transaction blob to be signed if unsigned.
        client: the network client to autofill from.
        wallet: the wallet with which to sign the transaction (optional, only needed
        if the transaction is not signed).
        check_fee: an optional bolean indicating whether to check if the fee is
            higher than the expected transaction type fee. Defaults to True.
        autofill: an optional boolean indicating whether to autofill the
            transaction. Defaults to True.
```

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +118,110 @@ async def send_reliable_submission(
...
+    if isinstance(transaction, str):
+        transaction = _decode_tx_blob(transaction)
+
+    if _is_signed(transaction):
+        return transaction
+
+    if not wallet:
+        raise XRPLException(
+            "Wallet must be provided when submitting an unsigned transaction"
+        )
```
JST5000, line 184: I'm not clear why `Wallet` is optional - it seems like instead of throwing an error, I think it'd be better to just require that a wallet be provided to use this function. 

There's no part of the code which takes advantage of the fact that you can just not pass in a `Wallet` and the case where you don't pass in a wallet, this function basically doesn't do anything - because either the transaction is already signed (no effect) or this function throws an exception.

xrpl/asyncio/transaction/reliable_submission.py
mvadari: It's a helper function - it doesn't actually need more than the basic docstring.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +118,110 @@ async def send_reliable_submission(
     return await _wait_for_final_transaction_outcome(
         transaction_hash, client, prelim_result, 0
     )
+
+
+def _is_signed(transaction: Transaction) -> bool:
```
mvadari, line 123: Does the `Transaction` object not have some sort of `is_signed` member function? I thought it did...

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -113,3 +118,110 @@ async def send_reliable_submission(
...
+        transaction = await _autofill(transaction, client)
+    return await sign(transaction, wallet, False)
+
+
+async def submit_and_wait(
+    transaction: Union[Transaction, str],
+    client: Client,
+    wallet: Optional[Wallet] = None,
+    *,
+    check_fee: bool = True,
```
mvadari, line 199: xrpl.js also takes a `fail_hard` param - I think that should be included here.

xrpl/asyncio/transaction/reliable_submission.py
mvadari: Requiring the wallet also makes this function essentially unusable by someone using a different method (e.g. XUMM) to sign their transactions.

JST5000: In practice, this function does require a wallet to do it's job. I'm also specifically talking about `_get_signed_tx` in this case, which is a private helper function, and so shouldn't be a problem for people using a different method since they won't call it. (Or maybe I'm misunderstanding the concern?)

mvadari: The current architecture of the `submit_and_wait` method involves everything going through `_get_signed_tx`, even already-signed transactions. So that's certainly doable, but it would involve some reworking.

pdp2121: Both `submit_and_wait` and `_get_signed_tx` in xrpl.js have `Wallet` as an optional param, only required when the transaction is unsigned. I think updating the docstring to notify user and having this error message is adequate, but I'm open to suggestions.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -1,17 +1,22 @@
...
+from xrpl.asyncio.transaction.main import sign, submit
 from xrpl.clients import XRPLRequestFailureException
 from xrpl.constants import XRPLException
+from xrpl.core.binarycodec.main import decode
 from xrpl.models.requests import Tx
 from xrpl.models.response import Response
 from xrpl.models.transactions.transaction import Transaction
+from xrpl.wallet.main import Wallet
 
 _LEDGER_CLOSE_TIME: Final[int] = 1
```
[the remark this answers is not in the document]
pdp2121: I believe so, and it will be a breaking change. @JST5000 any ideas?

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -1,17 +1,22 @@
...
 
 import asyncio
+from typing import Optional, Union
 
 from typing_extensions import Final
 
 from xrpl.asyncio.clients import Client
 from xrpl.asyncio.ledger import get_latest_validated_ledger_sequence
-from xrpl.asyncio.transaction.main import submit
+from xrpl.asyncio.transaction.main import _check_fee
```
mvadari, line 10: `_check_fee` should probably have its underscore removed - it's being imported somewhere, which means it's no longer a private function.

xrpl/asyncio/transaction/reliable_submission.py
mvadari: IMO this would be better suited as an instance method on the `Transaction` class.

mvadari (commented):
Please update the changelog

xrpl/asyncio/transaction/reliable_submission.py
xrpl-grants: If I changed it to `check_fee`, there would be a confusion with the `check_fee` boolean causing an error.

pdp2121: If I make the change there will be a confusion with `check_fee` boolean when the function is called, causing an error.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `getNFTokenID` to get the NFTokenID after minting a token.
 - Added `LedgerEntryType` enum and added `type` field to `Ledger` and `LedgerData` requests
 - Added the algorithm used to encode a wallet's seed to the wallet.
+- Added `submit_and_wait` to sign (if needed), autofill, submit a transaction and wait for its final outcome
```
mvadari, line 25:
```suggestion
- Added `submit_and_wait` to autofill, sign, and submit a transaction and wait for its final outcome
```

mvadari: The changelog should also have a note about `send_reliable_submission` now accepting a `fail_hard` param

mvadari: Also, this needs to move to the `Unreleased` section

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [[Unreleased]]
+### Added:
+- Added `submit_and_wait` to sign (if needed), autofill, submit a transaction and wait for its final outcome
+- `submit` and `send_reliable_submission` now accepting a boolean param `fail_hard` (default to `False`; if `True`, halt the submission
```
JST5000, line 11: nit:
```suggestion
- `submit` and `send_reliable_submission` now accept an optional boolean param `fail_hard` (if `True` halt the submission if it's not immediately validated)
```

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+if fails locally)
```
JST5000, line 12: I think this wording is a bit more clear - although maybe I'm misunderstanding what "fails locally" means?

```suggestion
- `submit` and `send_reliable_submission` now accept an optional boolean param `fail_hard` (if `True` halt the submission if it's not immediately validated)
```

tests/integration/sugar/test_transaction.py
```diff
@@ -354,3 +356,122 @@ async def test_reliable_submission_no_last_ledger_sequence(self, client):
...
+            destination=DESTINATION,
+        )
+        await accept_ledger_async(delay=1)
+        response = await submit_and_wait(payment_transaction, client, WALLET)
+        self.assertTrue(response.result["validated"])
+        self.assertEqual(response.result["meta"]["TransactionResult"], "tesSUCCESS")
+        self.assertTrue(response.is_successful())
+        self.assertEqual(response.result["Fee"], await get_fee(client))
+        WALLET.sequence += 1
+
```
JST5000, line 404: There should probably be a test for "fail hard" as well if that's a new feature :)

xrpl/asyncio/transaction/main.py
```diff
@@ -147,13 +147,17 @@ async def autofill_and_sign(
...
+    *,
+    fail_hard: bool = False,
 ) -> Response:
     """
     Submits a transaction to the ledger.
 
     Args:
         transaction: the Transaction to be submitted.
         client: the network client with which to submit the transaction.
+        fail_hard: an optional boolean. If True, and the transaction fails locally,
```
JST5000, line 159: I'm not quite sure what this means by locally - my understanding is that this failure is due to the initial server rejecting you, and so it's not local. (Maybe better phrasing could be "fails for the initial server"?)

xrpl/asyncio/transaction/reliable_submission.py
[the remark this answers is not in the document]
JST5000: I would be ok with that. I created a ticket for it and updated our 2.0 release plan to include it.

xrpl/asyncio/transaction/reliable_submission.py
JST5000: This seems fine to me for a private helper. I originallly thought you could just move the exception up out of the function and get the same behavior without a parameter, but now I realize you need to decode it first, and check the signature, which are decently helpful to have in this function :)

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -107,3 +114,104 @@ async def send_reliable_submission(
...
+    if check_fee:
+        await _check_fee(transaction, client)
+
+    if autofill:
+        transaction = await _autofill(transaction, client)
+
+    return (
+        await sign(transaction, wallet, True)
+        if transaction.signers
+        else await sign(transaction, wallet, True)
```
JST5000, line 175: Both sides of this if-statement are the same, which I'm guessing is a bug. I think you meant to do this:
```suggestion
        await sign(transaction, wallet, True, True)
        if transaction.signers
        else await sign(transaction, wallet, True, False)
```

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -107,3 +114,104 @@ async def send_reliable_submission(
...
+    autofill: bool = True,
+    fail_hard: bool = False,
+) -> Response:
+    """
+    Signs a transaction locally, without trusting external rippled nodes (only if
+    the input transaction is unsigned; otherwise, proceeds to the next steps), submits,
+    and verifies that it has been included in a validated ledger (or has errored
+    /will not be included for some reason).
+    `See Reliable Transaction Submission
+    <https://xrpl.org/reliable-transaction-submission.html>`_
```
JST5000, line 194: * Adding context for why they may want to look up that bigger reference
* The parentheses parts are implied by the "if ..." so can be cut
* Moved the "/" to the top so it seems like punctuation, rather than part of the next word (like a /command)
```suggestion
    Signs a transaction locally if the transaction is unsigned, then submits,
    and verifies that it has been included in a validated ledger (or has errored/
    will not be included for some reason).
    `See Reliable Transaction Submission for a full explanation of this workflow
    <https://xrpl.org/reliable-transaction-submission.html>`_
```

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -107,3 +114,104 @@ async def send_reliable_submission(
...
+            be submitted.
+        client: the network client with which to submit the transaction.
+        wallet: the wallet with which to sign the transaction (optional, only needed
+            if the transaction is unsigned).
+        check_fee: an optional bolean indicating whether to check if the fee is
+            higher than the expected transaction type fee. Defaults to True.
+        autofill: an optional boolean indicating whether to autofill the
+            transaction. Defaults to True.
+        fail_hard: an optional boolean. If True, and the transaction fails locally,
+            do not retry or relay the transaction to other servers. Defaults to False.
```
JST5000, line 207: nit: I'd recommend using the same phrasing for `wallet` as for the other optional parameters (i.e. "an optional wallet with which to sign the transaction. This is only needed if the transaction is unsigned."

xrpl/models/transactions/transaction.py
```diff
@@ -371,6 +371,24 @@ def has_flag(self: Transaction, flag: int) -> bool:
...
+        """
+        if self.signers:
+            for signer in self.signers:
+                if (
+                    signer.signing_pub_key is None or len(self.signing_pub_key) <= 0
+                ) and (signer.txn_signature is None or len(signer.txn_signature) <= 0):
+                    return False
+            return True
+        return (self.signing_pub_key is not None and len(self.signing_pub_key) > 0) or (
+            self.txn_signature is not None and len(self.txn_signature) > 0
```
JST5000, line 389: I think that both of these fields have to be there for it to be properly signed, but please double check - https://xrpl.org/transaction-common-fields.html#signers-field

```suggestion
        return (self.signing_pub_key is not None and len(self.signing_pub_key) > 0) and (
            self.txn_signature is not None and len(self.txn_signature) > 0
```

xrpl/asyncio/transaction/reliable_submission.py
mvadari: Honestly, this will be easier to debug/read if it's a normal if statement instead of trying to put it all in one line

xrpl/models/transactions/transaction.py
mvadari: I'd recommend adding some test cases to make sure the logic works out as expected.

pdp2121: @JST5000 both of these fields should be there. xrpl.js PR fix:
https://github.com/XRPLF/xrpl.js/pull/2308
I will update it once we are good with the js PR and add some tests.

xrpl/asyncio/transaction/main.py
pdp2121: Just talked with Elliot, locally means locally to the initial server (once the tx fail, the server will not send it to other servers to retry), so keeping it as is similarly to xrpl.js is fine.

JST5000: I think that if you have to ask what it means we should change the wording - it should be clear to people who don't have that additional context that "if it fails for the initial server it will not send to other servers to retry"

tests/integration/sugar/test_transaction.py
pdp2121: I'm not sure how to test fail hard since its logic is implemented on rippled server. The client library only made a request to the server with a fail hard variable that has already been defined here:

https://github.com/XRPLF/xrpl-py/blob/a6469f131ea2ccc964b43d394d1ff23145cc2da5/xrpl/models/requests/submit_only.py#L32-L63

mvadari: It's not easily testable, it's not worth testing IMO.

tests/integration/sugar/test_transaction.py
```diff
@@ -354,3 +356,122 @@ async def test_reliable_submission_no_last_ledger_sequence(self, client):
...
+        payment_transaction = Payment(
+            account=ACCOUNT,
+            amount="10",
+            destination=DESTINATION,
+        )
+        payment_transaction_signed = await autofill_and_sign(
+            payment_transaction, WALLET, client
+        )
+        await accept_ledger_async(delay=1)
+        payment_transaction_signed_blob = encode(payment_transaction_signed.to_xrpl())
```
mvadari, line 451: This is perhaps a separate ticket, but `encode(tx.to_xrpl())` feels like a good helper function on the `Transaction` object (something like `tx.blob()`

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -107,3 +115,103 @@ async def send_reliable_submission(
     return await _wait_for_final_transaction_outcome(
         transaction_hash, client, prelim_result, transaction.last_ledger_sequence
     )
+
+
+def _decode_tx_blob(tx_blob: str) -> Transaction:
```
mvadari, line 120: This seems like something that should be on the `Transaction` object - something like `Transaction.from_blob`

xrpl/models/transactions/transaction.py
pdp2121: Logic modified and tests added

mvadari (approved):

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -107,3 +114,90 @@ async def send_reliable_submission(
...
+
+async def _get_signed_tx(
+    transaction: Union[Transaction, str],
+    client: Client,
+    wallet: Optional[Wallet] = None,
+    check_fee: bool = True,
+    autofill: bool = True,
+) -> Transaction:
+    """
+    Initializes a signed transaction for a submit request.
```
JST5000, line 127: Initializes is a bit vague - attempted to add more of the keywords/nuance.

```suggestion
    Sets up a transaction to be submitted by optionally autofilling, signing, and validating the fee seems reasonable.
```

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -107,3 +114,90 @@ async def send_reliable_submission(
...
+    if isinstance(transaction, str):
+        transaction = Transaction.from_blob(transaction)
+
+    if transaction.is_signed():
+        return transaction
+
+    if not wallet:
+        raise XRPLException(
+            "Wallet must be provided when submitting an unsigned transaction"
+        )
```
JST5000, line 151: nit: I'd recommend including the else even though it's not strictly necessary because the logic relies on that return to make sense. Each time I've reviewed the PR it's taken me a minute or two to remember why it's guaranteed that at this point the transaction MUST require a signature, and I think adding this else would clarify that.
```suggestion
    if transaction.is_signed():
        return transaction
    else:
      if not wallet:
          raise XRPLException(
              "Wallet must be provided when submitting an unsigned transaction"
          )
```

xrpl/asyncio/transaction/reliable_submission.py
JST5000: Should `check_fee` happen after `autofill`? (We are in control of the autofill function, but it seems to me like the validation should check the finalized version of the transaction instead of an in-between state, especially since autofill can change the fee)

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -107,3 +114,90 @@ async def send_reliable_submission(
...
+            the initial server, do not retry or relay the transaction to other
+            servers. Defaults to False.
+
+    Returns:
+        The response from the ledger.
+    """
+    signed_transaction = await _get_signed_tx(
+        transaction, client, wallet, check_fee, autofill
+    )
+    return await send_reliable_submission(
```
JST5000, line 201: Note for later: If we make `send_reliable_submission` deprecated, we should move that logic into this function, or a helper.

JST5000 (approved):
LGTM - couple nitpicks but logic looks solid as far as I can tell :)

xrpl/asyncio/transaction/reliable_submission.py
pdp2121: I think adding a comment saying "Attempts to validate, autofill, and sign a transaction when unsigned" would suffice. The `return` statement already tells that we only proceed to the remaining steps for unsigned transaction and I would prefer not to have many conditional layers in the code.

xrpl/asyncio/transaction/reliable_submission.py
pdp2121: Agree, we can mention that in the deprecate ticket.

PR #534, opened 2023-03-06
Add get_nftoken_id for after minting an NFT with NFTokenMint

JST5000:

Create a utility method to extract the NFTokenID of a newly minted NFT from the metadata of a validated NFTokenMint transaction.

### Context of Change

Right now it's very hard to get the NFTokenID of NFTs you mint, but it's very important to handling your NFTs after minting.

### Type of Change

## Test Plan

CI Passes


dangell7:
Just an fyi, if the transaction was validated on the ledger, then you could use the meta response to construct the nftrokenID without iterating through the CreatedNodes. 

Instead of iterating through the nodes you would just need to get the `MintedNFTokens` from the previous account root. That is the nft sequence. Then you already have the `Flags`, `Taxon`, `Fee`, `Issuer` from the transaction.

Not that it matters.

JST5000:
> Just an fyi, if the transaction was validated on the ledger, then you could use the meta response to construct the nftrokenID without iterating through the CreatedNodes.
> 
> Instead of iterating through the nodes you would just need to get the `MintedNFTokens` from the previous account root. That is the nft sequence. Then you already have the `Flags`, `Taxon`, `Fee`, `Issuer` from the transaction.
> 
> Not that it matters.

Interesting! I didn't realize there was enough info there to do it :)

dangell7:
> > Just an fyi, if the transaction was validated on the ledger, then you could use the meta response to construct the nftrokenID without iterating through the CreatedNodes.
> > Instead of iterating through the nodes you would just need to get the `MintedNFTokens` from the previous account root. That is the nft sequence. Then you already have the `Flags`, `Taxon`, `Fee`, `Issuer` from the transaction.
> > Not that it matters.
> 
> Interesting! I didn't realize there was enough info there to do it :)

I didn't either until I saw your commits. :)

Try [this](https://github.com/Transia-RnD/xrpl-helpers/blob/a0fad8ea68bf3f7a5689d8f9b121cd6110ff1985/ts/src/utils/utils.ts#L51)

pdp2121 (commented):
LGTM!

pdp2121 (approved):

justinr1234 (approved):

PR #539, opened 2023-03-23
fix: handle relsub errors better

mvadari:

This PR handles request errors better. 

`XRPLRequestFailureException` is only triggered in one place upstream of the `tx` request, and that's when there's a `JSONDecodeError` (which occurs with an HTTP error and will therefore never have a value of `txnNotFound`). However, the response could have type `error`, and in that case the `"error"` value would be `txnNotFound` if that response error occurs. Other response errors should error normally now, instead of failing silently and getting caught elsewhere (namely in the opaque `KeyError: 'LastLedgerSequence'` that will occur on one of the immediately-following lines).

### Context of Change

A request failed in a semi-production environment and it was really difficult to debug.

### Type of Change

## Test Plan

This is a difficult edge case to write a testcase for, since it involves a `tx` error. I replicated it locally with a special rippled setup, where the `tx` command was disabled. I wasn't able to replicate the `txnNotFound` error, since that's a very rare edge case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed:
 - Properly type the instance functions of NestedModel
-- Added additional check to `txnNotFound` error from `reliable_submission` due to race condition
-- Added `nft_offer` type in `AccountObjects`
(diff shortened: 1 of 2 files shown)


JST5000 (approved):
LGTM

khancode (approved):

PR #547, opened 2023-04-05
Remove capped attempts from reliable_submission

JST5000:

Fixing a re-emergence of the error being thrown during reliable submission, but the tx being validated fine. https://github.com/XRPLF/xrpl-py/issues/312

### Context of Change

Issue was raised when bulk minting NFTs on mainnet, there would occasionally be crashes prompting the user to have to implement in-depth retry logic even though the transaction actually succeeded in the end.

### Type of Change

## Test Plan

CI Passes


shawnxie999:
could someone specify an insanely large last_ledger_sequence field trying cause near-infinite recursive calls here?

JST5000:
> could someone specify an insanely large `last_ledger_sequence` field trying cause near-infinite recursive calls here?

Yes, although only at a rate of one call per second. This is how we handle it in xrpl.js, and I think that's the expected behavior. (Link to xrpl.js equivalent function: https://github.com/XRPLF/xrpl.js/blob/18e777b093fbcb15710a6ee457c037e51a6f1591/packages/xrpl/src/sugar/submit.ts#L126)

jonathanlei:
I did the cap change to prevent the infinite call - what if it is actually not found instead of a sequence issue?

JST5000:
> I did the cap change to prevent the infinite call - what if it is actually not found instead of a sequence issue?

@jonathanlei `LastLedgerSequence` isn't actually a sequence issue, it's the closest thing the ledger has to a timer. It's talking about the "Ledger Sequence" which should be increasing by 1 every ledger (aka every 4 seconds currently unless the ledger is halted). So, if the transaction fails quietly for whatever reason, every second we'll check what number the ledger is at, and if we still don't see it 10-20 ledgers later we will fail. 

I still need to add a test case for sending a malformed transaction, but the expected behavior as I understand it is that a fully failed transaction returns within a ledger of it being validated, and a well-formed transaction that is not being validated for whatever reason should wait until the LastLedgerSequence passes to fail.

tests/integration/reqs/test_reliable_submission.py
```diff
@@ -0,0 +1,47 @@
+from tests.integration.integration_test_case import IntegrationTestCase
```
mvadari, line 1: `test_reliable_submission` is already under `tests/integration/sugar/test_transaction.py`

mvadari:
I still don't quite understand why this change fixes the issue.

shawnxie999:
@mvadari the person who had the issue was trying to bulk mint NFTs in a multi-threaded manner. He was getting exactly the problem in the ticket, which indicates that the package did not wait long enough for his transactions to become validated. So we can either increase the number of attempts, or do what this PR does,  waiting until the specified `last_ledger_sequence` passes

tests/integration/reqs/test_reliable_submission.py
JST5000: Ah, thanks, I was wondering why I couldn't find that test. I should've asked :)

jonathanlei:
> > I did the cap change to prevent the infinite call - what if it is actually not found instead of a sequence issue?
> 
> @jonathanlei `LastLedgerSequence` isn't actually a sequence issue, it's the closest thing the ledger has to a timer. It's talking about the "Ledger Sequence" which should be increasing by 1 every ledger (aka every 4 seconds currently unless the ledger is halted). So, if the transaction fails quietly for whatever reason, every second we'll check what number the ledger is at, and if we still don't see it 10-20 ledgers later we will fail.
> 
> I still need to add a test case for sending a malformed transaction, but the expected behavior as I understand it is that a fully failed transaction returns within a ledger of it being validated, and a well-formed transaction that is not being validated for whatever reason should wait until the LastLedgerSequence passes to fail.

Please add me as a reviewer as well once you add the test cases - thanks!

mvadari:
> @mvadari the person who had the issue was trying to bulk mint NFTs in a multi-threaded manner. He was getting exactly the problem in the ticket, which indicates that the package did not wait long enough for his transactions to become validated. So we can either increase the number of attempts, or do what this PR does, wait until the specified `last_ledger_sequence` passes

Was he using multiple nodes to submit these transactions? That's the only situation where this would arise.

mvadari:
> > I did the cap change to prevent the infinite call - what if it is actually not found instead of a sequence issue?
> 
> @jonathanlei `LastLedgerSequence` isn't actually a sequence issue, it's the closest thing the ledger has to a timer. It's talking about the "Ledger Sequence" which should be increasing by 1 every ledger (aka every 4 seconds currently unless the ledger is halted). So, if the transaction fails quietly for whatever reason, every second we'll check what number the ledger is at, and if we still don't see it 10-20 ledgers later we will fail.
> 
> I still need to add a test case for sending a malformed transaction, but the expected behavior as I understand it is that a fully failed transaction returns within a ledger of it being validated, and a well-formed transaction that is not being validated for whatever reason should wait until the LastLedgerSequence passes to fail.

It's not that it fails silently, it's that transactions are essentially considered potentially-successful until they're guaranteed-successful (`tesSUCCESS`) or guaranteed-failed (it gets a `tec` code that is validated, or it is impossible for it to be successful), and it just sticks around in the transaction queue until then.

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -34,40 +34,37 @@ async def _wait_for_final_transaction_outcome(
     await asyncio.sleep(_LEDGER_CLOSE_TIME)
     # new persisted transaction
 
+    latest_ledger_sequence = await get_latest_validated_ledger_sequence(client)
+
+    if last_ledger_sequence < latest_ledger_sequence:
```
mvadari, line 39: This needs to be flipped

JST5000: This flip is actually correct (Although the names are terrible, I'll try to find better names)

When you create a transaction, you set `LastLedgerSequence` to current ledger + 20 or something. 

So in the first iteration:
`LastLedgerSequence: +20` | `latest_ledger_sequence: 0`
Which means that we skip over the error and try again later, as expected. 

20 | 1 
20 | 2
...
20 | 21 -> `LastLedgerSequence` < `latest_ledger_sequence` -> Error, waited too long.

(And this is a nit, but it's not `<=` because it's supposed to be ok to validate if the ledger is on, or before that index, so we shouldn't error if we're on an "equal to" ledger index - https://xrpl.org/reliable-transaction-submission.html#:~:text=LastLedgerSequence%20is%20an%20optional%20parameter,than%20the%20transaction's%20LastLedgerSequence%20parameter.)

JST5000: The flip happened because I had to flip the if-statement to do the error before the recursion, since the error is the base case now (and otherwise it just goes infinite, never checking the ledger index)

mvadari: I'd recommend flipping it entirely then to make the logic easier to parse - 
```py
if latest_ledger_sequence > last_ledger_sequence:
```
Since `latest_ledger_sequence` is changing and `last_ledger_sequence` is not

mvadari (approved):

xrpl/asyncio/transaction/reliable_submission.py
JST5000: I think the name `current_ledger_sequence` is better than `latest` because it's too close to `last`. I'll go with that unless that seems confusing for some reason

JST5000: (And my nit was wrong - I thought it was the index of the current ledger, but it was the last validated ledger, which means that if we already see "LastLedgerSequence" there it's too late to be included)

JST5000:
> Please add me as a reviewer as well once you add the test cases - thanks!

@jonathanlei added :) (Mayukha pointed out that the tests already existed, I just didn't see them)

xrpl/asyncio/transaction/reliable_submission.py
```diff
@@ -109,7 +105,5 @@ async def send_reliable_submission(
         )
 
     return await _wait_for_final_transaction_outcome(
-        transaction_hash,
-        client,
-        prelim_result,
+        transaction_hash, client, prelim_result, transaction.last_ledger_sequence
```
ckniffen, line 108: Why were these moved to one line yet have more params?

ckniffen (approved):

jonathanlei (approved):

PR #550, opened 2023-04-11
refactor: remove calls to deprecated methods

mvadari:
### Context of Change

The deprecation notices were clogging up the CI output, making it difficult to identify what was actually causing errors in a PR.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Refactor (non-breaking change that only restructures code)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)

## Test Plan

CI no longer has deprecation warnings.

PR #552, opened 2023-04-11
chore: remove deprecated functions from coverage

mvadari:

This PR removes deprecated methods from codecov and fixes a minor bug with `get_balance` (the `ledger_index` param wasn't being used).

### Context of Change

Follow-up from #550 

### Type of Change

## Test Plan

CI passes, the total codecov percentage for integration tests increased a bit.


ckniffen:
> If they're deprecated, doesn't that mean that we aren't maintaining them anymore?

I see deprecated more as a warning that it will not always be available so it should be changed to the new way.

ckniffen (approved):

mvadari:
> > If they're deprecated, doesn't that mean that we aren't maintaining them anymore?
> 
> I see deprecated more as a warning that it will not always be available so it should be changed to the new way.

Honestly, I think it's pretty unlikely that any change we make will impact the deprecated functions before we hit v2.0.

PR #554, opened 2023-04-11
[BREAKING] fix: XRP().to_amount now expects XRP instead of drops

mvadari:

This PR fixes a disconnect between the docs for the `XRP().to_amount` method and what the function actually did. The docs seemed to be expecting an XRP amount, but the code was expecting an amount in drops.

I think an amount in XRP makes more sense. This is technically a breaking change, but I think it's more of a bugfix.

### Context of Change

Ran into this issue while working on sidechain tooling

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed deprecated request wrappers (the preferred method is to directly do client.request instead)
 - `sign` is now synchronous instead of async (done by removing the optional `check_fee` param & moving checks up to other functions)
 - In order to be internally consistent, all signing/submitting functions will follow the parameter order of `transaction`, `client`, `wallet`, and then other parameters. (This is because `wallet` is optional for `submit_and_wait` and so must come after `client`)
+- `XRP.to_amount` now converts from XRP to drops, instead of expecting a drops amount
 
 ### Removed:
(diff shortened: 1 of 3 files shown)


ckniffen:
I think this is a huge breaking change because anyone that relied on this will be having values orders of magnitude more.

justinr1234:
@mvadari conflicts

JST5000:
@mvadari why is using XRP instead of drops better here? (I know the docs say XRP, but in practice everyone using the function currently has it in drops, so should we just update the docs and call it good?) 

We should make a choice about this before 2.0

intelliot:
Suggestion:
1. Leave the behavior of the existing function alone
2. Deprecate the existing function (and update the docs/comments to be accurate)
3. Add 2 new methods:
    - to_drops_amount
    - to_xrp_amount

Hopefully the names of the new methods make the behavior obvious 😀

mvadari:
> Suggestion:
> 
> 1. Leave the behavior of the existing function alone
> 2. Deprecate the existing function (and update the docs/comments to be accurate)
> 3. Add 2 new functions:
>    
>    * to_drops_amount
>    * to_xrp_amount
> 
> Hopefully the names of the new methods make the behavior obvious 😀

The difficulty of this is that you can't use `Currency.to_amount` as easily, which is what I wanted.

JST5000 (approved):
LGTM

ckniffen (approved):

PR #559, opened 2023-04-20
adding `nft_info`, `nft_history` and `nfts_by_issuer` clio API models

shawnxie999:
Title says all

### Type of Change

- [ ] Documentation Updates
- [ ] Release


xrpl/models/requests/request.py
```diff
@@ -58,6 +58,9 @@ class RequestMethod(str, Enum):
     # NFT methods
     NFT_BUY_OFFERS = "nft_buy_offers"
     NFT_SELL_OFFERS = "nft_sell_offers"
+    NFT_INFO = "nft_info"
```
ckniffen, line 61: Should this be documented as clio only methods?

ckniffen:
I am unfamiliar with the request api.  Do we first perform a `server_info` call?  If so we should be able to know if it is a clio connection?

We also might want to add a catch for these specific requests that has a nice message if rippled is called and returns `unknownCmd`.

shawnxie999:
> I am unfamiliar with the request api. Do we first perform a `server_info` call? If so we should be able to know if it is a clio connection?
> 
> We also might want to add a catch for these specific requests that has a nice message if rippled is called and returns `unknownCmd`.

@ckniffen No I don't think that's the case. `server_info` is never called implicitly when I tried to set breakpoints in the library code.

shawnxie999:
But I definitely think it's worth to add that in the future, so that we can differentiate between rippled vs clio-only APIs and return some kind of message if they is using the wrong server. Currently it could be pretty confusing to naive users.

xrpl/models/requests/request.py
shawnxie999: added comments

ledhed2222 (approved):

mvadari:
Has `nfts_by_issuer` been released in a Clio version yet? I'd rather not include it in the libraries before it's been released.

shawnxie999:
> Has `nfts_by_issuer` been released in a Clio version yet? I'd rather not include it in the libraries before it's been released.

Not yet, this API is still in code review

mvadari:
> > Has `nfts_by_issuer` been released in a Clio version yet? I'd rather not include it in the libraries before it's been released.
> 
> Not yet, this API is still in code review

If it can't be used yet, and it's not 100% finalized (anything can happen in code review), it shouldn't be in the libraries.

xrpl/models/requests/nft_history.py
```diff
@@ -0,0 +1,39 @@
+"""
+The `nft_history` method retreives  a list of transactions that involved the
```
ledhed2222, line 2: accidental extra space before "a"

xrpl/models/requests/nft_history.py
```diff
@@ -0,0 +1,39 @@
...
+from xrpl.models.requests.request import Request, RequestMethod
+from xrpl.models.required import REQUIRED
+from xrpl.models.utils import require_kwargs_on_init
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True)
+class NFTHistory(Request):
+    """
+    The `nft_history` method retreives  a list of transactions that involved the
```
ledhed2222, line 17: ditto

shawnxie999:
close this for now since nfts_by_issuer is not supported yet. open a new PR for the existing two APIs https://github.com/XRPLF/xrpl-py/pull/573

PR #570, opened 2023-05-03
fix: assorted fixes

mvadari:

This PR:
* Allows `keypairs.sign` to take a hex string in addition to bytes
* Fixes some issues with the `SignerListSet` validation
* Fixes the integration tests so that the transaction tests are actually run on all 4 clients, instead of just on the JSON RPC clients
* Fixes an issue in the integration tests where there weren't actually ledgers closed after each transaction

### Context of Change

These were all issues I came across while working on #571.

### Type of Change

## Test Plan

CI passes.


tests/integration/transactions/test_trust_set.py
```diff
@@ -10,7 +13,7 @@ class TestTrustSet(IntegrationTestCase):
     @test_async_and_sync(globals())
     async def test_basic_functionality(self, client):
         issuer_wallet = Wallet.create()
-        response = await submit_transaction_async(
+        response = await sign_and_reliable_submission_async(
```
JST5000, line 16: If this function is necessary now for tests involving submitting transactions, can you add it to the CONTRIBUTING as a part of writing tests?

xrpl/models/transactions/signer_list_set.py
```diff
@@ -97,11 +97,19 @@ def _get_errors(self: SignerListSet) -> Dict[str, str]:
         if self.signer_entries is None:  # deletion of the SignerList object
             return errors
 
-        if self.signer_quorum <= 0:
+        if self.signer_quorum == REQUIRED:
```
JST5000, line 100: Does this work if the object is a dictionary? (I've never seen the REQUIRED keyword before, and my attempts to google it are not working well because it's such a common word - do you know how to find the docs for it / what it specifically does?)

xrpl/models/transactions/signer_list_set.py
```diff
@@ -97,11 +97,19 @@ def _get_errors(self: SignerListSet) -> Dict[str, str]:
...
+            errors["signer_quorum"] = "`signer_quorum` is not set."
+        elif self.signer_quorum <= 0:
             errors[
                 "signer_quorum"
             ] = "`signer_quorum` must be greater than or equal to 0."
```
JST5000, line 105: This error says that `signer_quorum` must be greater than or equal to 0, but if it is 0 this error still happens. Is that a typo?

xrpl/models/transactions/signer_list_set.py
mvadari: `REQUIRED` is an xrpl-py specific type, for all the required fields. There shouldn't be any issue regardless of what type the object is.

xrpl/models/transactions/signer_list_set.py
mvadari: Hm, not sure - it appears to have been the same previously.

mvadari: Looking at the code again, I think this condition is fine - line 97 returns `errors` if `signer_entries` is `None`, which should handle that situation.

pdp2121 (approved):

xrpl/models/transactions/signer_list_set.py
JST5000: @mvadari I think updating the error to what @pdp2121 suggested makes sense here - 

Relevant quote from xrpl.org on SignerlistSet fields: (https://xrpl.org/signerlistset.html#signerlistset-fields) 
"You cannot create a signer list such that the SignerQuorum could never be met. The SignerQuorum must be greater than 0 but less than or equal to the sum of the SignerWeight values in the list. Otherwise, the transaction fails with the error [temMALFORMED](https://xrpl.org/tem-codes.html)."

JST5000 (approved):
LGTM after updating the error message (logic looks good as is, just think we can be a bit clearer)

xrpl/models/transactions/signer_list_set.py
mvadari: Error message was fixed in this commit: `e3b0f0eb9f262ed3ae3cdcbc5f6c59fc47432f3c`

PR #572, opened 2023-05-11
add user agent and use case info to faucet json body

jonathanlei:

aadd user agent and use case info to faucet transactions for analytics

### Context of Change

This is part of the testnet analytics effort.

### Type of Change
[ X] New feature (non-breaking change which adds functionality)


xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -135,7 +135,10 @@ async def _check_wallet_balance(address: str, client: Client) -> int:
 
 async def _request_funding(url: str, address: str) -> None:
```
ckniffen, line 136: I think an optional parameter of usecase would be helpful.  That way we can say 'unit test' or 'integration test'.

jonathanlei: added

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -31,6 +31,7 @@ async def generate_faucet_wallet(
     wallet: Optional[Wallet] = None,
     debug: bool = False,
     faucet_host: Optional[str] = None,
+    use_case: Optional[str] = None,
```
mvadari, line 34: Why not just call this param `user_agent`? `use_case` is a very confusing name IMO

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -133,9 +138,18 @@ async def _check_wallet_balance(address: str, client: Client) -> int:
...
 
 
-async def _request_funding(url: str, address: str) -> None:
+async def _request_funding(
+    url: str, address: str, use_case: Optional[str] = None
+) -> None:
     async with httpx.AsyncClient() as http_client:
-        response = await http_client.post(url=url, json={"destination": address})
+        headers = {"Content-Type": "application/json"}
+        if use_case:
```
mvadari, line 146:
```suggestion
        if use_case is not None:
```

xrpl/asyncio/wallet/wallet_generation.py
justinr1234: We suggested that it be moved to json instead of http header

ckniffen:
Should there be a follow up PR which sets `use_case="integration_test"` or `use_case="unit_test"`?

xrpl/asyncio/wallet/wallet_generation.py
jonathanlei: user_agent is specific to xrpl.js - use-case is pointing towards things like: workshop, int test, unit tests..

jonathanlei: > Should there be a follow up PR which sets `use_case="integration_test"` or `use_case="unit_test"`?

I didn't see a unit test that call the fundWallet function

mvadari: There aren't any

ckniffen (approved):

xrpl/asyncio/wallet/wallet_generation.py
ckniffen: > > Should there be a follow up PR which sets `use_case="integration_test"` or `use_case="unit_test"`?
> 
> I didn't see a unit test that call the fundWallet function

No but generate_faucet_wallet is called in places.  That should allow for you pass use case.

mvadari: > user_agent is specific to xrpl.js - use-case is pointing towards things like: workshop, int test, unit tests..

I'm a little confused as to what this means - `user_agent` is in the HTTP header and I think it's a pretty clear name

ckniffen:
@jonathanlei pining about adding use_case to generate_faucet_wallet

justinr1234:
@jonathanlei conflicts after latest master

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -143,9 +148,14 @@ async def _check_wallet_balance(address: str, client: Client) -> int:
         raise
 
 
-async def _request_funding(url: str, address: str) -> None:
+async def _request_funding(
+    url: str, address: str, usage_context: Optional[str] = None
+) -> None:
     async with httpx.AsyncClient() as http_client:
-        response = await http_client.post(url=url, json={"destination": address})
+        json_body = {"destination": address, "userAgent": "xrpl-py"}
```
justinr1234, line 155: I think user agent should also be an input parameter that defaults to xrpl-py but the user can change to be None if they want

justinr1234 (changes requested):

ckniffen (approved):

justinr1234 (approved):

PR #577, opened 2023-05-25
Update the docker container to use latest amendments

JST5000:

Update the docker container we use for integration tests to have the latest amendments.

### Context of Change

Our tests were speed constrained if we were using public resources, and so having an updated standalone docker container should help us improve our test coverage while still having fast tests.

### Type of Change

## Test Plan

CI Passes


.ci-config/rippled.cfg
```diff
@@ -0,0 +1,151 @@
...
+# 2. Strip away the quotes and commas
+# 3. Add the amendment name to the same line as each amendment id (You can look them up via hash on https://xrpl.org/known-amendments.html)
+#   Ex. 4C97EBA926031A7CF7D7B36FDE3ED66DDA5421192D63DE53FFB46E43B9DC8373 Multisign
+#   The amendment name can be any string (including just a number)
+#
+# Note: The version of rippled you use this config with must have an implementation for the amendments you attempt to enable or it will crash.
+
+[amendments]
+# Devnet amendments as of March 28th, 2023
+B4E4F5D2D6FB84DF7399960A732309C9FD530EAE5941838160042833625A6076 NegativeUNL
```
justinr1234, line 107: Is there a way to avoid making this manual? Perhaps we can inject it via CI/CD at build time by downloading a list from the rippled repo?

JST5000: I think it would have to look at the Devnet network to do that because the Rippled repo doesn't have information on which amendments are actually voted in. I think automating it is definitely doable. I have a separate task in our backlog though to automate it because I think it's a more involved ask. (For now I think this PR is a win because it at least gets us back up to date & not stuck on an old version which we can't update even manually)

justinr1234: Can you link the github issue for the followup work here?

JST5000: https://github.com/XRPLF/xrpl-py/issues/578

justinr1234 (approved):

mvadari (approved):

PR #589, opened 2023-06-02
docs: Deprecate the underlying send_reliable_submission

JST5000:

Title says it all.

### Context of Change

`submit_and_wait` is the newer version of that function with more features & future support.

Note: I moved the internals of `send_reliable_submission` into a private helper because `submit_and_wait` relies on that functionality, but should not trigger deprecation warnings each time it's called. (Basically separating the "public face" which is deprecated, from the helper)

### Type of Change

- [X] Documentation Updates

## Test Plan

CI Passes (and does NOT mention the deprecation - I checked beforehand to verify the deprecation was working, but don't want to clog the logs going forward)

### DIFF
--- a/tests/integration/sugar/test_transaction.py
+++ b/tests/integration/sugar/test_transaction.py
@@ -1,3 +1,5 @@
+import warnings
+
 from tests.integration.integration_test_case import IntegrationTestCase
 from tests.integration.it_utils import (
     accept_ledger_async,
@@ -58,6 +60,9 @@ class TestTransaction(IntegrationTestCase):
             amount="100",
             destination=classic_address_to_xaddress(DESTINATION, None, False),
         )
+
+        # Silence warnings from send_reliable_submission being deprecated
+        warnings.filterwarnings("ignore", category=DeprecationWarning)
 
         # WHEN we sign locally, autofill, and submit the transaction
         response = await sign_and_reliable_submission_async(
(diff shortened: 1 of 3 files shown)


xrpl/transaction/reliable_submission.py
```diff
@@ -13,6 +13,7 @@
 from xrpl.wallet.main import Wallet
 
 
+# TODO: Remove this at the same time we remove async_send_reliable_submission
 def send_reliable_submission(
```
ckniffen, line 17: Should this also be deprecated?

ckniffen (approved):

justinr1234 (approved):

mvadari (commented):
Why deprecate it and not just delete it? It's an easy enough migration for people using it who are upgrading to 2.0.

JST5000:
> Why deprecate it and not just delete it? It's an easy enough migration for people using it who are upgrading to 2.0.

The thought process was that this is the most commonly used function, and so we were worried that people wouldn't update if it was changed. 

That being said - it is just a search + replace, so I see it being fine either way.

tests/integration/sugar/test_transaction.py
```diff
@@ -59,6 +61,9 @@ async def test_none_as_destination_tag(self, client):
             destination=classic_address_to_xaddress(DESTINATION, None, False),
         )
 
+        # Silence warnings from send_reliable_submission being deprecated
```
mvadari, line 64: IMO the tests should be updated to not use `send_reliable_submission` instead of adding the silencing

ckniffen: We did update most tests. Some should remain to ensure we arent breaking deprecated code.

ckniffen: Its saying that something is going away so switch to using something else.

JST5000:
Closing since we've decided to remove `send_reliable_submission` instead of deprecating. Will open a separate PR and link here afterwards.

PR #592, opened 2023-06-07
feat: Remove `send_reliable_submission`

JST5000:

Title says it all.

### Context of Change

`submit_and_wait` now does everything `send_reliable_submission` does and more - so we want people to migrate to the new version which is receiving all our support. 

### Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Test Plan

CI Passes (I migrated over the tests which were only for `send_reliable_submission` to apply to `submit_and_wait` to keep that coverage)

* Add a link to the `submit_and_wait` reference docs in the README once that becomes public.


justinr1234 (approved):

README.md
```diff
@@ -157,12 +157,12 @@ Use the [`xrpl.transaction`](https://xrpl-py.readthedocs.io/en/stable/source/xrp
 
 * [`sign`](https://xrpl-py.readthedocs.io/en/stable/source/xrpl.transaction.html#xrpl.transaction.sign) — Signs a transaction locally. This method **does  not** submit the transaction to the XRP Ledger.
 
-* [`send_reliable_submission`](https://xrpl-py.readthedocs.io/en/stable/source/xrpl.transaction.html#xrpl.transaction.send_reliable_submission) — An implementation of the [reliable transaction submission guidelines](https://xrpl.org/reliable-transaction-submission.html#reliable-transaction-submission), this method submits a signed transaction to the XRP Ledger and then verifies that it has been included in a validated ledger (or has failed to do so). Use this method to submit transactions for production purposes.
+* `submit_and_wait` — An implementation of the [reliable transaction submission guidelines](https://xrpl.org/reliable-transaction-submission.html#reliable-transaction-submission), this method submits a signed transaction to the XRP Ledger and then verifies that it has been included in a validated ledger (or has failed to do so). Use this method to submit transactions for production purposes.
```
mvadari, line 160: Shouldn't this still link to the docs?

mvadari (commented):
Some of the snippets need to be fixed - they have an incorrect param order for `submit_and_wait`

README.md
JST5000: `submit_and_wait` isn't in the docs yet - I have it as a follow up item to add the link later. (https://xrpl-py.readthedocs.io/en/stable/search.html?q=submit_and_wait&check_keywords=yes&area=default)

JST5000:
> Some of the snippets need to be fixed - they have an incorrect param order for `submit_and_wait`

Good catch - I thought I caught them all but the `set_regular_key` one had an incorrect order.

README.md
mvadari: The link will probably be the same if you just want to put it in for now.

mvadari (approved):

README.md
JST5000: I'll just include it in a follow up PR afterwards (Have to do it in a couple other places anyway)

PR #601, opened 2023-06-16
refactor(wallet): Rename `Wallet.classic_address` to `Wallet.address`

JST5000:

`classic_address` is jargon for most people - we should instead name it `address`

### Context of Change

We already were planning on adding a new alias, but in the spirit of having a clear API, this PR is instead planning to just set `address` to the new name. 

`classic_address` is only named that to differentiate itself from `x-address` which is a newer standard which is not supported at the ledger level. (You have to decode `x-addresses` back to `classic_addresses` before submitting a transaction to the ledger) - For a concept as basic as what an account on the XRPL is, we don't want to require understanding that nuance.

### Type of Change

- [X] Refactor (non-breaking change that restructures how the code works)

## Test Plan

CI Passes

Note: Originally this was a breaking change, but after discussion we've changed it to just affect the defaults we use - while keeping `classic_address` as an alias.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ Use [`xrpl.core.addresscodec`](https://xrpl-py.readthedocs.io/en/stable/source/x
 # convert classic address to x-address
 from xrpl.core import addresscodec
 testnet_xaddress = (
-    addresscodec.address_to_xaddress(
+    addresscodec.classic_classic_address_to_xaddress(
         "rMPUKmzmDWEX1tQhzQ8oGFNfAEhnWNFwz",
         tag=0,
         is_test_network=True,
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ test_wallet = generate_faucet_wallet(client)
 print(test_wallet)
 public_key: ED3CC1BBD0952A60088E89FA502921895FC81FBD79CAE9109A8FE2D23659AD5D56
 private_key: -HIDDEN-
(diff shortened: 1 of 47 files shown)


README.md
```diff
@@ -310,7 +310,7 @@ Use [`xrpl.core.addresscodec`](https://xrpl-py.readthedocs.io/en/stable/source/x
 # convert classic address to x-address
 from xrpl.core import addresscodec
 testnet_xaddress = (
-    addresscodec.classic_address_to_xaddress(
+    addresscodec.classic_classic_address_to_xaddress(
```
ckniffen, line 313: Was `classic_classic` intentional?

xrpl/wallet/main.py
```diff
@@ -17,17 +17,12 @@ class Wallet:
     """
 
     @property
-    def classic_address(self: Wallet) -> str:
+    def address(self: Wallet) -> str:
```
justinr1234, line 20: Is there any reason not to keep the old one as an Alias?

justinr1234: Given the docs still explicitly call our a difference on xrpl.org, I think it can be more confusing to remove this altogether as opposed to showing `address` as the default and providing docs inline in the functions

JST5000: Fair enough - re-added :) (Although I think it's worthwhile to still change everything to use `address` by default)

pdp2121 (approved):

jonathanlei (approved):

ckniffen (approved):

PR #520, opened 2023-02-17
feature `NetworkID`

dangell7:
## High Level Overview of Change

https://github.com/XRPLF/rippled/pull/4370

### Context of Change

Added network id to the base transaction
Added network id to the `server_info` response
Added network id to the binary codec
Added network id to autofill if custom network id was supplied to client.

### Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Test Plan

No tests were written

PR #603, opened 2023-06-20
fix(test): move integration test out from unit tests

mvadari:

There was an integration test that was in the unit test section. This PR moves that over.

### Context of Change

The test was added in #520. I noticed that unit tests were suddenly taking much longer to run.

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/tests/integration/sugar/test_network_id.py
+++ b/tests/integration/sugar/test_network_id.py
@@ -9,6 +9,7 @@ from xrpl.wallet.wallet_generation import generate_faucet_wallet
 _FEE = "0.00001"
 
 
+# TODO: move to test_transaction and use the standard integration test setup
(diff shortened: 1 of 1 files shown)


justinr1234:
@mvadari the diff seems weird?

ckniffen (approved):

justinr1234 (approved):

dangell7:
@mvadari These test are written to specific networks to get a specific network id response. Therefore we cannot use the `IntegrationTestCase` and must use an active network. This also means we cannot use `WALLET` because we need a wallet active on that specific network. 

So I would like your opinion on what we should do?

@pdp2121 If you can elaborate at all please share. I tried to implement `IntegrationTestCase` without success.

PR #611, opened 2023-07-03
Fix invalid account address

JST5000:

Update the test on master which was failing due to an invalid address which was passing silently incorrectly.

### Context of Change

Originally found as a result of the behavior changing in rippled 1.11 - it now throws `invalidAccount` as expected on that version.

### Type of Change

## Test Plan

CI Passes


ckniffen:
Since they are different there should be a malformed address test still.

ckniffen (approved):

mvadari (approved):

JST5000:
I changed this PR to point at the 2.0 branch because it already has a similar change in place, and if this were to be committed against master it'd create a merge conflict which may require a rebase to fix. Since this is all going to master anyway and is just a test, I think it's easier to just fix it all with the 2.0 merge.

Issue #616, opened 2023-07-06
[2.0] `classic_address` isn't typed properly


mvadari:
![image](https://github.com/XRPLF/xrpl-py/assets/8029314/6555aaf0-7336-43d2-b7ff-4fc9f1d32ab5)

It looks like property aliases aren't supported in mypy right now: https://github.com/python/mypy/issues/6700

The best solution is likely to just copy-paste the `address` code, with a `TODO` for when mypy has resolved the issue.

JST5000:
For people encountering this issue before the next release, if you change from `wallet.classic_address` to `.address` the issue goes away.

ckniffen:
Closed by #617

PR #617, opened 2023-07-06
fix: Fix `classic_address` aliasing issue

JST5000:

Fix mypy complaining about `Wallet.classic_address` being a `Callable` when it is an aliased `str` property. (Open mypy issue: https://github.com/python/mypy/issues/6700)

### Context of Change

Currently aliasing a property is not properly typed by mypy. As a workaround you can switch to using `address` instead of `classic_address`, but this should resolve the issue until the mypy issue is fixed.

### Type of Change

## Test Plan

CI passes

### DIFF
--- a/xrpl/wallet/main.py
+++ b/xrpl/wallet/main.py
@@ -24,12 +24,16 @@ class Wallet:
         """  # noqa: DAR201
         return self._address
 
-    classic_address = address
-    """
-    `classic_address` is the same as `address`. It is called `classic_address` to
-    differentiate it from the x-address standard, which encodes the network,
-    destination tag, and XRPL address into a single value. It's also a base58 string.
-    """
+    # TODO: Just alias classic_address once mypy has resolved this issue:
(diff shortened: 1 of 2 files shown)


JST5000:
Closes https://github.com/XRPLF/xrpl-py/issues/616

xrpl/wallet/main.py
```diff
@@ -24,12 +24,17 @@ def address(self: Wallet) -> str:
         """  # noqa: DAR201
         return self._address
 
-    classic_address = address
-    """
-    `classic_address` is the same as `address`. It is called `classic_address` to
-    differentiate it from the x-address standard, which encodes the network,
-    destination tag, and XRPL address into a single value. It's also a base58 string.
-    """
+    # TODO: Just alias classic_address once mypy has resolved this issue:
```
mvadari, line 27: I'd recommend opening an issue for this as a reminder

mvadari (approved):

ckniffen (approved):

PR #634, opened 2023-08-22
feat: Update `SetFee` with new syntax pending XRPFee amendment

JST5000:

With the XRPFee amendment, the `SetFee` Psuedotransaction will have new fields. 

### Context of Change

Ideally, the old fields should still be usable up until the amendment is enabled, and for historical analysis of the ledger it should still be able to interpret older transactions as `SetFee` transactions. But we want to enforce that only one of the new syntaxes is allowed, not any combination of fields. (Meaning it should not be possible to instantiate a `SetFee` which has BOTH a `base_fee` AND a `base_fee_drops`)

Useful to see how this was done in xrpl.js: https://github.com/XRPLF/xrpl.js/pull/2357

### Type of Change

- [X] Documentation Updates

## Test Plan

CI Passes


xrpl/models/transactions/pseudo_transactions/set_fee.py
```diff
@@ -19,42 +20,138 @@ class SetFee(PseudoTransaction):
...
+
+    After the XRPFees Amendment, base_fee_drops, reserve_base_drops,
+    and reserve_increment_drops are required fields.
+
+    No SetFee Psuedo Transaction should contain fields from BOTH before
+    and after the XRPFees amendment.
     """
 
-    base_fee: str = REQUIRED  # type: ignore
+    def __post_init__(self: SetFee) -> None:
```
ckniffen, line 38: I don't think these checks are needed.  A consumer can never create the `SetFee` transaction or change the Fees ledger object.

ckniffen (approved):

xrpl/models/transactions/pseudo_transactions/set_fee.py
```diff
@@ -19,42 +19,74 @@ class SetFee(PseudoTransaction):
...
+    `XRPFees Amendment<https://xrpl.org/known-amendments.html#xrpfees>`_
+
+    Before the XRPFees Amendment which was proposed in rippled 1.10.0
+    base_fee, reference_fee_units, reserve_base, and reserve_increment
+    were required fields.
+
+    After the XRPFees Amendment, base_fee_drops, reserve_base_drops,
+    and reserve_increment_drops are required fields.
+
+    No SetFee Psuedo Transaction should contain fields from BOTH before
```
mvadari, line 33: *Pseudo

mvadari, line 33:
```suggestion
    No SetFee Pseudo Transaction should contain fields from BOTH before
```

xrpl/models/transactions/pseudo_transactions/set_fee.py
JST5000: Tried doing that just now, that causes problems when trying to use `from_xrpl` / `from_dict` because it tries to look up the `SetFee` transaction, but just finds the type alias and can't find the function definition. I think the optional parameters are good enough here since I think `from_dict` might be involved in parsing even if we're just reading data from the ledger.

pdp2121 (approved):

PR #638, opened 2023-08-28
fix: update request models related to AMM

khancode:

Update request models related to AMM:
- Add `amm` to ledger_entry and account_objects
- Add `amm_account` to `amm_info` request and set asset pair params to be optional

### Type of Change

## Test Plan

All tests pass.


mvadari (approved):
Please edit the changelog (since IIRC the original AMM PR has already been published)

JST5000 (approved):

xrpl/models/requests/amm_info.py
```diff
@@ -3,9 +3,10 @@
 
 from dataclasses import dataclass, field
 
+from pyparsing import Optional
```
ckniffen, line 6: The tests look to be failing due to using `pyparsing` and it not being added to the poetry config.  I think `typing` should be used instead.

PR #642, opened 2023-09-11
docs: Update ReadTheDocs (rtd) configuration

JST5000:

ReadTheDocs has some breaking changes that need to be updated before September 15th.

### Context of Change

- Adds build > os (a new required field as of October 25th)
https://blog.readthedocs.com/use-build-os-config/

Also fixes a misnamed 

### Type of Change

- [X] Documentation Updates

- [X] No, this change does not impact library users

## Test Plan

Builds locally and by inspection follows the new rules.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,7 +9,6 @@ build:
     python: "3.11"
 
 python:
-  version: 3.11
   install:
     - method: pip
       path: .
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,9 +1,19 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
 version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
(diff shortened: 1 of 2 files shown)


ckniffen:
Will this fix the theme?

JST5000:
> Will this fix the theme?

No, I don't think so, I'll do that in a separate PR. I think that has to do with the `conf.py` setup.

.readthedocs.yaml
```diff
@@ -1,9 +1,19 @@
...
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
 python:
-  version: 3.7
+  version: 3.11
```
mvadari, line 12: This should be removed - it's been deprecated. The version is only needed in `build.tools.python`.
https://docs.readthedocs.io/en/stable/config-file/v2.html#python-version

JST5000: Good catch :) 

I mistakenly thought the version was needed for the install step to function.

jonathanlei (approved):

mvadari (approved):

PR #643, opened 2023-09-11
docs: Update theme to ReadTheDocs always (fixes regression)

JST5000:

Since 2.2.0, the xrpl-py docs have been loading with a new theme. This brings it back to the previous ReadTheDocs theme.

### Context of Change

It seems that the default theme changed without warning (as far as we saw):

Autogenerated theme code from before 2.2.0:
https://readthedocs.org/projects/xrpl-py/builds/21212455/
image

Autogenerated theme code from versions after/including 2.2.0:
https://readthedocs.org/projects/xrpl-py/builds/21889906/
image

In general you can check builds and autogenerated theming by going to this build page: https://readthedocs.org/projects/xrpl-py/builds/

1. Click an individual build 
2. Click the "cat docs/conf.py" 
3. Scroll down past the "auto-generated below here" comment
4. Look for where the theme is defined

### Type of Change

- [X] Documentation Updates

- [X] No, this change does not impact library users

## Test Plan

Inspection and the docs build locally with the RTD theme ✅ 

~~The true efficacy will have to be determined after merging since this is a bug that only shows up in the production docs page, not in the local builds.~~

How it looks after the changes on ReadTheDocs:
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,6 +157,14 @@ cd docs/_build/html/
 open index.html
 ```
 
+You can view docs builds for xrpl-py versions on the ReadTheDocs website here: https://readthedocs.org/projects/xrpl-py/builds/
+
+In order to test how a change in docs configuration looks like on ReadTheDocs before merging:
+1. Publish a branch with your docs configuration changes
+2. Active and hide the branch by scrolling down on this page: https://readthedocs.org/projects/xrpl-py/versions/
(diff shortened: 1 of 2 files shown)


mvadari (approved):

JST5000:
Thanks @mvadari for showing me how to activate a branch on ReadTheDocs :) 

Added a screenshot for how it looks there, and updated the CONTRIBUTING.md with steps so other people can do that when modifying that config in the future.

ckniffen (approved):

justinr1234 (approved):

PR #645, opened 2023-09-21
test: add AMM integration tests

khancode:

Adds AMM integration tests.

### Context of Change

Adds AMM integration tests.

Transactions:
- `AMMCreate`
- `AMMDeposit`
- `AMMWithdraw`
- `AMMVote`
- `AMMBid`

Requests:
- `amm_info`

`AMMDelete` is omitted since it requires for at least 512 trustlines to be created which seems overkill for an integ test. There's already a unit test for the model so that should be sufficient since it only requires 2 params (asset pair).

### Type of Change

- Not applicable for integ tests.

## Test Plan

Run this against the CI once the amendment is added to the Docker container.

## Future Tasks

- [x] Add AMM amendment to Docker container.


tests/integration/it_utils.py
```diff
@@ -302,3 +308,143 @@ def _get_non_decorator_code(function):
...
+    client: SyncClient = JSON_RPC_CLIENT,
+) -> Dict[str, Any]:
+    issuer_wallet = Wallet.create()
+    fund_wallet(issuer_wallet)
+    lp_wallet = Wallet.create()
+    fund_wallet(lp_wallet)
+    currency_code = "USD"
+
+    # test prerequisites - create trustline and send funds
+    sign_and_reliable_submission(
```
mvadari, line 323: Doesn't `submit_and_wait` take a wallet param for this same process?

mvadari: Also, most of these don't seem to be taking a client param...

tests/integration/it_utils.py
```diff
@@ -302,3 +308,143 @@ def _get_non_decorator_code(function):
         if "def " in code_lines[line]:
             return code_lines[line:]
         line += 1
+
+
+def setup_amm_pool(
```
mvadari, line 313: This shouldn't be in `it_utils` if it's only used in `reusable_values` - it should just be there (and you only need one version of this method that way).

xrpl/models/transactions/amm_bid.py
```diff
@@ -41,14 +41,14 @@ class AMMBid(Transaction):
...
-    bid_min: Optional[Amount] = None
+    bid_min: Optional[IssuedCurrencyAmount] = None
     """
     Pay at least this amount for the slot.
     Setting this value higher makes it harder for others to outbid you.
     If omitted, pay the minimum necessary to win the bid.
     """
 
-    bid_max: Optional[Amount] = None
+    bid_max: Optional[IssuedCurrencyAmount] = None
```
mvadari, line 51: Are XRP amounts not valid for these?

This fix should be added to the changelog.

tests/integration/it_utils.py
```diff
@@ -302,3 +308,143 @@ def _get_non_decorator_code(function):
...
+                currency=currency_code,
+                value="250",
+            ),
+            trading_fee=12,
+        ),
+        lp_wallet,
+        client,
+    )
+
+    return {
```
mvadari, line 442: Why not return an `AMM` ledger object, to match the return shape of all the other reusable values?

tests/integration/transactions/test_amm_bid.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+
+        self.assertGreater(
+            float(amm_info.result["amm"]["auction_slot"]["price"]["value"]),
+            float(pre_amm_info.result["amm"]["auction_slot"]["price"]["value"]),
+        )
+        self.assertLess(
+            float(amm_info.result["amm"]["lp_token"]["value"]),
+            float(pre_amm_info.result["amm"]["lp_token"]["value"]),
+        )
```
mvadari, line 69: Why is this not asserting exact values?

tests/integration/transactions/test_amm_create.py
```diff
@@ -0,0 +1,30 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+from tests.integration.it_utils import test_async_and_sync
+from tests.integration.reusable_values import AMM
+from xrpl.models.requests.amm_info import AMMInfo
+
+
+class TestAMMCreate(IntegrationTestCase):
+    @test_async_and_sync(globals())
+    async def test_basic_functionality(self, client):
+        # issuer_wallet = AMM["issuer_wallet"]
```
mvadari, line 10: this looks like it should be removed

mvadari (commented):
Looks like a test for `amm_info` is missing.

xrpl/models/transactions/amm_bid.py
khancode: Right, only LPToken amounts are accepted. I just merged in a fix for this - #649

tests/integration/transactions/test_amm_bid.py
khancode: 2 reasons:
- The math to compute/predict the result values is complex and their result computations are different in JavaScript vs C++.
- One AMM pool is being shared among all tests so the values are constantly changing

Thus given these conditions, rather than asserting exact values, I think it's sufficient to just check whether tokens were added/removed after each transaction. Would be ideal if there was a rippled API to give the result exact values but there isn't.

mvadari: Shouldn't the difference results still be determined in a test, since the numbers can all be hardcoded?

tests/integration/it_utils.py
mvadari: `it_utils` is only needed for methods that are being used in the actual tests.

tests/integration/it_utils.py
khancode: I decided to remove `AMM` so it doesn't get mistaken as AMM ledger object. Now, I just returned 3 variables that are needed for tests: `AMM_ASSET`, `AMM_ASSET2`, `AMM_ISSUER_WALLET`.

tests/integration/transactions/test_amm_bid.py
khancode: Looks like they can't be hardcoded to a single value since each test runs twice (sync/async). This results in differing values:
```
AssertionError: 3311.0 != 1311
```
```
AssertionError: 4311.0 != 1311
```

mvadari: Is the difference not constant? Seems like it is in this example.

khancode: It's not constant, the left hand value changes from `3311` to `4311`

mvadari: Yes, I mean the amount added/subtracted (from the value before the transaction was sent)

khancode: Ah I see what you mean :) Yes, looks like it

ckniffen (approved):

jonathanlei (approved):
LGTM

PR #417, opened 2022-07-11
feat: add support for XLS-38d

mvadari:
## High Level Overview of Change

This PR adds support for the latest version of the sidechain design. It also removes the helper methods that were used for the previous version, since those are no longer needed.

There are 8 new transaction types:
* `XChainCreateBridge` (creates a new cross-chain bridge)
* `XChainModifyBridge` (modifies the properties of a cross-chain bridge)
* `XChainCreateClaimID` (creates a new cross-chain claim ID)
* `XChainCommit` (locks/burns the funds on the source chain)
* `XChainAccountCreateCommit` (creates an account on the destination chain)
* `XChainClaim` (claims the transfer funds on the destination chain)
* `XChainAddClaimAttestation` (submits a claim attestation from a witness server)
* `XChainAddAccountCreateAttestation` (submits an account creation attestation from a witness server)

### Context of Change

sidechain prod release

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Test Plan

Tests were added for all the new features. CI passes. I'd ideally like https://github.com/XRPLF/xrpl-py/pull/571 to be included in this PR before it's merged, but if it has to wait on the Docker container and is in a separate commit, that's fine.

PR #646, opened 2023-09-21
fix: export `get_nftoken_id` and `parse_nftoken_id`, better error handling

mvadari:

This PR:
* Exports `get_nftoken_id` and `parse_nftoken_id` at the `xrpl.utils` level
* Adds better error handling to `get_nftoken_id` if a metadata for an unrelated transaction is accidentally passed in

### Context of Change

Noticed while working on #417 

### Type of Change

- [x] Yes

## Test Plan

Added a test.


xrpl/utils/get_nftoken_id.py
```diff
@@ -97,6 +97,9 @@ def has_nftoken_page(node: Node) -> bool:
 
     affected_nodes = [node for node in meta["AffectedNodes"] if has_nftoken_page(node)]
 
+    if len(affected_nodes) == 0:
```
ckniffen, line 100: Is it typical to throw a exception and not just return None.  This would be a normal occurance in cases where there is a tec code.

mvadari: I think it'd be more difficult to identify a bug if it returns `None` instead of erroring.

khancode (approved):

ckniffen (approved):

xrpl/utils/get_nftoken_id.py
khancode: I agree with Mayukha's approach of throwing an error because it indicates invalid user input; this justifies throwing an error. We should probably make it consistent with JS to throw an error too.

khancode: However, it's also valid to have a failed transaction and not throw an error. So now I'm leaning more towards not throwing an error and keeping things consistent 😄

PR #652, opened 2023-09-29
feat: add support for XLS-40 + add script to auto-generate models from rippled code

mvadari:

This PR adds support for XLS-40 and adds a script to automatically generate transaction models from rippled source code.

### Context of Change

https://github.com/XRPLF/XRPL-Standards/pull/136
https://github.com/XRPLF/rippled/pull/4636

### Type of Change

- [x] Yes

## Test Plan

Added model tests (integration tests will be added later). CI passes.


tools/generate_tx_models.py
```diff
@@ -0,0 +1,189 @@
+"""Script to generate transaction models from rippled source code."""
```
justinr1234, line 1: @mvadari Can you write a test for this that includes a sample file that is parsed and sample output?

mvadari: I thought about that but it's pretty difficult to do, since you're editing source code.

justinr1234: @mvadari I don't follow. I think you could write it using this: https://docs.python.org/3/library/ast.html

mvadari: The way this script works is that it basically edits the xrpl-py source code to inject the transaction models and details. Writing a test with AST would be more difficult than it's worth IMO.

justinr1234 (approved):

tests/integration/transactions/test_did_delete.py
```diff
@@ -0,0 +1,31 @@
...
+    test_async_and_sync,
+)
+from tests.integration.reusable_values import WALLET
+from xrpl.models.response import ResponseStatus
+from xrpl.models.transactions import DIDDelete, DIDSet
+
+_VALID_FIELD = "1234567890abcdefABCDEF"
+
+
+class TestDIDSet(IntegrationTestCase):
```
JST5000, line 13: This test name is confusing since this should be about DIDDelete
```suggestion
class TestDIDDelete(IntegrationTestCase):
```

tests/integration/transactions/test_did_delete.py
```diff
@@ -0,0 +1,31 @@
...
+        # Create DID to delete
+        setup_tx = DIDSet(
+            account=WALLET.address,
+            did_document=_VALID_FIELD,
+            uri=_VALID_FIELD,
+            data=_VALID_FIELD,
+        )
+        response = await sign_and_reliable_submission_async(setup_tx, WALLET, client)
+        self.assertEqual(response.status, ResponseStatus.SUCCESS)
+
```
JST5000, line 25: Can you add checks to see that the DID object was created / deleted like in the js version? (While that kind of is testing rippled, I'm worried about the `ResponseStatus` not necessarily 100% being the same as the transaction being applied - I'm not sure if some error codes count as a successful response, e.g. ter codes which could be applied in a future ledger)

tests/unit/models/transactions/test_did_set.py
```diff
@@ -0,0 +1,54 @@
+from unittest import TestCase
+
+from xrpl.models.exceptions import XRPLModelException
+from xrpl.models.transactions import DIDSet
+
+_ACCOUNT = "r9LqNeG6qHxjeUocjvVki2XR35weJ9mZgQ"
+_VALID_FIELD = "1234567890abcdefABCDEF"
+_TOO_LONG_FIELD = "A" * 257
+_BAD_HEX_FIELD = "random_non_hex_content"
+_BAD_HEX_TOO_LONG_FIELD = "r" * 257
```
JST5000, line 10: nit: r is an odd character to repeat 257 times for hex codes
```suggestion
_BAD_HEX_TOO_LONG_FIELD = "r" + "a" * 256 // Total 257 chars
```

JST5000: Oh wait, nevermind hahaha - I thought the "Bad" in bad hex was just referring to the length and got confused that the `r` was because of `r`-addresses (which aren't hex to begin with).

xrpl/models/transactions/did_set.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True)
+class DIDSet(Transaction):
+    """Represents a DIDSet transaction."""
+
+    did_document: Optional[str] = None
+    data: Optional[str] = None
+    uri: Optional[str] = None
```
JST5000, line 25: Does at least one of these need to be specified? (If so, it doesn't seem like this is enforced in `_get_errors`)

JST5000: Also I don't see a test for that if so

tests/integration/transactions/test_did_delete.py
mvadari: Oops, copy-paste typo

tests/unit/models/transactions/test_did_set.py
mvadari: I'll change it to a different non-hex character.

xrpl/models/transactions/did_set.py
mvadari: Good catch! 🙂

tests/integration/transactions/test_did_delete.py
mvadari: Done

tests/unit/models/transactions/test_did_delete.py
```diff
@@ -0,0 +1,13 @@
...
+
+_ACCOUNT = "r9LqNeG6qHxjeUocjvVki2XR35weJ9mZgQ"
+
+
+class TestDIDDelete(TestCase):
+    def test_valid(self):
+        tx = DIDDelete(
+            account=_ACCOUNT,
+        )
+        self.assertTrue(tx.is_valid())
```
JST5000, line 13: Can you add a test where tx.is_valid() throws / fails?

JST5000 (approved):
LGTM :)

PR #672, opened 2024-01-11
add CTID field in the tx command request

ckeshava:

This PR incorporates the CTID update (XLS-37d) in the Python wrapper library.

### Context of Change

Based on a cursory read of the CTID PR (https://github.com/XRPLF/rippled/pull/4418/files), I believe the `tx` request accepts a new parameter titled "ctid". This has been incorporated into the Python Request wrapper class.

I couldn't identify any other commands that might have been impacted by the CTID feature.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

This PR adds a unit test file for the `tx` request. Three combinations of the `ctid` and `transaction` inputs have been verified.


ckniffen (approved):

xrpl/models/requests/tx.py
```diff
@@ -17,18 +17,28 @@
...
-    This field is required.
-
-    :meta hide-value:
-    """
-
+    transaction: Optional[str] = None
     binary: bool = False
     min_ledger: Optional[int] = None
     max_ledger: Optional[int] = None
+    ctid: Optional[str] = None
```
JST5000, line 30: While you're here, I'd recommend putting in the docstrings from xrpl.org as in-line comments like the old """ This field is required. """ - https://xrpl.org/tx.html
(Note that docstrings in python with our docs generation tool `Sphinx` happen AFTER the line you're trying to document as far as I understand)

I'd also include in those docstrings for `transaction` and `ctid` that either it or the other is required. 

Lastly, I'd re-order the parameters here so they match the order on xrpl.org, but that's a nit :)

xrpl/models/requests/tx.py
```diff
@@ -17,18 +17,28 @@
...
+            errors[
+                "Tx"
+            ] = "Must have only one of `ctid` or `transaction`, but not both."
+        return errors
+
+    def _has_only_one_input(self: Tx) -> bool:
+        present_items = [
+            item for item in [self.transaction, self.ctid] if item is not None
+        ]
+        return len(present_items) == 1
```
JST5000, line 44: Instead of list comprehension here, you can use `filter` to remove items which don't pass a test. If you're just trying to remove falsey values like `None` `bool` is a good enough function. (https://www.geeksforgeeks.org/remove-falsy-values-from-a-list-in-python/)

Explanation of truthy / falsey values in python: [truthy](https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/)

And last note, for readability I'd just recommend pulling the array `[self.transaction, self.ctid]` into a separate variable as there's already a lot going on in the one-liner, and that'll allow you to label why those two variables are being grouped. Ex. `unique_ids = [self.transaction, self.ctid]`

ckeshava: okay, fixed in [ce890c0](https://github.com/XRPLF/xrpl-py/pull/672/commits/ce890c04ceb5fc35ef832fc1c8439e712c4cd3e0)

xrpl/models/requests/tx.py
ckeshava: okay, updated in [ce890c0](https://github.com/XRPLF/xrpl-py/pull/672/commits/ce890c04ceb5fc35ef832fc1c8439e712c4cd3e0)

jonathanlei (approved):

ckeshava:
can I merge this PR? I have addressed the residual comments

PR #674, opened 2024-01-16
update the ISO currency code regular expression

ckeshava:

This PR updates the ISO Currency Code regex to better match the documentation.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes
This change enforces the standards communicated in the docs. Hence, there is not a breaking change.
## Test Plan

The PR adds a few unit tests to validate certain currency codes as inputs.


ckniffen (approved):

ckeshava:
I'm not sure why the snippet tests are failing. I tested the `send_escrow.py` file on my local computer 12 times, it succeeded every time. I'll take a look again tomorrow.

JST5000 (approved):
One change, for the HISTORY.md file we update that for minor changes as well (almost all our changes are not breaking changes). 

In this case I would add a note in the HISTORY.md file under a `# Fixed` header like:

```
# Fixed
- Currency codes with special characters not being allowed by IssuedCurrency objects.
```

mvadari (commented):
Was this tested by hand to ensure that rippled supports it as well?

ckeshava:
> # Fixed
> - Currency codes with special characters not being allowed by IssuedCurrency objects.

ok 👍 
@JST5000 Are you referring to the changelog file? the javascript codebase has a History.md file, the python codebase has the changelog file

ckeshava:
> Was this tested by hand to ensure that rippled supports it as well?

I haven't looked into the rippled code, I used the xrpl docs to verify this regex. Let me look into rippled as well

ckeshava:
@mvadari I couldn't figure out how to do an end-to-end test on a rippled instance. But I verified these inputs as unit test cases against the rippled codebase and it works correctly. Here's the branch: https://github.com/ckeshava/rippled/tree/currCodeTests

Furthermore, here is the set of allowed characters in rippled: https://github.com/XRPLF/rippled/blob/87ee7868eacd4969b92981b4c93bfe63bd2af0be/src/ripple/protocol/impl/UintTypes.cpp#L35

JST5000:
> > # Fixed
> > - Currency codes with special characters not being allowed by IssuedCurrency objects.
> 
> ok 👍 
> @JST5000 Are you referring to the changelog file? the javascript codebase has a History.md file, the python codebase has the changelog file

Ah yep! I often get the name mixed up hahaha

mvadari:
> @mvadari I couldn't figure out how to do an end-to-end test on a rippled instance. But I verified these inputs as unit test cases against the rippled codebase and it works correctly. Here's the branch: https://github.com/ckeshava/rippled/tree/currCodeTests
> 
> Furthermore, here is the set of allowed characters in rippled: https://github.com/XRPLF/rippled/blob/87ee7868eacd4969b92981b4c93bfe63bd2af0be/src/ripple/protocol/impl/UintTypes.cpp#L35

You should test it via writing an integration test or writing a quick script to check.

ckeshava:
@mvadari  I have written integration tests to verify the new behavior.

mvadari (approved):

PR #663, opened 2023-10-27
Add Ledger Objects

LimpidCrypto:
## High Level Overview of Change

This PR adds (some) Ledger Object models:
- AccountRoot
- Amendment
- AMM
- Bridge
- Check
- DepositPreauth
- DID
- DirectoryNode
- Escrow
- FeeSettings
- LedgerHashes
- LedgerObject
- NegativeUNL
- NFTokenOffer
- NFTokenPage
- Offer
- PayChannel
- RippleState
- SignerList
- Ticket
- XChainOwnedClaimID
- XChainOwnedCreateAccountClaimID

You might want to look over the fields if they are still correct as it's been a while since I wrote them.

**TODO**
- [x] move `MD...` models to metadata PR (see #664)
### Context of Change

@JST5000 and I had a great chat this week. He told me one thing you might prioritize as a coming feature would be response models. I remembered I once had a PR adding Ledger Object and Metadata models and was told to split it up into two PRs, but somehow I forgot about it. This PR adds the Ledger objects. I could imagine there are some objects missing or are a bit deprecated but I still wanted to leave the progress made here for you to use.

### Type of Change

- [x] New feature (non-breaking change which adds functionality)

### Did you update CHANGELOG.md?

- [ ] TODO: Yes

## Test Plan

I wrote tests trying to de-/serialize them from/to json.

PR #677, opened 2024-01-31
refactor: make transaction flag mapping more maintainable

mvadari:

This PR makes transaction flag mapping more maintainable and fixes the tests so they actually run properly. The `XChainModifyBridge` validations were also fixed, due to catching a bug while writing tests for this PR.

This process also caught a bug with the `NFTokenCreateOffer` flag names.

### Context of Change

When reviewing #663 I looked at how transaction flags were processed, which made me realize that the `XChainModifyBridge` flags were missed. Then I wanted to make it more maintainable, to avoid this problem in the future (I didn't know this was a place I needed to update).

### Type of Change

- [x] Yes

## Test Plan

Existing tests pass. Added a test for `XChainModifyBridge` flags.


tests/unit/models/transactions/test_better_transaction_flags.py
```diff
@@ -49,20 +50,18 @@ def test_account_set_flags(self):
     def test_nftoken_create_offer_flags(self):
         actual = models.NFTokenCreateOffer(
             account=ACCOUNT,
-            nftoken_id="\
-                000100001E962F495F07A990F4ED55ACCFEEF365DBAA76B6A048C0A200000007",
+            nftoken_id=NFTOKEN_ID,
             amount="1000000",
             flags=models.NFTokenCreateOfferFlagInterface(
-                tf_sell_token=True,
+                TF_SELL_NFTOKEN=True,
```
mvadari, line 56: Noting this as originally having the wrong name

mvadari, line 56: Noting this as originally having the wrong name

khancode (approved):
LGTM

xrpl/models/flags.py
```diff
@@ -1,110 +1,74 @@
...
+    # Key is transaction type name, value is mapping of flag name to int value
+    all_tx_flags: Dict[str, Dict[str, int]] = {
+        # The `:-4` here removes the `Flag` at the end of the class type to just get
+        # the transaction type name
+        **{f[:-4]: _flag_enum_to_dict(getattr(transactions, f)) for f in flag_enums},
+        **{
+            f[:-4]: _flag_enum_to_dict(getattr(transactions.pseudo_transactions, f))
+            for f in pseudo_tx_flag_enums
+        },
+    }
```
khancode, line 57:
```suggestion
    tx_flag_enums = [
        f for f in transactions.__all__ if f.endswith("Flag") and "Asf" not in f
    ]
    pseudo_tx_flag_enums = [
        f for f in transactions.pseudo_transactions.__all__ if f.endswith("Flag")
    ]

    # Key is transaction type name, value is mapping of flag name to int value
    all_tx_flags: Dict[str, Dict[str, int]] = {
        # The `:-4` here removes the `Flag` at the end of the class type to just get
        # the transaction type name
        **{f[:-4]: _flag_enum_to_dict(getattr(transactions, f)) for f in tx_flag_enums},
        **{
            f[:-4]: _flag_enum_to_dict(getattr(transactions.pseudo_transactions, f))
            for f in pseudo_tx_flag_enums
        },
    }
```
Renaming `flag_enums` to `tx_flag_enums` makes it specific.

JST5000 (approved):
LGTM

PR #681, opened 2024-02-01
feat(contributing): update model script to process ledger objects

mvadari:

This PR updates the `generate_tx_models.py` script to also add ledger object models to xrpl-py. (and renames it to `generate_models.py`).

### Context of Change

Helping review #663. This should probably only be merged after that PR is merged.

### Type of Change

- [x] New feature for contributors

## Test Plan

The script runs properly and generates good models for all the existing ledger objects.


ckeshava:
@mvadari  The linked PR (titled: Add Ledger Objects) is no longer maintained. What do you propose to do with this PR's changes?

PR #684, opened 2024-02-08
incorporate DeliverMax in Payment transactions

ckeshava:
Add support for `DeliverMax` field in the Payment transactions.  The following links will give you more context on the `DeliverMax` field:
https://github.com/XRPLF/rippled/pull/4733, https://github.com/XRPLF/rippled/issues/3484

### Context of Change

`Payment` transactions accept a new field `DeliverMax`, which serves as an alias to the `Amount` field. This is intended to prevent misinterpretation of the response of `Payment` transactions and the consequent loss of funds.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Additional unit tests have been added to validate the usage of `Payment` transactions.


ckeshava:
@mvadari  I felt it is more appropriate to make the `deliver max` changes in the `from_dict` overload. The `autofill` and other utility functions work with a `Transaction` object. Since `DeliverMax` is not a part of protocol, it is not a recognized field in the Payment request.

mvadari:
> @mvadari I felt it is more appropriate to make the `deliver max` changes in the `from_dict` overload. The `autofill` and other utility functions work with a `Transaction` object. Since `DeliverMax` is not a part of protocol, it is not a recognized field in the Payment request.

This won't work in a lot of cases (including internal uses of `from_dict`). A lot of people use `Transaction.from_dict` instead of the specific transaction type.

ckeshava:
@mvadari I agree that there are limitations in my current design. 

If I were to make this change in `autofill` method, how can I go about it? `autofill` takes in a `Transaction` object and  I cannot address any field titled `deliver_max` because its not a part of the `Payment` transaction model here: https://github.com/XRPLF/xrpl-py/blob/2e1435c37c3f2e1a3b41ac21f02f1a692258cee0/xrpl/models/transactions/payment.py#L62

I'd like to test inputs such as 
```
transaction_dict = {
            "account": _ACCOUNT,
            "fee": _FEE,
            "sequence": _SEQUENCE,
            "deliver_max": _ISSUED_CURRENCY_AMOUNT,
            "send_max": _XRP_AMOUNT,
            "destination": _DESTINATION,
        }
# convert the above JSON into a Payment object, without throwing any errors.
```

We are passing along `Transaction` object to methods like `autofill`, `autofill_and_sign`, `sign_and_submit` etc. Where can I get the opportunity to modify the JSON version of a Transaction?

ckeshava:
I have moved the changes to `Transaction.from_dict` from the `Payment` class. `from_xrpl` invokes `from_dict` in the course of its actions. I felt the former is most appropriate to accomodate these changes.

mvadari:
> I felt the former is most appropriate to accomodate these changes.

I disagree, because it's only ever going to be necessary when pulling from XRPL data, not from any other dictionary.

xrpl/models/base_model.py
```diff
@@ -73,6 +73,34 @@ def _value_to_json(value: XRPL_VALUE_TYPE) -> XRPL_VALUE_TYPE:
...
+def process_json_binary_codec_input(
+    value: Union[str, Dict[str, Any]]
+) -> Dict[str, Any]:
+    """
+    Creates a dictionary object based on a JSON-like dictionary of keys in the JSON
+    format used by the binary codec, or an actual JSON string representing the same
+    data.
+
+    This utility function is used in BaseModel.from_xrpl and Transaction.from_xrpl for
+    handling JSON and JSON-string like inputs.
```
mvadari, line 85:
```suggestion
```

xrpl/models/base_model.py
```diff
@@ -73,6 +73,34 @@ def _value_to_json(value: XRPL_VALUE_TYPE) -> XRPL_VALUE_TYPE:
     return value
 
 
+def process_json_binary_codec_input(
```
mvadari, line 76:
```suggestion
def process_xrpl_json(
```
The binary codec is irrelevant to the format of the JSON.

xrpl/models/base_model.py
```diff
@@ -73,6 +73,34 @@ def _value_to_json(value: XRPL_VALUE_TYPE) -> XRPL_VALUE_TYPE:
...
+    value: Union[str, Dict[str, Any]]
+) -> Dict[str, Any]:
+    """
+    Creates a dictionary object based on a JSON-like dictionary of keys in the JSON
+    format used by the binary codec, or an actual JSON string representing the same
+    data.
```
mvadari, line 82:
```suggestion
    Creates a dictionary object based on a JSON or dictionary in the standard XRPL format.
```

xrpl/models/transactions/transaction.py
```diff
@@ -468,3 +472,42 @@ def from_blob(tx_blob: str) -> Transaction:
             The formatted transaction.
         """
         return Transaction.from_xrpl(decode(tx_blob))
+
+    @classmethod
+    def from_xrpl(cls: Type[T], inp_params: Union[str, Dict[str, Any]]) -> T:
```
mvadari, line 477:
```suggestion
    def from_xrpl(cls: Type[T], value: Union[str, Dict[str, Any]]) -> T:
```
This should use the same formatting as the function it's overriding.

xrpl/models/transactions/transaction.py
```diff
@@ -468,3 +472,42 @@ def from_blob(tx_blob: str) -> Transaction:
...
+        """
+        Creates a Transaction object based on a JSON-like dictionary of keys in the JSON
+        format used by the binary codec, or an actual JSON string representing the same
+        data.
+
+        In Payment transactions, the deliver_max field (alias of the amount field) is
+        handled in this function
```
mvadari, line 484:
```suggestion
        In Payment transactions, the DeliverMax field is renamed to the Amount field.
```

xrpl/models/base_model.py
ckeshava: this is much nicer :)

xrpl/models/transactions/transaction.py
ckeshava: this is a nice suggestion

ckeshava:
I've addressed your comments @mvadari  👍

xrpl/models/transactions/transaction.py
```diff
@@ -468,3 +468,43 @@ def from_blob(tx_blob: str) -> Transaction:
...
+        # handle the deliver_max alias in Payment transactions
+        if (
+            "transaction_type" in processed_value
+            and processed_value["transaction_type"] == "Payment"
+        ):
+            if "deliver_max" in processed_value:
+                if "amount" in processed_value:
+                    if processed_value["amount"] != processed_value["deliver_max"]:
+                        raise XRPLModelException(
+                            "Error: amount and deliver_max fields are not identical"
```
mvadari, line 501:
```suggestion
                            "Error: Amount and DeliverMax fields are not identical"
```

khancode (approved):

ckeshava:
@mvadari  does this design look okay? Can I do the same for JS too?

xrpl/models/base_model.py
```diff
@@ -73,6 +73,28 @@ def _value_to_json(value: XRPL_VALUE_TYPE) -> XRPL_VALUE_TYPE:
     return value
 
 
+def process_xrpl_json(value: Union[str, Dict[str, Any]]) -> Dict[str, Any]:
```
justinr1234, line 76: Seems like this function can return exceptions which we should note and/or handle

xrpl/models/transactions/transaction.py
```diff
@@ -468,3 +468,43 @@ def from_blob(tx_blob: str) -> Transaction:
...
+            value: The dictionary or JSON string to be instantiated.
+
+        Returns:
+            A Transaction object instantiated from the input.
+
+        Raises:
+            XRPLModelException: If Payment transactions have different values for
+                                amount and deliver_max fields
+        """
+        processed_value = process_xrpl_json(value)
```
justinr1234, line 490: This may not return successfully so we should likely note that in the docs

xrpl/models/transactions/transaction.py
```diff
@@ -468,3 +468,43 @@ def from_blob(tx_blob: str) -> Transaction:
...
+
+        # handle the deliver_max alias in Payment transactions
+        if (
+            "transaction_type" in processed_value
+            and processed_value["transaction_type"] == "Payment"
+        ):
+            if "deliver_max" in processed_value:
```
justinr1234, line 497: Seems like this should be moved into the above if

xrpl/models/transactions/transaction.py
```diff
@@ -468,3 +468,43 @@ def from_blob(tx_blob: str) -> Transaction:
...
+                if "amount" in processed_value:
+                    if processed_value["amount"] != processed_value["deliver_max"]:
```
justinr1234, line 499: This should be combined into the 498 if

xrpl/models/transactions/transaction.py
```diff
@@ -468,3 +468,43 @@ def from_blob(tx_blob: str) -> Transaction:
...
+                        raise XRPLModelException(
+                            "Error: Amount and DeliverMax fields are not identical"
```
justinr1234, line 501: Is there maybe a link we can give to explain why they should be equal in the error message?

justinr1234 (approved):
nit changes only

xrpl/models/transactions/transaction.py
ckeshava: Hmm, I can point to the issue discussion here: https://github.com/XRPLF/rippled/issues/3484

Here is an implementation commit to the rippled codebase: https://github.com/XRPLF/rippled/commit/397268394b80574419fddafe77aa75ea11ad9cfe 

I'm not aware of any other documents summarizing this issue.

`DeliverMax` is an alias for the `Amount` field. So if they contain different values, it introduces semantic ambiguity. Which one should we use? Both of them refer to the same logical quantity -- the value of transferred money

xrpl/models/base_model.py
ckeshava: I don't know which exceptions might be thrown by this code. `_key_to_json` and `_value_to_json` do not throw any exceptions themselves

ckeshava: Do you suspect that `json.load` could throw a `json.JSONDecodeError` exception?

xrpl/models/transactions/transaction.py
khancode: If both fields are passed in, then they have to be equal. We can update the error message for clarity: `"Error: Amount and DeliverMax fields must be equal if both are provided"`

xrpl/models/transactions/transaction.py
```diff
@@ -468,3 +468,44 @@ def from_blob(tx_blob: str) -> Transaction:
...
+        if (
+            "transaction_type" in processed_value
+            and processed_value["transaction_type"] == "Payment"
+        ) and "deliver_max" in processed_value:
+            if (
+                "amount" in processed_value
+                and processed_value["amount"] != processed_value["deliver_max"]
+            ):
+                raise XRPLModelException(
+                    "Error: amount and deliver_max fields are not identical"
```
khancode, line 502:
```suggestion
                    "Error: amount and deliver_max fields must be equal if both are provided"
```

khancode (changes requested):

xrpl/models/base_model.py
mvadari: This function will never be directly used by end users.

xrpl/models/transactions/transaction.py
mvadari: Don't link rippled issues from the libraries, point to the xrpl.org documentation. It's not useful to do so in error messages, because that makes the error message really long, but it can be in the docstring.

xrpl/models/base_model.py
mvadari, line 76: I'd recommend making this a helper class method, instead of a separate function - it saves you some imports. Similar to `_get_only_init_args`.

xrpl/models/transactions/transaction.py
ckeshava: I was brainstorming ideas for resource pages for this issue, I won't include it in the error message 👍 

I couldn't find a statement that says "If present, `DeliverMax` field must equal `Amount`" in the Payments docs. https://xrpl.org/docs/references/protocol/transactions/types/payment#payment-fields 

I feel it's a better developer experience to provide an elaborate error message, instead of making them search through a website.

mvadari: `DeliverMax` is only an RPC field, rippled doesn't care about it. It's only ever returned. That documentation should probably elaborate on that a bit.

xrpl/models/base_model.py
ckeshava: okay, I have done this. But I dislike the fact that the `self/cls` argument is not useful inside the `process_xrpl_json` function. I'm only using it to hook into this function

xrpl/models/transactions/transaction.py
ckeshava: Yes, documentation will help. `DeliverMax` can also be used as an input in APIv2 rippled. 

Every instance of `DeliverMax`, at the protocol (or) binary-codec-level, is translated into `Amount`.

mvadari: No, it was not changed at the protocol level, just at the RPC level.

xrpl/models/base_model.py
mvadari: I think the tradeoff is worth it given that it makes it easier to use, but if someone strongly objects I'll withdraw

ckeshava: yeah, I think it's a reasonable tradeoff 👍

ckeshava:
@khancode  can you please take a look at this PR? does this design look alright?

@justinr1234  I need some clarification on your prior comments. Let me know if I need to make any more changes.

thnks

xrpl/models/base_model.py
```diff
@@ -213,6 +213,30 @@ def _from_dict_single_param(
             )
         raise XRPLModelException(error_message)
 
+    @classmethod
+    def process_xrpl_json(
```
khancode, line 217:
```suggestion
    def _process_xrpl_json(
```

Looks like this is a private method so we should prefix it with underscore `_`

khancode (approved):
LGTM, just need to change a method to be private.

xrpl/models/base_model.py
ckeshava: thanks, I've addressed it. I had to make changes in two other function call places, introduced a separate commit

ckeshava:
the snippets tests are failing due to an unfunded account (We believe it lacks "USD" tokens). It will need to be addressed in a separate PR

ckeshava:
If there are no objections, I'd like to merge this PR.

Thank you all for your feedback

Issue #503, opened 2023-01-25
`Wallet` should throw a clearer error message when given an invalid seed


JST5000:
```
issuer_wallet = Wallet(seed="")
```
Results in 
```
  File "/Users/jacksonmills/Documents/bounty_submissions/xrpl-dev-portal/content/_code-samples/non-fungible-token/py/authorize-minter.py", line 21, in <module>
    issuer_wallet = Wallet(seed=seed, sequence=0)
  File "/Users/jacksonmills/Library/Caches/pypoetry/virtualenvs/xrpl-validator-dashboard-mm7G7WMB-py3.9/lib/python3.9/site-packages/xrpl/wallet/main.py", line 41, in __init__
    pk, sk = derive_keypair(self.seed, algorithm=algorithm)
  File "/Users/jacksonmills/Library/Caches/pypoetry/virtualenvs/xrpl-validator-dashboard-mm7G7WMB-py3.9/lib/python3.9/site-packages/xrpl/core/keypairs/main.py", line 69, in derive_keypair
    decoded_seed, algorithm = addresscodec.decode_seed(seed, algorithm)
  File "/Users/jacksonmills/Library/Caches/pypoetry/virtualenvs/xrpl-validator-dashboard-mm7G7WMB-py3.9/lib/python3.9/site-packages/xrpl/core/addresscodec/codec.py", line 123, in decode_seed
    decoded_result = _decode(seed, bytes(prefix))
  File "/Users/jacksonmills/Library/Caches/pypoetry/virtualenvs/xrpl-validator-dashboard-mm7G7WMB-py3.9/lib/python3.9/site-packages/xrpl/core/addresscodec/codec.py", line 61, in _decode
    decoded = base58.b58decode_check(b58_string, alphabet=XRPL_ALPHABET)
  File "/Users/jacksonmills/Library/Caches/pypoetry/virtualenvs/xrpl-validator-dashboard-mm7G7WMB-py3.9/lib/python3.9/site-packages/base58/__init__.py", line 157, in b58decode_check
    raise ValueError("Invalid checksum")
ValueError: Invalid checksum
```

Ideally this would throw an error which says something like "Attempted to initialize a Wallet with an invalid seed." to help people debug since seeds can be a bit tricky to verify by look alone. Also, not everyone will know that base58 is used during encoding/decoding seeds.

PR #686, opened 2024-02-14
validate the contents of seed before creating a Wallet object

ckeshava:
Fixes https://github.com/XRPLF/xrpl-py/issues/503

It's ideal to validate a seed before creating a `Wallet` object. This is helpful in generating readable error messages for the users.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes
Invalid seed input to `Wallet` will throw a different error. The current code throws a `ValueError`, this is hard to understand for users.

## Test Plan
I have added a unit test that validates the Exception thrown upon empty seed inputs into the `Wallet` class constructor.

### DIFF
--- a/tests/unit/asyn/wallet/test_main.py
+++ b/tests/unit/asyn/wallet/test_main.py
@@ -1,6 +1,7 @@
 from unittest import TestCase
 
 from xrpl.constants import CryptoAlgorithm, XRPLException
+from xrpl.core.addresscodec.exceptions import XRPLAddressCodecException
 from xrpl.wallet.main import Wallet
 
 constants = {
@@ -138,6 +139,15 @@ class TestWalletMain(TestCase):
         )
 
         _test_wallet_values(self, wallet, "seed", "secp256k1")
+
+    def test_wallet_contructor_throws_on_empty_seed(self):
+        with self.assertRaises(XRPLAddressCodecException):
+            Wallet(
+                constants["regular_key_pair"]["secp256k1"]["public_key"],
+                constants["regular_key_pair"]["secp256k1"]["private_key"],
+                master_address=constants["regular_key_pair"]["master_address"],
+                seed="",
+            )
 
     def test_constructor_using_regular_key_pair(self):
         wallet = Wallet(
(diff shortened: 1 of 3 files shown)


ckeshava:
i don't understand why the `seed` is an `Optional[str]` in `Wallet`'s constructor here: https://github.com/XRPLF/xrpl-py/blob/a38f61d3c88b4a9a0807392ef4b0cb30aa688ce5/xrpl/wallet/main.py#L45

In what kind of situations can we get away with a non-existent seed? I saw a unit test that does not specify a `seed` for the `Wallet`, but I couldn't find any other examples. https://github.com/XRPLF/xrpl-py/blob/a38f61d3c88b4a9a0807392ef4b0cb30aa688ce5/tests/unit/asyn/wallet/test_main.py#L81

ckeshava:
I see. So a seed is used to generate a public/private pair of keys. But I can bypass that step if I already know the cryptographic key pair.

tests/unit/asyn/wallet/test_main.py
```diff
@@ -139,6 +140,15 @@ def test_constructor(self):
...
         _test_wallet_values(self, wallet, "seed", "secp256k1")
 
+    def test_wallet_contructor_throws_on_empty_seed(self):
+        with self.assertRaises(XRPLAddressCodecException):
+            Wallet(
+                constants["regular_key_pair"]["secp256k1"]["public_key"],
+                constants["regular_key_pair"]["secp256k1"]["private_key"],
+                master_address=constants["regular_key_pair"]["master_address"],
+                seed="",
+            )
```
JST5000, line 150: This test case can be simplified I believe by removing the master_address piece and using a clearly invalid seed (as opposed to one which could be mistaken for an "empty" seed - aka not passing one in)

```suggestion
    def test_wallet_contructor_throws_with_invalid_seed(self):
        with self.assertRaises(XRPLAddressCodecException):
            Wallet(
                constants["regular_key_pair"]["secp256k1"]["public_key"],
                constants["regular_key_pair"]["secp256k1"]["private_key"],
                seed="abc",
            )
```

JST5000: Or add another test case for the empty string case as opposed to the "regular" error case

JST5000 (approved):
Looks good other than a simplification in the test case :)

ckeshava:
@justinr1234 can you please review this PR at your convenience?

justinr1234 (approved):

PR #687, opened 2024-02-15
Custom timeout for requests

ckeshava:

Certain requests could take more time to return a response, than the default `10 seconds` duration. This change allows users to specify a custom timeout parameter to all requests.
Additionally, this PR sets up uniform interfaces for all `Client._request_impl` overridden functions.

### Context of Change

In the present code, all requests are given 10 seconds duration to return a response. Any longer latency results in a timeout. This is inconvenient for certain commands that are known to take up more time, for instance, `admin`-only commands like the `ledger` command.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

This is an existing integration test that invokes the `ledger` command at `tests/integration/reqs/test_ledger.py`. Which other command might take up more time?


xrpl/asyncio/clients/async_websocket_client.py
```diff
@@ -262,7 +262,9 @@ async def send(self: AsyncWebsocketClient, request: Request) -> None:
             raise XRPLWebsocketException("Websocket is not open")
         await self._do_send(request)
 
-    async def _request_impl(self: WebsocketBase, request: Request) -> Response:
+    async def _request_impl(
+        self: WebsocketBase, request: Request, timeout: Optional[float] = None
```
mvadari, line 266:
```suggestion
        self: WebsocketBase, request: Request, timeout: Optional[float] = None
```
This should be a keyword-only argument, to avoid accidental usage.
Ditto everywhere else.

mvadari (commented):
It doesn't look like this PR actually modifies the external-facing methods.

xrpl/asyncio/clients/client.py
```diff
@@ -4,9 +4,15 @@
 from abc import ABC, abstractmethod
 from typing import Optional
 
+from typing_extensions import Final
+
 from xrpl.models.requests.request import Request
 from xrpl.models.response import Response
 
+# The default request timeout duration. If you need more time, please pass the required
```
justinr1234, line 12: I would reword this to not be addressing someone.

`The default request timeout duration. Set in Client._request_impl to allow more time for longer running commands.`

xrpl/asyncio/clients/websocket_base.py
```diff
@@ -199,7 +199,9 @@ async def _do_pop_message(self: WebsocketBase) -> Dict[str, Any]:
         cast(_MESSAGES_TYPE, self._messages).task_done()
         return msg
 
-    async def _do_request_impl(self: WebsocketBase, request: Request) -> Response:
+    async def _do_request_impl(
+        self: WebsocketBase, request: Request, timeout: Optional[float]
```
justinr1234, line 203: Why is timeout a positional instead of named arg here?

justinr1234 (approved):

xrpl/asyncio/clients/client.py
```diff
@@ -27,14 +33,22 @@ def __init__(self: Client, url: str) -> None:
...
         async-first. Implement this in a given Client.
 
         Arguments:
             request: An object representing information about a rippled request.
+            timeout: The maximum tolerable delay on waiting for a response.
+                Note: Optional is used in the type in order to honor the existing
+                behavior in certain overridden functions. WebsocketBase.do_request_impl
+                waits indefinitely for the completion of a request, whereas
+                JsonRpcBase._request_impl waits for 10 seconds before timing out a
+                request.
```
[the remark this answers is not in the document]
mvadari: Also, this note shouldn't be in the docstring.

xrpl/asyncio/clients/websocket_base.py
ckeshava: I don't understand, how can I make it a named argument?

xrpl/asyncio/clients/client.py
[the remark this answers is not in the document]
ckeshava: addressed in [70dfb80](https://github.com/XRPLF/xrpl-py/pull/687/commits/70dfb80af3b82a90664da5e1b8e6ef60fe409912)

ledhed2222:
i just happened to see this. it looks good to me! great idea :)

PR #688, opened 2024-02-15
test: run tests on python 3.12

mvadari:

Title says it all.

### Context of Change

I was running something on Python 3.12 and wanted to make sure everything would work properly.

### Type of Change

- [ ] Yes

## Test Plan

CI passes.

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -6,7 +6,7 @@ env:
 
 on:
   push:
-    branches: [master]
+    branches: [main]
   pull_request:
   workflow_dispatch:
 
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -6,7 +6,7 @@ env:
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
   workflow_dispatch:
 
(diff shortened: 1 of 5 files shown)


.github/workflows/integration_test.yml
```diff
@@ -6,7 +6,7 @@ env:
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
```
ckeshava, line 9: shouldn't we use the `main` branch for integration testing?

mvadari: Good catch - the branch was renamed but CI wasn't updated.

.github/workflows/unit_test.yml
```diff
@@ -2,7 +2,7 @@ name: Unit test
 
 on:
   push:
-    branches: [master]
+    branches: [main]
   pull_request:
   workflow_dispatch:
```
ckeshava, line 8: the `lint-and-type-check` job is executed only on Python 3.7. It shouldn't cause any issue unless there is a breaking change in the linter in between python versions. Should we change it? What do you think?

ckeshava (approved):

.github/workflows/unit_test.yml
mvadari: That is on purpose, to ensure backwards compatibility.

mvadari: That change could also be done in a separate PR, since it's irrelevant to this PR.

jonathanlei (approved):

PR #704, opened 2024-05-14
refactor: use single Oracle class for Price Oracles

khancode:

Use single Oracle class for Price Oracles support.

### Context of Change

- Use single `Oracle` class for `get_aggregate_price` and `ledger_entry` to resolve duplicate name conflict.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

## Test Plan

Updated unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Construction of Wallet throws an "Invalid Seed" error, if the secret is not decode-able.
 - Rectify the incorrect usage of a transaction flag name: Update `TF_NO_DIRECT_RIPPLE` to `TF_NO_RIPPLE_DIRECT`
 - Add the missing `AMMDeposit` Flag `TF_TWO_ASSET_IF_EMPTY`
-- Minor refactor changes for Price Oracles support.
 
 ## [2.5.0] - 2023-11-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Construction of Wallet throws an "Invalid Seed" error, if the secret is not decode-able.
 - Rectify the incorrect usage of a transaction flag name: Update `TF_NO_DIRECT_RIPPLE` to `TF_NO_RIPPLE_DIRECT`
 - Add the missing `AMMDeposit` Flag `TF_TWO_ASSET_IF_EMPTY`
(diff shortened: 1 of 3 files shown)


xrpl/models/requests/get_aggregate_price.py
```diff
@@ -53,7 +51,7 @@ def _get_errors(self: GetAggregatePrice) -> Dict[str, str]:
 
 @require_kwargs_on_init
 @dataclass(frozen=True)
-class Oracle(TypedDict):
+class OracleEntry:
```
mvadari, line 54: Would this be confusing given that the other `Oracle` class is in `LedgerEntry`?

Naming is hard 😕

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Construction of Wallet throws an "Invalid Seed" error, if the secret is not decode-able.
 - Rectify the incorrect usage of a transaction flag name: Update `TF_NO_DIRECT_RIPPLE` to `TF_NO_RIPPLE_DIRECT`
 - Add the missing `AMMDeposit` Flag `TF_TWO_ASSET_IF_EMPTY`
+- Minor refactor changes for Price Oracles support.
```
mvadari, line 21: The changelog doesn't need to be updated here if it's in the same version that the price oracle PR is in.

xrpl/models/requests/get_aggregate_price.py
khancode: Yeah not sure 😅 We could rename it to `OracleParamEntry` or `OracleParam`?

mvadari: Looks like they're identical, so why not just have one object?

khancode: Not quite identical in types: one has `oracle_document_id` set to `Union[str, int]` and the other has `int`. But I guess using `Union[str, int]` would be sufficient.

mvadari: I'm guessing `Union[str, int]` is probably correct for the RPC too, tbh.

mvadari: https://github.com/XRPLF/rippled/issues/5014

khancode: Done in https://github.com/XRPLF/xrpl-py/pull/704/commits/56f1260af99e5101de24eb6c31d6c60f4caa4a57

mvadari (approved):

pdp2121 (approved):

PR #706, opened 2024-05-15
refactor: use `Self` instead of class name

mvadari:

This PR replaces the `self: ClassName` paradigm with `self: Self`, added in [Python 3.11](https://peps.python.org/pep-0673/) but supported in `typing_extensions` for all versions. This allows us to remove most of the awkward, unintuitive uses of `TypeVar`.

### Context of Change

It's cleaner and better for handling inheritance.

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/tests/unit/utils/test_get_nftoken_id.py
+++ b/tests/unit/utils/test_get_nftoken_id.py
@@ -1,4 +1,5 @@
 """Test the get_nftoken_id util."""
+
 from __future__ import annotations
 
 import json
@@ -18,22 +19,22 @@ with open(path_to_json + "offer_cancelled.json", "r") as infile:
 class TestGetNFTokenID(TestCase):
     """Test get_nftoken_id."""
 
-    def test_decoding_a_valid_nftoken_id(self: TestGetNFTokenID):
+    def test_decoding_a_valid_nftoken_id(self):
         result = get_nftoken_id(nftokenmint_response1["meta"])
         expected_nftoken_id = (
             "00081388DC1AB4937C899037B2FDFC3CB20F6F64E73120BB5F8AA66A00000228"
         )
         self.assertEqual(result, expected_nftoken_id)
 
-    def test_a_different_valid_nftokenmint_metadata(self: TestGetNFTokenID):
+    def test_a_different_valid_nftokenmint_metadata(self):
         result = get_nftoken_id(nftokenmint_response2["meta"])
         expected_nftoken_id = (
             "0008125CBE4B401B2F62ED35CC67362165AA813CCA06316FFA766254000003EE"
         )
         self.assertEqual(result, expected_nftoken_id)
 
-    def test_error_with_wrong_tx_metadata(self: TestGetNFTokenID) -> None:
+    def test_error_with_wrong_tx_metadata(self) -> None:
         self.assertIsNone(get_nftoken_id(wrong_fixture["meta"]))
 
-    def test_error_when_given_raw_instead_of_meta(self: TestGetNFTokenID) -> None:
+    def test_error_when_given_raw_instead_of_meta(self) -> None:
         self.assertRaises(TypeError, lambda: get_nftoken_id(nftokenmint_response1))
--- a/xrpl/asyncio/clients/async_websocket_client.py
+++ b/xrpl/asyncio/clients/async_websocket_client.py
@@ -1,9 +1,12 @@
 """A client for interacting with the rippled WebSocket API."""
+
 from __future__ import annotations
 
 from collections.abc import AsyncIterator
 from types import TracebackType
 from typing import Any, Dict, Type
+
+from typing_extensions import Self
 
 from xrpl.asyncio.clients.async_client import AsyncClient
 from xrpl.asyncio.clients.client import REQUEST_TIMEOUT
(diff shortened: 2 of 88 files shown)


khancode (approved):
That was a lot to cover but LGTM 😄

mvadari:
> That was a lot to cover but LGTM 😄

It's all just the same change over and over again 😅

xrpl/asyncio/clients/async_websocket_client.py
```diff
@@ -206,17 +209,17 @@ async def main():
...
             await self._do_open()
 
-    async def close(self: AsyncWebsocketClient) -> None:
+    async def close(self: Self) -> None:
         """Closes the connection."""
         if self.is_open():
             await self._do_close()
 
-    async def __aenter__(self: AsyncWebsocketClient) -> AsyncWebsocketClient:
+    async def __aenter__(self: Self) -> AsyncWebsocketClient:
```
ckeshava, line 222:
```suggestion
    async def __aenter__(self: Self) -> Self:
```
If this class is going to be inherited from in the future, it's beneficial to return the `Self`

xrpl/clients/websocket_client.py
```diff
@@ -135,7 +138,7 @@ def close(self: WebsocketClient) -> None:
         self._loop = None
         self._thread = None
 
-    def __enter__(self: WebsocketClient) -> WebsocketClient:
+    def __enter__(self: Self) -> WebsocketClient:
```
ckeshava, line 141:
```suggestion
    def __enter__(self: Self) -> Self:
```

xrpl/core/binarycodec/types/account_id.py
```diff
@@ -41,7 +42,7 @@ def __init__(self: AccountID, buffer: Optional[bytes] = None) -> None:
             super().__init__(bytes(self.LENGTH))
 
     @classmethod
-    def from_value(cls: Type[AccountID], value: str) -> AccountID:
+    def from_value(cls: Type[Self], value: str) -> Self:
```
ckeshava, line 45: what does `Type[Self]` do? 

I tried to use `cls: Self`, but MyPy complains:
```
The erased type of self "xrpl.core.binarycodec.types.account_id.AccountID" is not a supertype of its class "Type[xrpl.core.binarycodec.types.account_id.AccountID]"Mypy[misc](https://mypy.readthedocs.io/en/latest/_refs.html#code-misc)
```

I'm trying to adapt to the last example in this sub-section: https://typing.readthedocs.io/en/latest/spec/generics.html#use-in-classmethod-signatures

xrpl/core/binarycodec/types/currency.py
```diff
@@ -113,12 +114,12 @@ def from_value(cls: Type[Currency], value: str) -> Currency:
             )
 
         if _is_iso_code(value):
-            return Currency(_iso_to_bytes(value))
+            return cls(_iso_to_bytes(value))
```
ckeshava, line 117: nice update!

xrpl/models/transactions/transaction.py
```diff
@@ -423,7 +421,7 @@ def get_hash(self: Transaction) -> str:
 
     @classmethod
     def get_transaction_type(
-        cls: Type[Transaction], transaction_type: str
+        cls: Type[Self], transaction_type: str
     ) -> Type[Transaction]:
```
ckeshava, line 425:
```suggestion
    ) -> Type[Self]:
```
?

ckeshava (changes requested):
thanks mayukha, this must have taken a lot of patience.
I have a couple of doubts that need clarification

xrpl/models/transactions/transaction.py
mvadari: That is actually incorrect. If I call `Payment.get_transaction_type`, it's going to return type `Transaction`, not type `Payment`.

xrpl/core/binarycodec/types/account_id.py
mvadari: Classmethods are different than normal methods. Class methods are done as `Class.method()` as opposed to `instance.method()`. Since the first parameter is the type that the method is applied on, `instance instanceof Self` and `Class instanceof Type[Self]`.

tests/unit/utils/test_get_nftoken_id.py
```diff
@@ -1,4 +1,5 @@
 """Test the get_nftoken_id util."""
+
```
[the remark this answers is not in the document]
mvadari: `Self` isn't used in this file.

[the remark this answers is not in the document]
mvadari: `Self` isn't used in the test files. We don't care too much about keeping the test files perfectly typed - IMO adding `Self` to every single test adds more clutter than benefit.

xrpl/models/transactions/transaction.py
mvadari:
```
correct_type = cls.get_transaction_type(value["transaction_type"])
```
`cls.get_transaction_type` takes the transaction type passed into the dictionary and uses that model instead. So if you do `Payment.get_transaction_type(accountSetTx)`, that should redirect you to the `AccountSet` model. That's the check that's done here:
https://github.com/XRPLF/xrpl-py/blob/4779d796e6f808c952214e794be5d9d342412e14/xrpl/models/transactions/transaction.py#L355-L361

ckeshava: I agree with your explanation.

I wrote some unit tests to observe what happens if `Type[Self]` is returned - https://github.com/ckeshava/xrpl-py/tree/type_self

I suspect `Type[Transaction]` and `Type[Self]` are identical because this method (`get_trasaction_type`) is not overridden by any of the child classes

ckeshava (approved):

xrpl/models/transactions/transaction.py
mvadari: Exactly - which means that `Payment.get_transaction_type` still returns type `Transaction`, which is no longer `Self` (`Self` is now `Payment`).

ckeshava: hmm, it makes sense now. thanks

PR #708, opened 2024-05-17
feat: add model validation for types

mvadari:

This PR adds additional validation for models to also check parameter types and ensure they match what is expected. It also fixes issues that were uncovered via adding this additional checking.

### Context of Change

It's a common annoyance and improves the user experience.

### Type of Change

- [x] Yes

## Test Plan

Added tests, and added a few additional tests to improve code coverage in the `base_model.py` file. CI passes.


xrpl/models/transactions/transaction.py
```diff
@@ -266,9 +264,6 @@ def _get_errors(self: Transaction) -> Dict[str, str]:
             ] = """If ticket_sequence is provided,
             account_txn_id must be None and sequence must be None or 0"""
 
-        if isinstance(self.account, Wallet):
```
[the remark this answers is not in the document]
ckeshava: I get the following traceback when I tried to reproduce this error from one of the snippets code: 

```
  File "/Users/ckeshavabs/xrpl-py/snippets/bridge_transfer.py", line 37, in <module>
    fund_tx = XChainAccountCreateCommit(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ckeshavabs/xrpl-py/xrpl/models/utils.py", line 60, in new_init
    original_init(self, **kwargs)
  File "<string>", line 20, in __init__
  File "/Users/ckeshavabs/xrpl-py/xrpl/models/base_model.py", line 247, in __post_init__
    self.validate()
  File "/Users/ckeshavabs/xrpl-py/xrpl/models/base_model.py", line 258, in validate
    raise XRPLModelException(str(errors))
xrpl.models.exceptions.XRPLModelException: {'account': "account is <class 'xrpl.wallet.main.Wallet'>, expected <class 'str'>"}
```

snippets/paths.py
```diff
@@ -41,7 +43,7 @@
     account=wallet.address,
     amount=destination_amount,
     destination=destination_account,
-    paths=paths,
+    paths=[[PathStep(**path_data) for path_data in path] for path in paths],
```
ckeshava, line 46: Due to the more rigorous type checks, developers cannot "round-trip" the values i.e. the output of ripple-path-find is not longer compatible with the `Payment` transaction.

Can we write a utility (explicit) type-cast constructor that takes an input JSON and returns a 2-dimensional nested `PathStep` array?

```
def convertPathsJsonToList(paths: Dict[ ??? ]) -> List[List[PathStep]]:
    return [[PathStep(**path_data) for path_data in path] for path in paths]
```

xrpl/models/transactions/transaction.py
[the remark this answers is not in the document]
mvadari: The check is no longer needed, since it's handled by the model type validation code.

snippets/paths.py
mvadari: I thought dictionaries weren't handled properly in the `from_xrpl` logic, but turns out they are, so I added an extra check to allow dictionaries in place of models.

xrpl/models/base_model.py
```diff
@@ -266,20 +266,78 @@ def is_valid(self: BaseModel) -> bool:
...
+        if expected_type_origin is list:
+            # expected a List, received a List
+            # TODO: check for
+            if not isinstance(value, list):
+                return {"attr": f"{attr} is {type(value)}, expected {expected_type}"}
+            result = {}
+            for i in range(len(value)):
+                result.update(
+                    self._check_type(
+                        f"{attr}[{i}]", value[i], get_args(expected_type)[0]
```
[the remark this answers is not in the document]
mvadari: This is actually incorrect - `expected_type` doesn't have the same setup as the list, `get_args(expected_type)[0]` is the subtype of the list.

xrpl/models/transactions/transaction.py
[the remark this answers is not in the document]
ckeshava: hmm, ok. that makes sense

xrpl/models/base_model.py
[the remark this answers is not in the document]
ckeshava: the below flagged example involving `paths=[[PathStep(**path_data) for path_data in path] for path in paths],` in `snippets/paths.py` adheres to your hypothesis because all of the elements have the same type.

But can't we have a `List` of non-identical types in a model? Ex: `List[int, str, float]` ?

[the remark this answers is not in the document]
mvadari: No, it would be a `List[Union[str, int, float]]`, in which case the child object is still `Union[str, int, float]`. `Union`s are handled in a separate clause.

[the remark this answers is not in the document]
ckeshava: No, I understand Unions. Can't I group attributes of different data types into a `List` inside a model? Ex: `asset3: List[int, str, float]` ?

One could refactor this into three distinct attributes (or perhaps a `Tuple` type). Shouldn't we account for this possibility in the future?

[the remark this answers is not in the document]
mvadari: If you want to group different data types in a `List`, you'd have to do `List[Union[int, str, float]]`. If you want a `["str", 1, 1.2]` list in specifically that order, then that's a tuple. We don't have any tuple types, so none of the functions (including `from_xrpl`) support it. It won't be hard to add support for tuples if we do need tuple models, but I don't see that happening since rippled's serialization doesn't support that.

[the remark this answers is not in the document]
ckeshava: I see, okay. that makes sense

xrpl/models/base_model.py
```diff
@@ -314,7 +374,7 @@ def to_dict(self: BaseModel) -> Dict[str, Any]:
             if getattr(self, key) is not None
         }
 
-    def _to_dict_elem(self: BaseModel, elem: Any) -> Any:
+    def _to_dict_elem(self: Self, elem: Any) -> Any:
         if isinstance(elem, BaseModel):
```
ckeshava, line 378: are there any advantages to annotating the parameter `self: Self`, instead of `self: BaseModel` ? None of these functions are returning the `self` parameter, so I don't see the benefits of `typing.Self` class.

This was my take-away from the docs:
image
image

tests/unit/models/test_base_model.py
```diff
@@ -82,15 +83,44 @@ def test_is_dict_of_model_when_not_true(self):
...
+
+    def test_bad_type_flags(self):
+        transaction_dict = {
+            "account": 1,
+            "amount": 10,
+            "destination": 1,
+            "flags": "1234",
+        }
+        with self.assertRaises(XRPLModelException):
+            Payment(**transaction_dict)
```
ckeshava, line 103: I feel it's better to isolate the bad type to a single attribute, instead of coalescing with the previous test case 
```suggestion
            Payment(account="source_acct", destination="dest_acct", flags="1234", amount='200')
```
```suggestion
            Payment(**transaction_dict)
```

tests/unit/models/test_base_model.py
```diff
@@ -82,15 +83,44 @@ def test_is_dict_of_model_when_not_true(self):
...
+
+    def test_bad_type_enum(self):
+        path_find_dict = {
+            "subcommand": "blah",
```
ckeshava: I don't understand this test case. It does not throw an exception if `"subcommand": "create",` --  Hence, the validity rests more on the content of the string, rather than it's data type.


Am I understanding this correctly? Where is this bad enum?

mvadari: The `subcommand` param in the `PathFindResponse` is an enum: https://github.com/XRPLF/xrpl-py/blob/main/xrpl/models/requests/path_find.py#L83

tests/unit/models/test_base_model.py
```diff
@@ -388,6 +418,83 @@ def test_from_dict_submit(self):
...
+                "currency": "USD",
+                "issuer": "raoV5dkC66XvGWjSzUhCUuuGM3YFTitMxT",
+                "value": "100",
+            },
+        }
+        with self.assertRaises(XRPLModelException):
+            TrustSet.from_dict(tx)
+
+    def test_bad_literal(self):
+        tx = {
```
ckeshava, line 436: could you add a comment and point out the `bad_literal` attribute in this test case?

xrpl/models/base_model.py
ckeshava: ok

tests/unit/models/test_base_model.py
ckeshava: ah okay

tests/unit/models/test_base_model.py
```diff
@@ -397,14 +504,18 @@ def test_from_xrpl(self):
             for test in fixtures_json["transactions"]:
                 x_json = test["xjson"]
                 r_json = test["rjson"]
-                with self.subTest(json=x_json):
+                with self.subTest(json=x_json, use_json=False):
```
ckeshava, line 507: The `use_json` parameter in these function calls are purely for debugging purposes right? I flipped the `True/False` values in these three tests, it did not affect the tests in any way.

image

ckeshava:
thanks, this is very interesting work. I feel we could have used this library to enforce type checks, had it been better maintained: https://pypi.org/project/enforce-typing/

tests/unit/models/test_base_model.py
mvadari: Yes. Note how the `use_json=True` test is the only one that uses `json.dumps`. I did that for better code coverage.

xrpl/models/base_model.py
```diff
@@ -284,20 +284,80 @@ def is_valid(self: BaseModel) -> bool:
...
+            )
+
+        if expected_type_origin is Literal:
+            arg = get_args(expected_type)
+            return {} if value in arg else {attr: f"{attr} is {value}, expected {arg}"}
+
+        if issubclass(expected_type, BaseModel) and isinstance(value, dict):
+            return {}
+
+        if not isinstance(value, expected_type):
```
ckeshava, line 339: What about `Optional` type?  `return {} # last line of the function` will be hit for the `Optional` type. I think we'd want to validate it separately right?

tests/unit/models/test_base_model.py
ckeshava: Hmm right. In that test case, `json.dumps` returns a JSON string, rather than a Python object. This forces `Transaction.from_xrpl` to use a different codepath, namely the first type in the Union: `def from_xrpl(cls: Type[BM], value: Union[str, Dict[str, Any]]) -> BM:`

good observation

xrpl/models/base_model.py
mvadari: `Optional` is `Union[type, None]` under the hood.

ckeshava (approved):

mvadari:
@coderabbitai full review

mvadari:
@coderabbitai full review

tests/unit/models/test_base_model.py
```diff
@@ -392,49 +422,83 @@ def test_from_dict_submit(self):
...
+    def test_bad_literal(self):
+        tx = {
+            "account": issuer,
+            "xchain_bridge": {
+                "locking_chain_door": issuer,
+                "locking_chain_issue": {"currency": "XRP"},
+                "issuing_chain_door": issuer,
+                "issuing_chain_issue": {"currency": "XRP"},
+            },
+            "public_key": "0342E083EA762D91D621714C394",
```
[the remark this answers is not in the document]
mvadari: This is a sample string (that isn't even a valid public key), just used as a placeholder.

xrpl/models/base_model.py
```diff
@@ -291,18 +291,79 @@ def is_valid(self: Self) -> bool:
...
+            return {attr: f"{attr} is {type(value)}, expected {expected_type}"}
+
+        # unsure what the problem with mypy is here
+        if expected_type is Any:  # type: ignore[comparison-overlap]
+            return {}
+
+        if expected_type_origin is list:
+            # expected a List, received a List
+            if not isinstance(value, list):
+                return {"attr": f"{attr} is {type(value)}, expected {expected_type}"}
```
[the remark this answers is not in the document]
mvadari: Fixed in [786ad8d](https://github.com/XRPLF/xrpl-py/pull/708/commits/786ad8dd59d332e098dcdde78823c930e19c528f)

tests/unit/models/test_base_model.py
```diff
@@ -83,15 +84,44 @@ def test_is_dict_of_model_when_not_true(self):
...
+
+    def test_bad_type_flags(self):
+        transaction_dict = {
+            "account": account,
+            "amount": value,
+            "destination": destination,
+            "flags": "1234",  # should be an int
+        }
+        with self.assertRaises(XRPLModelException):
+            Payment(**transaction_dict)
```
khancode, line 104: Can you also validate the error message thrown?

khancode (approved):
LGTM

Issue #709, opened 2024-05-22
AsyncWebsocketClient uses ever increasing memory with subsequent requests


ubuddemeier:
I think the problem is that received messages are enqueued (in WebsocketBase._handler) regardless of whether the message corresponds to a request or not:

https://github.com/XRPLF/xrpl-py/blob/main/xrpl/asyncio/clients/websocket_base.py#L139

This should probably be inside of an else-clause to the preceding if-clause.

ckeshava:
Hello, thanks for pointing this out.

How did you observe the increasing memory load? Are you using any profiler?

Jbekker:
@ckeshava ,

I'm noticing the same, it's best seen when you do a larger requests like requesting BookOffers. If you can don't have anything iterating over the client, the queue continuously grows.
```
import asyncio

from xrpl.asyncio.clients import AsyncWebsocketClient
from xrpl.models.requests import BookOffers
from xrpl.models.currencies import XRP, IssuedCurrency
from xrpl.utils import (
    drops_to_xrp,
    get_balance_changes,
    get_order_book_changes,
    hex_to_str,
    ripple_time_to_posix,
    xrp_to_drops,
)

async def main():
    async with AsyncWebsocketClient("wss://s1.ripple.com") as client:
                
                while True:
                    orderbook_asks_info = await client.request(
                        BookOffers(
                            ledger_index="current",
                            taker_gets=XRP(),
                            taker_pays=IssuedCurrency(
                                  currency="USD",
                                  issuer="rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq"
                            ),
                            limit=100,
                        )
                    )

                    orderbook_bids_info = await client.request(
                        BookOffers(
                            ledger_index="current",
                            taker_gets=IssuedCurrency(
                                  currency="USD",
                                  issuer="rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq"
                            ),
                            taker_pays=XRP(),
                            limit=100,
                        )
                    )

                    await asyncio.sleep(2)

if __name__ == "__main__":
    asyncio.run(main())
```

ckeshava:
thanks, this is helpful @Jbekker

ckeshava:
@Jbekker 
I have incorporated your test case at the tip of this branch: https://github.com/XRPLF/xrpl-py/pull/713 -- many thanks, this is insightful.

[...]

PR #713, opened 2024-06-06
[FIX] Issue #709: do not enqueue all responses

ckeshava:
Do not enqueue messages that correspond to an existing request

### Context of Change
Bug Fix: At the moment, we are enqueuing all the incoming messages.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes
This change impacts the memory usage of the system, hence is not directly perceivable by end users.


ckeshava:
I'm not able to find requests which increase the queue size
```
class TestWebSocketClient(TestCase):
    """Performace test of the async web_socket_client."""

    def test_web_socket_client(self: TestWebSocketClient) -> None:
        URL = "wss://s.devnet.rippletest.net:51233"

        with WebsocketClient(URL) as client:
            pendings_msg_size = []

            for _ in range(100):
                client.request(ServerInfo())
                pendings_msg_size.append(len(client._open_requests))

            print(max(pendings_msg_size)) # returns 0
```

mvadari:
> I'm not able to find requests which increase the queue size
> 
> ```
> class TestWebSocketClient(TestCase):
>     """Performace test of the async web_socket_client."""
> 
>     def test_web_socket_client(self: TestWebSocketClient) -> None:
>         URL = "wss://s.devnet.rippletest.net:51233"
> 
>         with WebsocketClient(URL) as client:
>             pendings_msg_size = []
> 
>             for _ in range(100):
>                 client.request(ServerInfo())
>                 pendings_msg_size.append(len(client._open_requests))
> 
>             print(max(pendings_msg_size)) # returns 0
> ```

Try `client._messages` instead of `client._open_requests`.

tests/integration/it_utils.py
```diff
@@ -119,8 +122,13 @@ async def fund_wallet_async(
         destination=wallet.address,
         amount=FUNDING_AMOUNT,
     )
-    await sign_and_submit_async(payment, client, MASTER_WALLET, check_fee=True)
-    await client.request(LEDGER_ACCEPT_REQUEST)
+
+    # concurrent wallet_fund operations will attempt to advance the ledger at the same
+    # time. Consequently, all the funding operations fail, except for the first one.
+    # using a lock will serialize the access to this critical operation
+    async with async_wallet_fund_lock:
```
ckeshava, line 129: @mvadari  I found that adding a lock to this section of the code eliminated the "Account Not Found" error.

ckeshava:
> _messages

`client._messages` has a `qsize` of `0`

tests/integration/reusable_values.py
```diff
@@ -29,13 +29,15 @@
...
 async def _set_up_reusable_values():
     wallet = Wallet.create()
-    await fund_wallet_async(wallet)
     destination = Wallet.create()
-    await fund_wallet_async(destination)
     door_wallet = Wallet.create()
-    await fund_wallet_async(door_wallet)
     witness_wallet = Wallet.create()
-    await fund_wallet_async(witness_wallet)
+    await asyncio.gather(
```
mvadari, line 35: Please stick to one change per PR - this is unrelated to this PR.

mvadari:
How did you test this change? If not by adding a test case, then what?

ckeshava:
I don't have a good idea to test this change. But irrespective of the impact on this issue, I think this change is useful.

I don't know which inputs would reproduce the issue description.

tests/integration/clients/test_async_websocket_client.py
```diff
@@ -0,0 +1,91 @@
+"""Memory usage of Websocket clients."""
```
mvadari, line 1: This file should be called `test_websocket_client` and the description should be more general, as it may be used for other tests in the future.

tests/integration/clients/test_async_websocket_client.py
```diff
@@ -0,0 +1,91 @@
...
+try:
+    from unittest import IsolatedAsyncioTestCase
+except ImportError:
+    from aiounittest import AsyncTestCase as IsolatedAsyncioTestCase  # type: ignore
+
+
+class TestAsyncWebsocketClient(IsolatedAsyncioTestCase):
+    """Memory usage of async-websocket client"""
+
+    async def test_msg_queue_async_websocket_client(
```
mvadari, line 22: This test can still use the `test_async_and_sync` framework - there's a `websockets_only` parameter in the decoration.

xrpl/asyncio/clients/websocket_base.py
```diff
@@ -135,9 +136,10 @@ async def _handler(self: WebsocketBase) -> None:
             # if this response corresponds to request, fulfill the Future
             if "id" in response_dict and response_dict["id"] in self._open_requests:
                 self._open_requests[response_dict["id"]].set_result(response_dict)
-
-            # enqueue the response for the message queue
-            cast(_MESSAGES_TYPE, self._messages).put_nowait(response_dict)
+                # a response that fulfills a future is not enqueued again
```
mvadari, line 139: This comment is more confusing than useful IMO.

tests/integration/clients/test_async_websocket_client.py
ckeshava: The current version of the code uses `test_async_and_sync` construct, thanks for your suggestion.

but I don't like this version. Due to the use of `await client.request(...)`, the requests are fulfilled in a serialized fashion. The queue size will never exceed one in this test.

Ideally, here is what I'd like to do: The `asyncio.create_task(...)` concurrently creates many requests. Now, we can verify that the responses are *not* logged into the queue.

```
class TestAsyncWebsocketClient(IntegrationTestCase):
    """Memory usage of async-websocket client"""

    @test_async_and_sync(globals(), websockets_only=True)
    async def test_msg_queue_async_websocket_client(self, client):
        """Test the rate of growth of the Message queue in async_websocket_client under
        persistent load. Admittedly, this is not a precise measure, rather its a proxy
        to measure the memory footprint of the client
        """

        for _ in range(5):
            asyncio.create_task(
                client.request(
                    BookOffers(
                        ledger_index="current",
                        taker_gets=XRP(),
                        taker_pays=IssuedCurrency(
                            currency="USD", issuer="rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq"
                        ),
                        limit=500,
                    )
                )
            )

            asyncio.create_task(
                client.request(
                    BookOffers(
                        ledger_index="current",
                        taker_gets=IssuedCurrency(
                            currency="USD", issuer="rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq"
                        ),
                        taker_pays=XRP(),
                        limit=500,
                    )
                )
            )

            # messages corresponding to the above two requests might reside in the
            # queue. If these requests have already been fulfilled, the queue size will be reduced by an appropriate amount
            self.assertEqual(client._messages.qsize() <= 2, True)

        # wait for some time, we want to ensure that all the requests are registered in
        # the websocket client. asyncio library asynchronously populates the queue
        await asyncio.sleep(2)
        # the messages queue has not increased in proportion to the requests/responses
        self.assertEqual(client._messages.qsize() <= 2, True)

```

I'm trying to accomodate `asyncio.create_task` inside `test_async_and_sync` decorator.

what do you think?

xrpl/asyncio/clients/websocket_base.py
ckeshava: fixed in [3a53ffc](https://github.com/XRPLF/xrpl-py/pull/713/commits/3a53ffc66ab7b99675d8a5fbb59aa52ae73d6d4f)

tests/integration/clients/test_async_websocket_client.py
mvadari: What about using subscriptions? Can those stay in the queue?

ckeshava: I'm not able to test the `Subscribe` method correctly. Contrary to my understanding, this code print only one message and gets stuck. It does not indefinitely print the stream of messages from the other peer.

What am I missing? The structure of the code is similar to the tests [here](https://github.com/XRPLF/xrpl-py/blob/main/tests/integration/reqs/test_subscribe.py)

```
class TestSubscriptionWebsocketClient(IntegrationTestCase):
    @test_async_and_sync(globals(), websockets_only=True)
    async def test_msg_queue_size(self, client):
        req = Subscribe(streams=[StreamParameter.LEDGER])
        # NOTE: this code will run forever without a timeout, until the process is killed
        await client.send(req)

        # Note: AsyncWebSocket clients need the "async for" construct for execution.
        # Alternatively, we will need to implement __aiter__ method in the respective class.
        async for message in client:
            print(message)
            print("Queue size: ")
            print(client._messages.qsize())
```

mvadari: If you're connecting to a standalone node, you also need to close ledgers independently. A standalone node won't do that automatically for you. That's why there's the additional `await accept_ledger_async()` in the for loop in the test.

Alternatively, you can add `use_testnet=True` in the `test_async_and_sync` decorator.

mvadari: Might also be worth considering rewriting the `use_testnet=True` parameter into something that'll also allow you to connect to mainnet (i.e. not a boolean, so you can connect to other things as well).

ckeshava: got it, thanks.

`Subscribe` does use the above message queue. Every response from the subscription stream adds one message, but that message is removed almost immediately. I don't know which method is popping out the messages, but the message queue is empty.

are you proposing that we use the `Subscribe` request for the message queue tests? but, it is not representative of a request-response back-and-forth

mvadari: Ok, fair. I was just thinking about commands that would legitimately grow the queue.

tests/integration/clients/test_websocket_client.py
```diff
@@ -0,0 +1,52 @@
...
+from tests.integration.it_utils import test_async_and_sync
+from xrpl.models.currencies import XRP, IssuedCurrency
+from xrpl.models.requests import BookOffers
+from xrpl.models.response import ResponseStatus
+
+
+class TestWebsocketClient(IntegrationTestCase):
+    """Memory usage of websocket client"""
+
+    @test_async_and_sync(globals(), websockets_only=True)
```
mvadari, line 15: This test is against standalone, not mainnet. The data isn't going to really test anything.

ckeshava: Testing with a Standalone mode provides a useful baseline, because the performance will only get worse with Mainnet/Testnet.

I'm measuring the memory usage of the client when multiple successive requests are input to the client. (Since no new transactions are provided, there is no need to advance the ledger)

I can set `use_testnet=True`, I'm not opposed to that.

ckeshava: The current integration test framework does not easily let us test against the mainnet (unless we write additional utility methods). I'm assuming you're referring to the `use_testnet` parameter in `test_async_and_sync` decorator.

mvadari: The integration test framework can be changed, it's not set in stone.

ckeshava: are you suggesting that I test the performance against the mainnet? Why would that be different than testing with the testnet or the standalone node?

The performance of the xrpl-py client is in question, does it matter if I use a standalone/testnet/mainnet server for this test?

mvadari: There isn't enough traffic on testnet/standalone to test the performance of xrpl-py, it's not going to run into any issues. There may not even be any data in this pair (or any pair) for the few seconds that you're testing on. You could theoretically generate that traffic yourself, but that'd require a much more complex test.

ckeshava:
This work is very old, I have lost context of the issue. If somebody finds this work useful, please let me know. I can rebase and revive this work.

PR #715, opened 2024-06-12
[Fix #710] Use network_id to identify faucet_url

### LINKED ISSUE #710: Use Network ID to determine the correct faucet to use
Code reference: https://github.com/XRPLF/xrpl-py/blob/main/xrpl/asyncio/wallet/wallet_generation.py

Currently, the faucet generation code guesses the correct faucet to use based on substrings in the URL. This isn't a very robust system. It would be better (and more robust) to use the Network ID.

ckeshava:
- use network_id to disambiguate between different XRPL chains
- move the utility function _get_network_id_and_build_version into client.py file for convinience

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes
This change should not modify the experience for external developers

## Test Plan

Tests have been modified to accomodate the new API for grabbing the faucet_url


xrpl/asyncio/clients/client.py
```diff
@@ -51,3 +54,24 @@ async def _request_impl(
         :meta private:
         """
         pass
+
+
+async def _get_network_id_and_build_version(client: Client) -> None:
```
mvadari, line 59: Why not automatically run this on every client upon instantiation, instead of having it be a helper function? Then it only needs to be run once, instead of every time a function that uses this is called.

ckeshava: We are not duplicating invocations of this function. Here are [two](https://github.com/XRPLF/xrpl-py/blob/4779d796e6f808c952214e794be5d9d342412e14/xrpl/asyncio/transaction/main.py#L233) [examples](https://github.com/XRPLF/xrpl-py/blob/ea6365af1b3da29c45f4441fefcfb298f92d897d/xrpl/asyncio/wallet/wallet_generation.py#L62) where the `network_id` information persists in the client object.

But I agree, this could be done in a cleaner way.
I can only invoke `_get_network_id_and_build_version` function after a client has opened the socket-connection. So, I'll need to invoke this in this `open()` method of all the child classes of `Client`

let me try it out

mvadari: I'm open to the option of doing it lazily - i.e. only if and as needed. All that would require is checking if the network ID has been instantiated before running the request.

ckeshava: okay, I've added a check: [23d3eb3](https://github.com/XRPLF/xrpl-py/pull/715/commits/23d3eb31df830a1c55f2cae51e720fde14c673a3)

xrpl/asyncio/transaction/main.py
```diff
@@ -1,17 +1,19 @@
 """High-level transaction methods with XRPL transactions."""
+
 import math
 from typing import Any, Dict, Optional, cast
 
 from typing_extensions import Final
 
 from xrpl.asyncio.account import get_next_valid_seq_number
 from xrpl.asyncio.clients import Client, XRPLRequestFailureException
+from xrpl.asyncio.clients.client import _get_network_id_and_build_version
```
mvadari, line 10: Any function that gets imported elsewhere shouldn't start with a `_`.

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -57,7 +59,9 @@ async def generate_faucet_wallet(
 
     .. # noqa: DAR402 exception raised in private method
     """
-    faucet_url = get_faucet_url(client.url, faucet_host)
+    if not client.network_id:
+        await _get_network_id_and_build_version(client)
```
mvadari, line 63:
```suggestion
    await _get_network_id_and_build_version(client)
```
It's already in the helper function.

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,34 +154,40 @@ def process_faucet_host_url(input_url: str) -> str:
...
 
     Returns:
         The URL of the matching faucet.
 
     Raises:
-        XRPLFaucetException: if the provided URL is not for the testnet or devnet.
+        XRPLFaucetException: if the provided network_id does not correspond to testnet
+            or devnet.
     """
+    if network_id is not None:
```
mvadari, line 176: This check shouldn't be needed (or should throw an error) because the network ID should never be `None`.

ckeshava: well, network_id is `None` in some cases -- the rippled integration tests do not specify a `network_id` in the config file, hence this value is `None`

I've updated the config file in one of the recent xrpl-py PRs to fix this issue. Nevertheless, I don't think it's good idea to throw an error if `network_id == None`

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,34 +154,40 @@ def process_faucet_host_url(input_url: str) -> str:
...
+            return _TEST_FAUCET_URL
+        elif network_id == 2:
+            return _DEV_FAUCET_URL
+        elif network_id == 0:
+            raise XRPLFaucetException("Cannot create faucet with a client on mainnet.")
+        elif network_id < 0:
+            raise XRPLFaucetException("network_id cannot be negative.")
+
     if faucet_host is not None:
         return process_faucet_host_url(faucet_host)
```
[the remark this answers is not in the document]
ckeshava: 👍  fixed in  ab918dc

xrpl/asyncio/transaction/main.py
ckeshava: fixed in ab918dc

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,34 +154,40 @@ def process_faucet_host_url(input_url: str) -> str:
...
+
+    if network_id is not None:
+        if network_id == 1:
+            return _TEST_FAUCET_URL
+        elif network_id == 2:
+            return _DEV_FAUCET_URL
+        elif network_id == 0:
+            raise XRPLFaucetException("Cannot create faucet with a client on mainnet.")
+        elif network_id < 0:
+            raise XRPLFaucetException("network_id cannot be negative.")
```
mvadari, line 187: Does rippled allow this scenario?

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,34 +154,40 @@ def process_faucet_host_url(input_url: str) -> str:
...
+
     raise XRPLFaucetException(
-        "Cannot fund an account with a client that is not on the testnet or devnet."
+        "Cannot create faucet URL without network_id or the faucet_host information."
     )
```
mvadari, line 191: Thoughts on throwing this error in the main `generate_faucet_wallet` function instead of here? e.g.
```python
if network_id is None and faucet_host is None:
  raise XRPLFaucetException(...)
```

ckeshava: I don't object to that idea, but how is that different from the current implementation?

At line 189 (the last line of the `get_faucet_url` method), we have established that `faucet_host == None`.  You are correct, an unknown devnet might have a networkID, but how can I determine the host_url from a networkID (say `1345` ?)

ckeshava: > ```python
> if network_id is None and faucet_host is None:
>   raise XRPLFaucetException(...)
> ```
The said devnet will have a networkID, hence this condition needs to be:
```python
if faucet_host is None and networkID >= 1025:
  raise XRPLFaucetException(...)
```

The current implementation is equivalent to the above if-condition

mvadari: The difference comes from typing and failing fast. IMO it makes more sense for `get_faucet_url` to not accept `None` as a valid network ID.

You can't determine the host URL from an unknown Devnet, but the error message should be different.

xrpl/asyncio/wallet/wallet_generation.py
ckeshava: the config file does not allow negative numbers as input for network_id. But it accepts really large numbers (for eg: `4294967295` which could be negative numbers, if typecast into smaller signed-types)

also, rippled unit tests allow negative numbers - https://github.com/XRPLF/rippled/compare/develop...ckeshava:rippled:negNetworkID

mvadari: I don't think xrpl-py needs any specific checks for this, "we don't understand this network ID" is fine.

xrpl/asyncio/wallet/wallet_generation.py
ckeshava: we cannot expect all networks to have a valid `NetworkID` isn't it? 
If we force `get_faucet_url` to accept valid networkIDs only, then some of the processing logic shifts into `generate_faucet_wallet` method.

```
def get_faucet_url(network_id: int) -> str:
    if network_id == 1:
        return _TEST_FAUCET_URL
    elif network_id == 2:
        return _DEV_FAUCET_URL
    elif network_id == 0:
        raise XRPLFaucetException("Cannot create faucet with a client on mainnet.")
    elif network_id < 0:
        raise XRPLFaucetException("network_id cannot be negative.")

    # unreachable code
    return ""

async def generate_faucet_wallet(
    client: Client,
    wallet: Optional[Wallet] = None,
    debug: bool = False,
    faucet_host: Optional[str] = None,
    usage_context: Optional[str] = None,
    user_agent: Optional[str] = "xrpl-py",
) -> Wallet:
    if not client.network_id:
        await get_network_id_and_build_version(client)

    if client.network_id is None and faucet_host is None:
        raise XRPLFaucetException(
            "Cannot create faucet URL without network_id or the faucet_host information"
        )

    if faucet_host is not None:
        faucet_url = process_faucet_host_url(faucet_host)
    else:
        faucet_url = get_faucet_url(client.network_id)
        
    # remaining code in the function ...

```

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,34 +154,40 @@ def process_faucet_host_url(input_url: str) -> str:
...
     Returns:
         The URL of the matching faucet.
 
     Raises:
-        XRPLFaucetException: if the provided URL is not for the testnet or devnet.
+        XRPLFaucetException: if the provided network_id does not correspond to testnet
+            or devnet.
     """
     if faucet_host is not None:
         return process_faucet_host_url(faucet_host)
```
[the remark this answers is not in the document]
ckeshava: yeah, I put down a snippet in a comment below

xrpl/asyncio/wallet/wallet_generation.py
mvadari: That flow makes more sense to me, personally.

ckeshava: alrighty

tests/unit/asyn/wallet/test_wallet.py
```diff
@@ -16,30 +17,16 @@ def test_wallet_get_xaddress(self):
...
-    def test_get_faucet_wallet_custom(self):
-        json_client_url = "https://s.devnet.rippletest.net:51234"
-        ws_client_url = "wss://s.devnet.rippletest.net/"
-        custom_host = "my_host.org"
-        expected_faucet = "https://my_host.org/accounts"
-
-        self.assertEqual(get_faucet_url(json_client_url, custom_host), expected_faucet)
-        self.assertEqual(get_faucet_url(ws_client_url, custom_host), expected_faucet)
-
-    def test_get_faucet_wallet_test(self):
```
ckeshava, line 36: these test cases are already covered in the below class `TestProcessFaucetHostURL`

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,35 +162,32 @@ def process_faucet_host_url(input_url: str) -> str:
...
-    raise XRPLFaucetException(
-        "Cannot fund an account with a client that is not on the testnet or devnet."
-    )
+    elif network_id == 0:
+        raise XRPLFaucetException("Cannot create faucet with a client on mainnet.")
+    elif network_id < 0:
+        raise XRPLFaucetException("Unable to parse the specified network_id.")
+
+    # this line is unreachable. Custom devnets must specify a faucet_host input
+    return ""
```
mvadari, line 190: You should raise an error here.

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,35 +162,32 @@ def process_faucet_host_url(input_url: str) -> str:
...
-    if faucet_host is not None:
-        return process_faucet_host_url(faucet_host)
-    if "altnet" in url or "testnet" in url:  # testnet
+    if network_id == 1:
         return _TEST_FAUCET_URL
-    if "sidechain-net2" in url:  # sidechain issuing chain devnet
-        raise XRPLFaucetException(
-            "Cannot fund an account on an issuing chain. Accounts must be created via "
-            "the bridge."
-        )
```
mvadari, line 176: This piece still needs to be included. The network ID is 262.

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,35 +162,32 @@ def process_faucet_host_url(input_url: str) -> str:
...
 
     Raises:
-        XRPLFaucetException: if the provided URL is not for the testnet or devnet.
+        XRPLFaucetException: if the provided network_id does not correspond to testnet
+            or devnet.
     """
-    if faucet_host is not None:
-        return process_faucet_host_url(faucet_host)
-    if "altnet" in url or "testnet" in url:  # testnet
+    if network_id == 1:
```
mvadari, line 180: Can this if-else be replaced with a dictionary? It'll be easier to maintain that way.

ckeshava: updated in [f06e5c1](https://github.com/XRPLF/xrpl-py/pull/715/commits/f06e5c15e618816ee7478909695f37a38618de1d)

xrpl/asyncio/wallet/wallet_generation.py
ckeshava: thanks. fixed in [f06e5c1](https://github.com/XRPLF/xrpl-py/pull/715/commits/f06e5c15e618816ee7478909695f37a38618de1d)

xrpl/asyncio/wallet/wallet_generation.py
ckeshava: fixed in [f06e5c1](https://github.com/XRPLF/xrpl-py/pull/715/commits/f06e5c15e618816ee7478909695f37a38618de1d)

xrpl/asyncio/wallet/wallet_generation.py
ckeshava: updated in [f06e5c1](https://github.com/XRPLF/xrpl-py/pull/715/commits/f06e5c15e618816ee7478909695f37a38618de1d)

xrpl/asyncio/wallet/wallet_generation.py
ckeshava: updated sometime before [f06e5c1](https://github.com/XRPLF/xrpl-py/pull/715/commits/f06e5c15e618816ee7478909695f37a38618de1d)

xrpl/asyncio/wallet/wallet_generation.py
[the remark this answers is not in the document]
ckeshava: updated somewhere prior to [f06e5c1](https://github.com/XRPLF/xrpl-py/pull/715/commits/f06e5c15e618816ee7478909695f37a38618de1d)

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Allow empty strings for the purpose of removing fields in DIDSet transaction
+- Refactor the API of `get_faucet_wallet` method. This produces a non-ambiguous URL for Testnet and Devnet faucet hosts.
```
mvadari, line 15:
```suggestion
- Use `NetworkID` in faucet processing to produce a non-ambiguous URL for faucet hosts.
```

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,34 +162,40 @@ def process_faucet_host_url(input_url: str) -> str:
...
-        XRPLFaucetException: if the provided URL is not for the testnet or devnet.
+        XRPLFaucetException: if the provided network_id does not correspond to testnet
+            or devnet.
     """
-    if faucet_host is not None:
-        return process_faucet_host_url(faucet_host)
-    if "altnet" in url or "testnet" in url:  # testnet
-        return _TEST_FAUCET_URL
-    if "sidechain-net2" in url:  # sidechain issuing chain devnet
+    map_network_id_to_url: Dict[int, str] = {1: _TEST_FAUCET_URL, 2: _DEV_FAUCET_URL}
```
mvadari, line 180: This object should be outside of the function, since it's not dependent on the data inside. And the name should be in uppercase.

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -150,34 +162,40 @@ def process_faucet_host_url(input_url: str) -> str:
...
-        return _DEV_FAUCET_URL
+    elif network_id == 0:
+        raise XRPLFaucetException("Cannot create faucet with a client on mainnet.")
+    elif network_id < 0:
+        raise XRPLFaucetException("Unable to parse the specified network_id.")
+
+    # this line is unreachable. Custom devnets must specify a faucet_host input
     raise XRPLFaucetException(
-        "Cannot fund an account with a client that is not on the testnet or devnet."
+        "The specified network_id " + str(network_id) + " is not recognizable"
```
mvadari, line 198:
```suggestion
        "The specified network_id " + str(network_id) + " does not have a known faucet."
```

xrpl/asyncio/wallet/wallet_generation.py
```diff
@@ -17,6 +17,7 @@
 _DEV_FAUCET_URL: Final[str] = "https://faucet.devnet.rippletest.net/accounts"
 
 _TIMEOUT_SECONDS: Final[int] = 40
+_MAP_NETWORK_ID_TO_URL: Dict[int, str] = {1: _TEST_FAUCET_URL, 2: _DEV_FAUCET_URL}
```
mvadari, line 20: nit:
```suggestion
_NETWORK_ID_URL_MAP: Dict[int, str] = {1: _TEST_FAUCET_URL, 2: _DEV_FAUCET_URL}
```

mvadari (approved):

anissa-ripple (approved):

Issue #716, opened 2024-06-17
Delete data, URI, DIDDocument fields in DID object


ckeshava:
It is not possible to delete Data, URI and DIDDocument fields in a DID object. Users encounter a "data" : "Must be a hex" error. This is contrary to the expected behavior in the docs: 
```
Note: To delete the Data, DIDDocument, or URI field from an existing DID ledger entry, add the field as an empty string.
```
Source: https://xrpl.org/docs/references/protocol/transactions/types/didset/

A unit test to reproduce this issue can be found at the tip of this branch: https://github.com/ckeshava/xrpl-py/tree/DIDemptyData

Many thanks to this Discord user (0xΞarl) for bringing this to our attention: https://discord.com/channels/886050993802985492/886095603669544991/1251546703602126888

PR #717, opened 2024-06-17
Empty strings in DIDSet transaction

ckeshava:
Fix https://github.com/XRPLF/xrpl-py/issues/716

### Context of Change

Fields within a DID object can only be deleted by setting an empty string. However, the regex in the `DIDSet` model do not account for empty strings.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Unit tests have been added to this effect

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [[Unreleased]]
+
+### Added
 - Add support for the DeliverMax field in Payment transactions
+
+### Fixed
 - Allow empty strings for the purpose of removing fields in DIDSet transaction
 
 ## [2.6.0] - 2024-06-03
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [[Unreleased]]
 - Add support for the DeliverMax field in Payment transactions
+- Allow empty strings for the purpose of removing fields in DIDSet transaction
 
 ## [2.6.0] - 2024-06-03
 
--- a/tests/unit/models/transactions/test_did_set.py
+++ b/tests/unit/models/transactions/test_did_set.py
@@ -63,7 +63,12 @@ class TestDIDSet(TestCase):
                 "{'did_set': 'Must have one of `did_document`, `data`, and `uri`.'}",
             )
 
-    def test_delete_data_field(self):
+    def test_DIDSet_all_empty_fields(self):
+        with self.assertRaises(XRPLModelException):
+            DIDSet(account=_ACCOUNT, data="", did_document="", uri="")
+
+    def test_delete_multiple_DID_fields(self):
+        # create a valid DID object
         tx = DIDSet(
             account=_ACCOUNT,
             did_document=_VALID_FIELD,
@@ -76,6 +81,27 @@ class TestDIDSet(TestCase):
         tx = DIDSet(
             account=_ACCOUNT,
             data="",
+        )
+
+        self.assertTrue(tx.is_valid())
+
+        # delete URI field
+        tx = DIDSet(
+            account=_ACCOUNT,
+            uri="",
+        )
+
+        self.assertTrue(tx.is_valid())
+
+        # delete did_document field
+
+        # Note: It is invalid to render a DID LedgerObject empty by removing all the
+        # fields. However, that condition can only be detected by reading the
+        # LedgerEntry, which is out-of-scope for the library. rippled generated a
+        # `tecEMPTY_DID` error code in this case
+        tx = DIDSet(
+            account=_ACCOUNT,
+            did_document="",
         )
 
         self.assertTrue(tx.is_valid())
--- a/tests/unit/models/transactions/test_did_set.py
+++ b/tests/unit/models/transactions/test_did_set.py
@@ -77,31 +77,10 @@ class TestDIDSet(TestCase):
         )
         self.assertTrue(tx.is_valid())
 
-        # delete data field
+        # remove the data field from the above DID object
         tx = DIDSet(
             account=_ACCOUNT,
             data="",
-        )
-
-        self.assertTrue(tx.is_valid())
-
-        # delete URI field
-        tx = DIDSet(
-            account=_ACCOUNT,
-            uri="",
-        )
-
-        self.assertTrue(tx.is_valid())
-
-        # delete did_document field
-
-        # Note: It is invalid to render a DID LedgerObject empty by removing all the
-        # fields. However, that condition can only be detected by reading the
-        # LedgerEntry, which is out-of-scope for the library. rippled generated a
-        # `tecEMPTY_DID` error code in this case
-        tx = DIDSet(
-            account=_ACCOUNT,
-            did_document="",
         )
 
         self.assertTrue(tx.is_valid())
--- a/tests/unit/models/transactions/test_did_set.py
+++ b/tests/unit/models/transactions/test_did_set.py
@@ -63,11 +63,11 @@ class TestDIDSet(TestCase):
                 "{'did_set': 'Must have one of `did_document`, `data`, and `uri`.'}",
             )
 
-    def test_DIDSet_all_empty_fields(self):
+    def test_all_empty_fields(self):
         with self.assertRaises(XRPLModelException):
             DIDSet(account=_ACCOUNT, data="", did_document="", uri="")
 
-    def test_delete_multiple_DID_fields(self):
+    def test_delete_multiple_fields(self):
         # create a valid DID object
         tx = DIDSet(
             account=_ACCOUNT,
--- a/tests/unit/models/transactions/test_did_set.py
+++ b/tests/unit/models/transactions/test_did_set.py
@@ -67,7 +67,7 @@ class TestDIDSet(TestCase):
         with self.assertRaises(XRPLModelException):
             DIDSet(account=_ACCOUNT, data="", did_document="", uri="")
 
-    def test_remove_data_field(self):
+    def test_empty_data_field(self):
         # create a valid DID object
         tx = DIDSet(
             account=_ACCOUNT,
--- a/tests/unit/models/transactions/test_did_set.py
+++ b/tests/unit/models/transactions/test_did_set.py
@@ -58,6 +58,10 @@ class TestDIDSet(TestCase):
             DIDSet(
                 account=_ACCOUNT,
             )
+        self.assertEqual(
+            error.exception.args[0],
+            "{'did_set': 'Must have one of `did_document`, `data`, and `uri`.'}",
+        )
 
     def test_create_did_object_all_empty_fields(self):
         with self.assertRaises(XRPLModelException):
--- a/tests/unit/models/transactions/test_did_set.py
+++ b/tests/unit/models/transactions/test_did_set.py
@@ -71,16 +71,6 @@ class TestDIDSet(TestCase):
         # create a valid DID object
         tx = DIDSet(
             account=_ACCOUNT,
-            did_document=_VALID_FIELD,
-            uri=_VALID_FIELD,
-            data=_VALID_FIELD,
-        )
-        self.assertTrue(tx.is_valid())
-
-        # remove the data field from the above DID object
-        tx = DIDSet(
-            account=_ACCOUNT,
             data="",
         )
-
         self.assertTrue(tx.is_valid())
--- a/tests/unit/models/transactions/test_did_set.py
+++ b/tests/unit/models/transactions/test_did_set.py
@@ -63,11 +63,11 @@ class TestDIDSet(TestCase):
                 "{'did_set': 'Must have one of `did_document`, `data`, and `uri`.'}",
             )
 
-    def test_all_empty_fields(self):
+    def test_create_did_object_all_empty_fields(self):
         with self.assertRaises(XRPLModelException):
             DIDSet(account=_ACCOUNT, data="", did_document="", uri="")
 
-    def test_delete_multiple_fields(self):
+    def test_remove_data_field(self):
         # create a valid DID object
         tx = DIDSet(
             account=_ACCOUNT,
--- a/tests/unit/models/transactions/test_did_set.py
+++ b/tests/unit/models/transactions/test_did_set.py
@@ -62,3 +62,20 @@ class TestDIDSet(TestCase):
                 error.exception.args[0],
                 "{'did_set': 'Must have one of `did_document`, `data`, and `uri`.'}",
             )
+
+    def test_delete_data_field(self):
+        tx = DIDSet(
+            account=_ACCOUNT,
+            did_document=_VALID_FIELD,
+            uri=_VALID_FIELD,
+            data=_VALID_FIELD,
+        )
+        self.assertTrue(tx.is_valid())
+
+        # delete data field
+        tx = DIDSet(
+            account=_ACCOUNT,
+            data="",
+        )
+
+        self.assertTrue(tx.is_valid())
--- a/xrpl/models/transactions/did_set.py
+++ b/xrpl/models/transactions/did_set.py
@@ -12,8 +12,6 @@ from xrpl.models.transactions.transaction import Transaction
 from xrpl.models.transactions.types import TransactionType
 from xrpl.models.utils import require_kwargs_on_init
 
-# DID LedgerObject fields are removed with an empty string as input.
-# Hence, empty strings need to be supported in the regex
 HEX_REGEX: Final[Pattern[str]] = re.compile("[a-fA-F0-9]*")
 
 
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [[Unreleased]]
 - Add support for the DeliverMax field in Payment transactions
+- Allow empty strings for the purpose of removing fields in DIDSet transaction
```
mvadari, line 10: This should be under a `### Fixed` heading

xrpl/models/transactions/did_set.py
```diff
@@ -37,6 +39,14 @@ def _get_errors(self: DIDSet) -> Dict[str, str]:
             # Can return here because there are no fields to process
             return errors
 
+        if self.did_document == "" and self.data == "" and self.uri == "":
+            errors["did_set"] = (
+                "At least one of DID fields `did_document`, `data`, and `uri` "
```
mvadari, line 44:
```suggestion
                "At least one of the fields `did_document`, `data`, and `uri` "
```

xrpl/models/transactions/did_set.py
```diff
@@ -12,7 +12,9 @@
 from xrpl.models.transactions.types import TransactionType
 from xrpl.models.utils import require_kwargs_on_init
 
-HEX_REGEX: Final[Pattern[str]] = re.compile("[a-fA-F0-9]+")
+# DID LedgerObject fields are removed with an empty string as input.
+# Hence, empty strings need to be supported in the regex
```
mvadari, line 16: Code comments aren't useful if they're only relevant to the PR. This comment won't make any sense later on when looking at the code.

ckeshava: I can remove this comment. But I believe it's useful.

It tells future developers why the regex shouldn't support empty strings.

ckeshava: "... regex should support empty strings"

mvadari: I think that should be discussed in the docs for each of the parameters, which should be copied from the docs here: https://xrpl.org/docs/references/protocol/transactions/types/didset

CHANGELOG.md
ckeshava: fixed in [d0d5d75](https://github.com/XRPLF/xrpl-py/pull/717/commits/d0d5d753e528f5dacbd99e4bf04373d2ba6cb602)

xrpl/models/transactions/did_set.py
ckeshava: I agree. I've updated it in [e96a8ce](https://github.com/XRPLF/xrpl-py/pull/717/commits/e96a8cec419b1473086bfa76b0d996bdbb42b038)

tests/unit/models/transactions/test_did_set.py
```diff
@@ -62,3 +62,46 @@ def test_empty(self):
                 error.exception.args[0],
                 "{'did_set': 'Must have one of `did_document`, `data`, and `uri`.'}",
             )
+
+    def test_all_empty_fields(self):
+        with self.assertRaises(XRPLModelException):
+            DIDSet(account=_ACCOUNT, data="", did_document="", uri="")
+
+    def test_delete_multiple_fields(self):
```
mvadari, line 70: super nit:
```suggestion
    def test_remove_multiple_fields(self):
```
I was a bit confused at first because I thought this might be a `DIDDelete` test.

tests/unit/models/transactions/test_did_set.py
```diff
@@ -62,3 +62,46 @@ def test_empty(self):
...
+        # create a valid DID object
+        tx = DIDSet(
+            account=_ACCOUNT,
+            did_document=_VALID_FIELD,
+            uri=_VALID_FIELD,
+            data=_VALID_FIELD,
+        )
+        self.assertTrue(tx.is_valid())
```
mvadari, line 78: This is unnecessary because you're not actually sending the transaction.

tests/unit/models/transactions/test_did_set.py
```diff
@@ -62,3 +62,46 @@ def test_empty(self):
...
+        )
+
+        self.assertTrue(tx.is_valid())
+
+        # delete did_document field
+
+        # Note: It is invalid to render a DID LedgerObject empty by removing all the
+        # fields. However, that condition can only be detected by reading the
+        # LedgerEntry, which is out-of-scope for the library. rippled generated a
+        # `tecEMPTY_DID` error code in this case
```
mvadari, line 101: This is also unnecessary since the transaction is never actually submitted.

xrpl/models/transactions/did_set.py
```diff
@@ -21,8 +21,24 @@ class DIDSet(Transaction):
     """Represents a DIDSet transaction."""
 
     did_document: Optional[str] = None
+    """
+    The DID document associated with the DID.
+    You must include either Data, DIDDocument, or URI when you submit the DIDSet
+    transaction. If all three fields are missing, the transaction fails.
+
+    Note: To delete the Data, DIDDocument, or URI field from an existing DID ledger
+    entry, add the field as an empty string.
```
mvadari, line 30:
```suggestion
    The DID document associated with the DID.
    To delete this field from an existing DID ledger entry, include it as an empty string.
```

xrpl/models/transactions/did_set.py
```diff
@@ -21,8 +21,24 @@ class DIDSet(Transaction):
...
+    """
+
     data: Optional[str] = None
+    """
+    The public attestations of identity credentials associated with the DID.
```
mvadari, line 35:
```suggestion
    The public attestations of identity credentials associated with the DID.
    To delete this field from an existing DID ledger entry, include it as an empty string.
```

xrpl/models/transactions/did_set.py
```diff
@@ -21,8 +21,24 @@ class DIDSet(Transaction):
...
+    """
+
     uri: Optional[str] = None
+    """
+    The Universal Resource Identifier associated with the DID.
```
mvadari, line 40:
```suggestion
    The Universal Resource Identifier associated with the DID.
    To delete this field from an existing DID ledger entry, include it as an empty string.
```

tests/unit/models/transactions/test_did_set.py
ckeshava: are you proposing that I remove this entire unit test? Or just the `self.assertTrue(...)` statement?

tests/unit/models/transactions/test_did_set.py
ckeshava: Yeah, but I wanted to exercise the `_get_errors` function in the `DIDSet` model. Otherwise, we will probably need to do the same with an integration test

khancode (changes requested):
Add integration test for submitting `DIDSet` with empty string.

tests/unit/models/transactions/test_did_set.py
mvadari: `_get_errors` runs on model creation, not on transaction submission. I was referring to the comment, not the transaction, since it's not actually submitting the transaction.

tests/unit/models/transactions/test_did_set.py
```diff
@@ -62,3 +62,46 @@ def test_empty(self):
                 error.exception.args[0],
                 "{'did_set': 'Must have one of `did_document`, `data`, and `uri`.'}",
             )
+
+    def test_all_empty_fields(self):
+        with self.assertRaises(XRPLModelException):
+            DIDSet(account=_ACCOUNT, data="", did_document="", uri="")
+
+    def test_remove_multiple_fields(self):
```
mvadari, line 70: IMO each transaction should be a separate test. Easier to run.

ckeshava: I'd like to remove consecutive fields in the DID object, in order to demonstrate that users cannot leave behind an empty DID object.

I need to perform consecutive removals on the same object, it cannot be demonstrated with a new DID ledger object

mvadari: That's not a library test, that's a rippled test. And even if it were a library test, it should happen in an integration test, because it won't happen in a unit test (the unit test doesn't submit transactions and therefore doesn't create/modify/delete ledger objects). All you're doing here is creating the models, not submitting anything.

tests/unit/models/transactions/test_did_set.py
ckeshava: I've removed the comment and the test in [ed1c57c](https://github.com/XRPLF/xrpl-py/pull/717/commits/ed1c57c3e995b02cceb4b70a5e587dc7cfae3c52)

tests/unit/models/transactions/test_did_set.py
ckeshava: okay, I've created an integration test for the same

ckeshava:
I have reverted back to an older commit. I removed the integration test and restored the unit test. Let me know if this looks ok

tests/unit/models/transactions/test_did_set.py
```diff
@@ -62,3 +62,25 @@ def test_empty(self):
                 error.exception.args[0],
                 "{'did_set': 'Must have one of `did_document`, `data`, and `uri`.'}",
             )
+
+    def test_all_empty_fields(self):
+        with self.assertRaises(XRPLModelException):
+            DIDSet(account=_ACCOUNT, data="", did_document="", uri="")
+
+    def test_delete_multiple_fields(self):
```
khancode, line 70: I would keep the wording consistent with "empty fields" and not use "delete" since this is a unit test.

tests/unit/models/transactions/test_did_set.py
```diff
@@ -62,3 +62,25 @@ def test_empty(self):
...
+        tx = DIDSet(
+            account=_ACCOUNT,
+            did_document=_VALID_FIELD,
+            uri=_VALID_FIELD,
+            data=_VALID_FIELD,
+        )
+        self.assertTrue(tx.is_valid())
+
+        # remove the data field from the above DID object
+        tx = DIDSet(
```
mvadari, line 81: Do you still want a transaction for each field?

tests/unit/models/transactions/test_did_set.py
```diff
@@ -62,3 +62,25 @@ def test_empty(self):
...
                 "{'did_set': 'Must have one of `did_document`, `data`, and `uri`.'}",
             )
+
+    def test_all_empty_fields(self):
+        with self.assertRaises(XRPLModelException):
+            DIDSet(account=_ACCOUNT, data="", did_document="", uri="")
+
+    def test_delete_multiple_fields(self):
+        # create a valid DID object
+        tx = DIDSet(
```
mvadari, line 72: All of these transactions should be in separate tests.

tests/unit/models/transactions/test_did_set.py
ckeshava: updated in [1b7400a](https://github.com/XRPLF/xrpl-py/pull/717/commits/1b7400a0f12a732cfaf79f521cf69c3199b33c4f)

tests/unit/models/transactions/test_did_set.py
ckeshava: there are only two transactions -- create a `DID` object, and then remove the `data` field from the said object.

I cannot decouple them into separate unit tests, I need a `DID` ledger object to demonstrate the removal of the `data` field.

tests/unit/models/transactions/test_did_set.py
ckeshava: I'm only removing the `data` field. I'm not using any other transaction

tests/unit/models/transactions/test_did_set.py
mvadari: It's a unit test, the ledger object is never created in the first place.

tests/unit/models/transactions/test_did_set.py
mvadari: You originally also removed the other fields.

tests/unit/models/transactions/test_did_set.py
khancode: I would use "empty fields" rather than "remove" since this is a unit test.

tests/unit/models/transactions/test_did_set.py
ckeshava: okay, that's true.

should I remove this unit test altogether? it appears that rippled would have tested this piece of functionality.
I don't see other ways of testing this through a unit test

tests/unit/models/transactions/test_did_set.py
ckeshava: yes, that's correct. but now, I'm only empty-ing the `data` field

tests/unit/models/transactions/test_did_set.py
mvadari: Testing that one empty field works is still important.

tests/unit/models/transactions/test_did_set.py
mvadari: Yes, should there be tests for emptying each field?

ckeshava: can you tell me the name of the unit test that you are objecting to? 

`test_empty_data_field` -- a unit test that empties only the `data` field. all the other fields contain valid values.

`test_create_did_object_all_empty_fields` --  a unit test that validates the initial input for the `DIDSet` transaction. This verifies that all the input fields are not empty.


You are correct that rippled has already tested this functionality. Which one of these warrants a change?

tests/unit/models/transactions/test_did_set.py
ckeshava: okay. you mean like creating a DID object with one empty field?

```
def test_single_empty_field(self):
        # create a valid DID object
        tx = DIDSet(
            account=_ACCOUNT,
            did_document=_VALID_FIELD,
            uri="", # URI field is empty
            data=_VALID_FIELD,
        )
        self.assertTrue(tx.is_valid())
```

mvadari: Yes. Or without including the `did_document` and `data` fields.

tests/unit/models/transactions/test_did_set.py
mvadari: The transaction.

tests/unit/models/transactions/test_did_set.py
ckeshava: I gave this input to one of the integration tests:
```
tx = DIDSet(account=WALLET.address, uri="")
```

This fails with an error -- we cannot input an empty URI field, while other fields are also non-existent.
```
======================================================================
FAIL: test_all_fields (tests.integration.transactions.test_did_set.TestDIDSet.test_all_fields) (version='async', client='json')
----------------------------------------------------------------------
  File "/Users/ckeshavabs/xrpl-py/tests/integration/it_utils.py", line 290, in modified_test
    asyncio.run(
  File "/Users/ckeshavabs/.pyenv/versions/3.11.6/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/ckeshavabs/.pyenv/versions/3.11.6/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ckeshavabs/.pyenv/versions/3.11.6/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/ckeshavabs/xrpl-py/tests/integration/it_utils.py", line 281, in _run_async_test
    raise e
  File "/Users/ckeshavabs/xrpl-py/tests/integration/it_utils.py", line 277, in _run_async_test
    await test_function(self, client)
  File "/Users/ckeshavabs/xrpl-py/tests/integration/transactions/test_did_set.py", line 22, in test_all_fields
    self.assertEqual(response.result["engine_result"], "tesSUCCESS")
AssertionError: 'tecEMPTY_DID' != 'tesSUCCESS'
- tecEMPTY_DID
+ tesSUCCESS
```

mvadari: You're running it as an integration test, not a unit test. I don't need this to be tested as an integration test, I know that's going to work. That's what rippled unit tests are for. I just want to make sure that xrpl-py accepts it as a valid transaction, without knowing anything about the ledger state.

ckeshava: okay. I've updated this in 10f7f38

I am slightly opposed to this type of tests. I'd prefer to have unit tests that are in consonance with the integration tests too, that reduces ambiguity for future developers

mvadari: > that reduces ambiguity for future developers

What do you mean by this?

The way I think about it is that unit tests are checking xrpl-py Python code and integration tests are ensuring that xrpl-py Python code matches rippled code. So for checking xrpl-py error code, you only want to write unit tests.

ckeshava: that is true.

When looking at the unit tests, I'm led to believe that a DID object can be created with empty fields. But that is not the case with a real rippled instance.

I feel this contradiction could have been avoided, but it's not a strong objection

mvadari: I would assume anyone advanced enough to be looking at unit tests to understand the system is probably well-versed enough in the XRPL that they're not going to fall into that kind of trap - or will at least know that unit tests aren't being submitted to the ledger.

ckeshava: fair point

mvadari (approved):

khancode (approved):

PR #720, opened 2024-07-01
feat: add rippled API v2 support and use as default

ckeshava:

All the requests are updated to use the newer rippled v2 API responses

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Existing integration tests have been modified to use the newer API version responses.


xrpl/models/requests/generic_request.py
```diff
@@ -62,7 +63,16 @@ def from_dict(cls: Type[GenericRequest], value: Dict[str, Any]) -> GenericReques
 
         elif "method" in value:  # JSON RPC formatting
             if "params" in value:  # actual JSON RPC formatting
-                value = {"method": value["method"], **value["params"]}
+                if "api_version" in value:
```
mvadari, line 66: Why does the API version need to be specified in generic requests specifically?

xrpl/models/response.py
```diff
@@ -3,19 +3,24 @@
...
 
+from typing_extensions import Final
+
 from xrpl.models.base_model import BaseModel
 from xrpl.models.required import REQUIRED
 from xrpl.models.transactions import PaymentFlag
 from xrpl.models.transactions.types import TransactionType
 from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
 
+DEFAULT_API_VERSION: Final[int] = 2
```
mvadari, line 22: Why is this defined in this file, where it's not used, instead of `request.py`?

.ci-config/rippled.cfg
```diff
@@ -178,3 +178,7 @@ PriceOracle
 fixEmptyDID
 fixXChainRewardRounding
 fixPreviousTxnID
+
+# Use a specific network_id for this ripled instance. Otherwise, network_id is read as None
+[network_id]
+63456
```
khancode, line 184:
```suggestion
# network_id is required otherwise it's read as None.
[network_id]
63456
```

Can you also move this above `[features]`?

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [[Unreleased]]
 - Add support for the DeliverMax field in Payment transactions
 
+### Added
+- Update all requests to use rippled API v2 responses
```
khancode, line 12:
```suggestion
- Use rippled API v2 as default in requests
```

xrpl/models/requests/generic_request.py
ckeshava: Certain requests like `tx_history` are implemented through `GenericRequest`. But this request ahs been deprecated since api version 2. This PR intends to make API v2 as the default response in xrpl-py.

An `api_version < 2` needs to be specified for certain generic-requests, otherwise they fail with an `invalid_API_version` error.

Look at `test_generic_request` for an example

xrpl/models/response.py
ckeshava: I was looking for a global constants file, but I couldn't find it.
I've updated the location to requests.py file 👍

xrpl/models/requests/generic_request.py
mvadari: Seems fine to leave that as existing behavior then - default to v2. For example, `GenericRequest` is also used for `ledger_accept`, which doesn't differ between V1 and V2.

ckeshava: Yes, but `api_version: 2` is specified in all requests, by default (including `ledger_accept`).

Without this change, `tx_history` requests are transmitted with `api_version: 2` and they fail. Since we need to support API v1, how do you propose to handle `tx_history` requests?

mvadari: It should error if `api_version` isn't included, since the default is `2`. If `api_version: 1` is included, it will work. I assume that's what would still happen.

Note: we don't have to specifically support `tx_history`, that was just something to use for the test since the RPC doesn't have a model.

ckeshava: > It should error if `api_version` isn't included, since the default is `2`. If `api_version: 1` is included, it will work. I assume that's what would still happen.
> 

No, `api_version: 2` is always included in every request, because `Request` is the parent class of every Request. By default, this is the Request JSON:

```
{
    "method": "tx_history",
    "params": {
        "start": 0,
    },
    "api_version": 2, // we need to change this value into 1
}
```

ckeshava: > Note: we don't have to specifically support `tx_history`, that was just something to use for the test since the RPC doesn't have a model.

I think `tx_history` is a good example because it does not work in rippled API v2 (or later versions). This change "forwards" the `api_version` as specified by the user.

khancode: Agree with @mvadari - it's ok for `tx_history` fail now that we default to v2.

mvadari: > No, `api_version: 2` is always included in every request, because `Request` is the parent class of every Request.

Yes, I understand. I would expect this to fail:
```python
GenericRequest(method="tx_history", start=0) # api_version defaults to 2
```

But this to succeed:
```python
GenericRequest(method="tx_history", start=0, api_version=1)
```

ckeshava: > But this to succeed:
> 
> ```python
> GenericRequest(method="tx_history", start=0, api_version=1)
> ```
> 

This input will not succeed, because we are over-writing the `value` variable, hence we lose the `api_version` information even if it specified.

In the below line, `value` variable is overwritten as:
```
value = {
              # "api_version": value["api_version"],
              "method": value["method"],
              **value["params"],
          }
```






> I disagree that this behavior means that any user of `GenericRequest` needs to specify an API version, since most APIs did not change from API v1 to v2.

I'm not forcing users to specify any version at all. I'm preserving the `api_version` information from being lost.

mvadari: You don't need the API check for that. I tried running just this code in the test and it worked:
```
if "params" in value:  # actual JSON RPC formatting
  value = {
    "api_version": value["api_version"],
    "method": value["method"],
    **value["params"],
  }
```

ckeshava: what if a user submits a raw JSON input, but forgets to specify the `api_version`?

Since the user is not using the `GenericRequest/Request` class, the `api_version` field will need to be specified manually

mvadari: The user can't submit raw JSON input in xrpl-py, you have to go through a model.

Even if you could, it would then default to whatever the rippled instance has as default (probably v1).

ckeshava: I added a test case at the end of this file: `tests/integration/reqs/test_generic_request.py`

Users can submit raw-JSON inputs through the `from_dict` methods. This request does not reach rippled due to processing errors in the `from_dict` function.

We will need to document that `api_version` is a mandatory field, if raw-JSON input is provided. If `GenericRequest` model is used, then the default `api_version` is provided automatically.

mvadari: It's not a mandatory field if a default option is provided. The same is true of any other API method - if you don't specify an `api_version` field, you'll get the default version.

tests/integration/sugar/test_network_id.py
```diff
@@ -1,24 +0,0 @@
...
-from xrpl.asyncio.transaction.main import _RESTRICTED_NETWORKS
-from xrpl.clients import WebsocketClient
-from xrpl.models.transactions import AccountSet
-from xrpl.transaction import autofill
-
-_FEE = "0.00001"
-
-
-# TODO: move to test_transaction and use the standard integration test setup
-class TestNetworkID(TestCase):
```
mvadari, line 12: Why is this file removed?

mvadari:
This PR should be merged into a different branch, called something like `3.0`, since this is going to be in a major version bump. That allows us to still merge/deploy other PRs in minor/patch bumps without waiting to be ready for a major version bump.

tests/integration/sugar/test_network_id.py
ckeshava: I have moved this test into `test_transaction.py` file

khancode (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Add support for the DeliverMax field in Payment transactions
+- Use rippled API v2 as default in requests
```
pdp2121, line 12: I think this should be at the top (or highlighted some how) since this is a major change. Other than that, LGTM.

pdp2121 (approved):

.ci-config/rippled.cfg
```diff
@@ -105,6 +105,10 @@ validators.txt
 # Note: The version of rippled you use this config with must have an implementation for the amendments you attempt to enable or it will crash.
 # If you need the version of rippled to be more up to date, you may need to make a comment on this repo: https://github.com/WietseWind/docker-rippled
 
+# network_id is required otherwise it's read as None
+[network_id]
+63456
```
mvadari, line 110: Why do we need a network ID?

snippets/multisign.py
```diff
@@ -57,10 +58,10 @@
 if multisigned_tx_response.result["validated"]:
     print("The multisigned transaction was accepted by the ledger:")
     print(multisigned_tx_response)
-    if multisigned_tx_response.result["Signers"]:
+    if multisigned_tx_response.result["tx_json"]["Signers"]:
```
mvadari, line 61: nit: this would be easier to read with a helper variable for `multisigned_tx_response.result["tx_json"]["Signers"]`

tests/integration/reqs/test_generic_request.py
```diff
@@ -32,10 +31,19 @@ async def test_json_formatting(self, client):
...
-                    "start": 0,
-                }
+                {"command": "tx_history", "start": 0, "api_version": 1}
             )
         )
         self.assertTrue(response.is_successful())
+
+    def test_from_dict_json_without_api_version_input(self):
+        with self.assertRaises(KeyError):
+            # input JSON does not contain an api_version field
```
mvadari, line 41:
```suggestion
            # tx_history is invalid in the default API version 2
```

tests/integration/sugar/test_network_id.py
mvadari: We're no longer testing both scenarios (network ID < 1024 vs network ID >= 1024), then.

xrpl/models/requests/request.py
```diff
@@ -9,14 +9,16 @@
...
-from typing_extensions import Self
+from typing_extensions import Final, Self
 
 import xrpl.models.requests  # bare import to get around circular dependency
 from xrpl.models.base_model import BaseModel
 from xrpl.models.exceptions import XRPLModelException
 from xrpl.models.required import REQUIRED
 from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
 
+DEFAULT_API_VERSION: Final[int] = 2
```
mvadari, line 20:
```suggestion
_DEFAULT_API_VERSION: Final[int] = 2
```
It's a helper variable that's only used here (other than in tests).

tests/integration/reqs/test_generic_request.py
```diff
@@ -4,13 +4,11 @@
...
 class TestGenericRequest(IntegrationTestCase):
+    # Note: Support for the tx_history command has been removed since rippled API v2
     @test_async_and_sync(globals())
     async def test_constructor(self, client):
         response = await client.request(
-            GenericRequest(
-                method="tx_history",
-                start=0,
-            )
+            GenericRequest(method="tx_history", start=0, api_version=1)
```
mvadari, line 11: There should be an additional test for another request that makes sure that the API version is handled properly - perhaps `account_info` since there is a significant difference in the shape of the output.

xrpl/models/requests/request.py
```diff
@@ -106,6 +108,13 @@ class Request(BaseModel):
 
     id: Optional[Union[str, int]] = None
 
+    api_version: int = DEFAULT_API_VERSION
```
mvadari, line 111: There should be an additional test for another request that makes sure that the `api_version` field is handled properly - perhaps `account_info` since there is a significant difference in the shape of the output.

pdp2121: Agree with this

.ci-config/rippled.cfg
ckeshava: This is related to the integration test: `test_autofill_populate_networkid`. Originally, this test was using the hooks-testnet for sending transactions. But it was decided that client libraries are not supporting (or) using hooks test-nets.

The docker container for the integration test uses this config file for startup.


The autofill method's [control flow](https://github.com/XRPLF/xrpl-py/blob/70ed9c034ad32aed1c105e042690c420a00a091e/xrpl/asyncio/transaction/main.py#L235)  is contingent on `network_id > 1025` for this particular test.

This change is not directly related to the API v2 updates

mvadari: > But it was decided that client libraries are not supporting (or) using hooks test-nets.

We can still use it for a test.

> This change is not directly related to the API v2 updates

Why is this change included in this PR, then?

xrpl/models/requests/request.py
mvadari: There should be an integration test that ensures that including an `api_version` param in a request type that is not `GenericRequest` still works correctly and properly returns the correct API version.

xrpl/models/requests/request.py
ckeshava: fixed in 4179077

snippets/multisign.py
ckeshava: fixed in 4179077

CHANGELOG.md
ckeshava: fixed in 4179077

.ci-config/rippled.cfg
ckeshava: i believe the integration tests are better off without the dependency on hooks-test-network -- this particular function can be tested without making network calls to the hooks-network

I can remove this change from this PR

mvadari: It'll be easier if that discussion can happen separately.

xrpl/models/requests/request.py
ckeshava: Doesn't a unit test suffice?

It is rippled's responsibility to return the correct `api_version` response, why should the client libraries test the response?

Even in the case of `GenericRequest`, I'm only forwarding the `api_version` input specified by the user. The integration test pertaining to `tx_history` tests the older versions of rippled. I have not tested the validity of responses of rippled

tests/integration/sugar/test_network_id.py
ckeshava: For the cases where `network_ID <= 1024`, we don't need to populate `network_id` field inside transactions.

Hence, we don't need to test this function for those networks

xrpl/models/requests/request.py
mvadari: > Doesn't a unit test suffice?

In this case, I'd like to ensure that it matches rippled compatibility, so an integration test is needed.

> It is rippled's responsibility to return the correct `api_version` response, why should the client libraries test the response?

I'm not suggesting a test that actually fully tests the rippled response, but a test that ensures that when I put in `api_version: 1` in my request model, it actually gives me the rippled API v1 response and not the rippled API v2 response (in which case there's something wrong with the implementation of `api_version` in the xrpl-py model).

tests/integration/sugar/test_network_id.py
mvadari: It will error if `NetworkID` _is_ specified, in those cases. So that scenario does still need to be tested.

.ci-config/rippled.cfg
ckeshava: I have reverted the changes to the `main` branch. But it appears that the hooks testnet is using `rippledv1.10` : https://github.com/XRPL-Labs/xrpld-hooks/blob/b13e7db18d1376d40cdbeeea849702b568324279/src/ripple/protocol/impl/BuildInfo.cpp#L36

Here is the Request and Response from hooks testnet validator
```
# request
ServerInfo(method=<RequestMethod.SERVER_INFO: 'server_info'>, id=None, api_version=2)

# response
Response(status=<ResponseStatus.ERROR: 'error'>, result={'api_version': 2, 'error': 'invalid_API_version', 'id': 'RequestMethod.SERVER_INFO_174181', 'request': {'api_version': 2, 'command': 'server_info', 'id': 'RequestMethod.SERVER_INFO_174181'}, 'status': 'error', 'type': 'response'}, id='RequestMethod.SERVER_INFO_174181', type=<ResponseType.RESPONSE: 'response'>)

```

ckeshava: I'll try to find a fix tomorrow

mvadari: Make a quick PR to fix those issues first?

ckeshava: For the validators using v1 API, we will need to specify `api_version: 1` in *all* the requests. These requests are not localized to a particular file (or) a class.

I need to set a global switch (for `Request.api_version`) based on the API version used by the client. Let me try to see this works

mvadari: I don't think that's necessary, and I'm not really sure how that'd work anyways. People using an older version of rippled that doesn't support APIv2 likely don't need to upgrade to a newer version of the library anyways.

xrpl/models/requests/request.py
ckeshava: In this PR, every `Request` will always specify an `api_version` field. It is mandatory for all the requests. 

If a `Request` omits a `api_version` field, then rippled returns v1 Response -- [docs](https://xrpl.org/docs/references/http-websocket-apis/api-conventions/request-formatting/#websocket-format)

Having said that, I can manually construct a JSON and validate API version of the `Response` object. There would be three cases:
- Request specifying  `api_version : 1`
- Request specifying  `api_version : 2`
- Request not specifying  `api_version` (through a dictionary object)

.ci-config/rippled.cfg
mvadari: > I didn't understand your second sentence, you're saying that there will always be legacy networks (pre `APIv2`), but how do we deal with them?

Right, but they don't benefit from upgrading to newer versions of the libraries. They can continue to use older versions.

xrpl/models/requests/request.py
mvadari: Don't manually construct a JSON, that's not useful for testing the library. Construct a request model through the proper process.

ckeshava: added tests in ae5b24f

.ci-config/rippled.cfg
ckeshava: I see -- can we use another testnet in their stead? We will need a candidate with `network_id > 1024` (Now, I'm able to place your offline question in light of this discussion 👍 )

ckeshava: alternatively, we could revert to specifying a network_id inside the config file for the integration tests

tests/integration/sugar/test_network_id.py
ckeshava: I understand, but that falls under the purview of rippled. I agree that we can write integration tests to that effect, but it introduces tech responsibility due to future updates to rippled codebase

.ci-config/rippled.cfg
mvadari: We need to test two things:
* network ID > 1024
* network ID <= 1024

Regardless of what the standalone config says, we'll only be testing one of those.

ckeshava: Why do we need to test `networkID <= 1024` ? The test expects a `RippledError` object stating that `NetworkID` is not required. But should xrpl-py validate all possible RippledErrors? I don't understand the separation of responsibilities between rippled and client libraries.

But, if we need to test both these cases through xrpl-py, then:
1. We need to set up test networks with network_id > 1024 -- This is useful for testing the interfaces of other client libraries too.

2. Use two docker containers with two different network_id's -- this covers both the possibilities. I need to figure out a clean port-mapping of the docker containers to disambiguate the two cases

mvadari: `networkID <= 1024` _must not_ have the`NetworkID` included in the transaction. `networkID > 1024` _must_ have the `NetworkID` included in the transaction. We need to ensure that xrpl-py behaves accordingly. This was what we had the hooks testnet testcase for - to ensure the latter case.

Neither of the scenarios you suggest are really worth the effort just to test one edgecase. The better option is probably to change the standalone config to have a network ID > 1024 and then add a network ID test for one of the testnets that have a network ID < 1024.

ckeshava: okay, I've reverted back one of my older commits, and I also added a new test case
[8632aef](https://github.com/XRPLF/xrpl-py/pull/720/commits/8632aef2f8fb981353d5f5ce2df187a415130f61)

tests/integration/reqs/test_account_tx.py
```diff
@@ -13,3 +13,43 @@ async def test_basic_functionality(self, client):
...
+
+        # API v1 returns a tx field
+        self.assertTrue("tx" in response.result["transactions"][0])
+        self.assertTrue(response.is_successful())
+
+    @test_async_and_sync(globals())
+    async def test_validate_response_type_no_explicit_api_version(self, client):
+        response_v1 = await client.request(AccountTx(account=WALLET.address))
+
+        response_v2 = await client.request(
```
mvadari, line 49: The variable names here are a bit confusing because I thought they were going to be APIv1 and APIv2.

tests/integration/reqs/test_account_tx.py
```diff
@@ -13,3 +13,43 @@ async def test_basic_functionality(self, client):
...
+        # use the below proxies to ensure that the correct API response is returned
+        # API v1 does not contain a tx_json field
+        self.assertTrue("tx_json" not in response.result["transactions"][0])
+
+        # API v1 returns a tx field
+        self.assertTrue("tx" in response.result["transactions"][0])
+        self.assertTrue(response.is_successful())
+
+    @test_async_and_sync(globals())
+    async def test_validate_response_type_no_explicit_api_version(self, client):
```
mvadari, line 46: Can the test names be a bit less verbose? e.g. 
```suggestion
    async def test_no_explicit_api_version(self, client):
```

tests/integration/sugar/test_transaction.py
```diff
@@ -216,6 +216,37 @@ async def test_calculate_payment_fee(self, client):
         expected_fee = await get_fee(client)
         self.assertEqual(payment_autofilled.fee, expected_fee)
 
+    @test_async_and_sync(
+        globals(),
+        ["xrpl.transaction.autofill"],
+    )
+    # Autofill should populate the tx networkID and build_version from 1.11.0 or later.
+    # NetworkID field is populated only for networks where network_id > 1024
```
mvadari, line 224: These comments should be inside the tests, not above them

tests/integration/sugar/test_transaction.py
```diff
@@ -216,6 +216,37 @@ async def test_calculate_payment_fee(self, client):
...
+        tx = AccountSet(
+            account="rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh",
+            fee=FEE,
+            domain="www.example.com",
+        )
+        transaction = await autofill(tx, client)
+
+        # Although the client network_id property is set,
+        # the corresponding field in transaction is not populated
+        self.assertTrue(transaction.network_id is None)
```
mvadari, line 247:
```suggestion
        self.assertIsNone(transaction.network_id)
```

tests/integration/reqs/test_account_tx.py
```diff
@@ -13,3 +13,43 @@ async def test_basic_functionality(self, client):
...
+
+    @test_async_and_sync(globals())
+    async def test_validate_response_type_v2(self, client):
+        response = await client.request(
+            AccountTx(account=WALLET.address, api_version=2)
+        )
+
+        # use the below proxies to ensure that the correct API response is returned
+        # API v2 returns tx_json field
+        self.assertTrue("tx_json" in response.result["transactions"][0])
```
mvadari, line 25:
```suggestion
        self.assertIn("tx_json", response.result["transactions"][0])
```
Ditto for all the other tests here

snippets/multisign.py
```diff
@@ -57,11 +58,18 @@
...
+        print("ERROR: Response does not contain a Signers field")
+    elif len(signers_in_response) != len(signer_entries):
         print(
-            "The transaction had "
-            f"{len(multisigned_tx_response.result['Signers'])} signatures"
+            "ERROR: Response does not contain "
+            + str(len(signer_entries))
+            + " Signers fields"
         )
+    else:
```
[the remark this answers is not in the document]
ckeshava: I feel these capture all the corner cases too -- whether signers field is present in the response, number of signers, etc. But yeah, this appears to be a well-tested feature that has persisted for many years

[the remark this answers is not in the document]
ckeshava: removed these error messages in 394d1fd

tests/integration/reqs/test_account_tx.py
ckeshava: changed the variable names in 394d1fd

tests/integration/reqs/test_account_tx.py
ckeshava: fixed in 394d1fd

tests/integration/sugar/test_transaction.py
ckeshava: fixed in 394d1fd

tests/integration/reqs/test_account_tx.py
ckeshava: fixed in 394d1fd

snippets/multisign.py
```diff
@@ -57,11 +58,9 @@
...
     print("The multisigned transaction was accepted by the ledger:")
     print(multisigned_tx_response)
-    if multisigned_tx_response.result["Signers"]:
-        print(
-            "The transaction had "
-            f"{len(multisigned_tx_response.result['Signers'])} signatures"
-        )
+    signers_in_response = multisigned_tx_response.result["tx_json"]["Signers"]
+
+    print("The transaction had " f"{len(signers_in_response)} signatures")
```
mvadari, line 63: You still need the `if signers_in_response` that existed previously in the `if`

tests/integration/sugar/test_transaction.py
```diff
@@ -216,6 +216,39 @@ async def test_calculate_payment_fee(self, client):
         expected_fee = await get_fee(client)
         self.assertEqual(payment_autofilled.fee, expected_fee)
 
+    @test_async_and_sync(
+        globals(),
+        ["xrpl.transaction.autofill"],
+    )
+    async def test_autofill_populate_networkid_non_reserved_networks(self, client):
```
mvadari, line 223:
```suggestion
    async def test_networkid_non_reserved_networks(self, client):
```

tests/integration/sugar/test_transaction.py
```diff
@@ -216,6 +216,39 @@ async def test_calculate_payment_fee(self, client):
...
+        )
+        transaction = await autofill(tx, client)
+
+        # Autofill should populate the tx networkID and build_version from 1.11.0 or
+        # later. NetworkID field is populated only for networks where network_id > 1024
+        self.assertEqual(client.network_id, 63456)
+        self.assertEqual(transaction.network_id, 63456)
+
+    @test_async_and_sync(globals(), ["xrpl.transaction.autofill"], use_testnet=True)
+    async def test_autofill_populate_networkid_reserved_networks(self, client):
```
mvadari, line 237:
```suggestion
    async def test_networkid_reserved_networks(self, client):
```

snippets/multisign.py
ckeshava: thanks, fixed in [376d823](https://github.com/XRPLF/xrpl-py/pull/720/commits/376d823fc7e88f14b2d8a7f398a71fbaa6bfc8c1)

snippets/multisign.py
```diff
@@ -57,11 +58,10 @@
 if multisigned_tx_response.result["validated"]:
     print("The multisigned transaction was accepted by the ledger:")
     print(multisigned_tx_response)
-    if multisigned_tx_response.result["Signers"]:
-        print(
-            "The transaction had "
-            f"{len(multisigned_tx_response.result['Signers'])} signatures"
-        )
+    signers_in_response = multisigned_tx_response.result["tx_json"]["Signers"]
```
mvadari, line 61:
```suggestion
    signers_in_response = multisigned_tx_response.result["tx_json"].get("Signers")
```
Otherwise, the code will error instead of hitting the if.

mvadari (approved):

PR #724, opened 2024-07-09
Remove deprecated `accounts` and `type` parameters from ledger request model

kennyzlei:

Remove deprecated `accounts` and `type` parameters from ledger request model

### Context of Change

In the [ledger API](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/ledger-methods/ledger/), the accounts and type request parameters has been deprecated for quite awhile now. As of Clio 2.2.0 with this particular commit https://github.com/XRPLF/clio/pull/1360, this API will return a request error if you provide that parameter. This change removes this parameter from the ledger request model. This change would not impact lib users as this setting this parameter previously would not have been used by rippled or Clio

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Existing integration tests should continue to pass and manual testing of ledger API with xrpl-py against Clio 2.2.0 should succeed

### DIFF
--- a/xrpl/models/requests/ledger.py
+++ b/xrpl/models/requests/ledger.py
@@ -3,9 +3,7 @@ Retrieve information about the public ledger.
 `See ledger <https://xrpl.org/ledger.html>`_
 """
 from dataclasses import dataclass, field
-from typing import Optional
 
-from xrpl.models.requests.ledger_entry import LedgerEntryType
 from xrpl.models.requests.request import LookupByLedgerRequest, Request, RequestMethod
 from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
 
--- a/xrpl/models/requests/ledger.py
+++ b/xrpl/models/requests/ledger.py
@@ -19,10 +19,8 @@ class Ledger(Request, LookupByLedgerRequest):
     """
 
     method: RequestMethod = field(default=RequestMethod.LEDGER, init=False)
-    accounts: bool = False
     transactions: bool = False
     expand: bool = False
     owner_funds: bool = False
     binary: bool = False
     queue: bool = False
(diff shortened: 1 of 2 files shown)


mvadari (approved):

xrpl/models/requests/ledger.py
```diff
@@ -19,10 +17,8 @@ class Ledger(Request, LookupByLedgerRequest):
     """
 
     method: RequestMethod = field(default=RequestMethod.LEDGER, init=False)
-    accounts: bool = False
     transactions: bool = False
     expand: bool = False
     owner_funds: bool = False
     binary: bool = False
     queue: bool = False
```
ckeshava, line 24: This change looks good to me.

But, the current model does not represent `ledger_hash`  or `ledger_index` fields in the `Ledger` request. I think we can include these changes in this PR (since there are no methods in this `Ledger` method)

Alternatively, I can note this down and work on it later too.

ckeshava (approved):

xrpl/models/requests/ledger.py
kennyzlei: I think I would like to we keep the scope of this PR to removing deprecated fields.  If we can make those changes in a separate PR that would be greatly appreciated 🙏

kennyzlei:
@mvadari @ckeshava if either of you have merge perms, would be greatly appreciated if you can merge when convenient 🙏

xrpl/models/requests/ledger.py
mvadari: > But, the current model does not represent `ledger_hash`  or `ledger_index` fields in the `Ledger` request.

It's included via inheritance - see `LookupByLedgerRequest`.

PR #737, opened 2024-08-14
fix: remove Python 3.7 support to fix dependency installation and use 3.8 as new default

khancode:

Project can no longer build dependencies using Python 3.7. This PR removes Python 3.7 support to fix dependency installation and uses 3.8 as new default.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,9 +59,6 @@ MarkupSafe = "2.1.5"
 Sphinx = "^5.3.0"
 poethepoet = "^0.19.0"
 
-[tool.poetry.group.dev.dependencies]
-packaging = "^24.1"
-
 [tool.isort]
 # Make sure that isort's settings line up with black
 profile = "black"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.7"
-zipp = "3.14.0"
 base58 = "^2.1.0"
 ECPy = "^1.2.5"
 typing-extensions = "^4.2.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.7"
+zipp = "3.14.0"
 base58 = "^2.1.0"
 ECPy = "^1.2.5"
 typing-extensions = "^4.2.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.7"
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -59,6 +59,9 @@ MarkupSafe = "2.1.5"
 Sphinx = "^5.3.0"
 poethepoet = "^0.19.0"
 
+[tool.poetry.group.dev.dependencies]
+packaging = "^24.1"
```
ckeshava, line 63: where is the `packaging` module used? why do we need a new dev-dependency ?

xrpl/models/base_model.py
```diff
@@ -177,7 +177,7 @@ def _from_dict_single_param(
 
         # no more collections (no params expect a Dict)
 
-        if param_type is Any:
+        if param_type is Any:  # type: ignore
```
ckeshava, line 180: Is this type annotation necessary?
I updated this line (without the type annotation comment) -- it appears to work correctly

https://github.com/XRPLF/xrpl-py/compare/test-branch-builds...ckeshava:xrpl-py:py38-updates?expand=1

pyproject.toml
ckeshava: here is a commit without this dependency. feel free to pick it up: https://github.com/XRPLF/xrpl-py/compare/test-branch-builds...ckeshava:xrpl-py:py38-updates?expand=1

ckeshava: I couldn't find the usages of `packaging` module outside of the poetry.lock file

khancode: This can be removed. It was automatically added when I updated the Python version.

xrpl/models/base_model.py
khancode: It's needed to ignore a new error generated by mypy even though it's working code:

```
poetry run mypy --strict --implicit-reexport xrpl snippets
xrpl/models/base_model.py:180: error: Non-overlapping identity check (left operand type: "Type[Any]", right operand type: "<typing special form>")  [comparison-overlap]
Found 1 error in 1 file (checked 232 source files)
Error: Process completed with exit code 1.
```

ckeshava (approved):
this looks good to me.

there is one more occurence of Python 3.7 in an .rst file (for reference: my custom branch changes it to Python 3.8). That needs to be corrected in your branch

ckeshava:
In order to merge this branch, you might have to remove the Branch Protection Rules -- regarding the Python 3.7 CI tests

khancode:
> In order to merge this branch, you might have to remove the Branch Protection Rules -- regarding the Python 3.7 CI tests

I updated the rules to remove them, thanks!

pdp2121 (approved):

PR #757, opened 2024-10-10
feat: add support for Batch amendment

mvadari:

This PR:
* Adds models for the `Batch` transactions
* Updates definitions.json to handle the new field types
* Adds support to the binary codec to properly sign multi-account `Batch` transactions
* Adds helper functions for signing and combining multi-account `Batch` transactions
* Adds support to `autofill` for filling in the `BatchTxn` and `TxIDs` fields
* Adds support for Batchnet to the faucets

### Context of Change

https://github.com/XRPLF/rippled/pull/5060

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### Did you update HISTORY.md?

- [x] Yes

## Test Plan

Added tests for the new features.

### DIFF
--- a/tests/integration/sugar/test_transaction.py
+++ b/tests/integration/sugar/test_transaction.py
@@ -6,22 +6,30 @@ from tests.integration.it_utils import (
 )
 from tests.integration.reusable_values import DESTINATION as DESTINATION_WALLET
 from tests.integration.reusable_values import WALLET
+from xrpl.asyncio.account import get_next_valid_seq_number
 from xrpl.asyncio.ledger import get_fee, get_latest_validated_ledger_sequence
 from xrpl.asyncio.transaction import (
     XRPLReliableSubmissionException,
     autofill,
     autofill_and_sign,
     sign,
+    sign_and_submit,
 )
 from xrpl.asyncio.transaction import submit as submit_transaction_alias_async
 from xrpl.asyncio.transaction import submit_and_wait
-from xrpl.asyncio.transaction.main import sign_and_submit
 from xrpl.clients import XRPLRequestFailureException
 from xrpl.core.addresscodec import classic_address_to_xaddress
-from xrpl.core.binarycodec.main import encode
+from xrpl.core.binarycodec import encode
 from xrpl.models.exceptions import XRPLException
 from xrpl.models.requests import ServerState, Tx
-from xrpl.models.transactions import AccountDelete, AccountSet, EscrowFinish, Payment
+from xrpl.models.transactions import (
+    AccountDelete,
+    AccountSet,
+    Batch,
+    DepositPreauth,
+    EscrowFinish,
+    Payment,
+)
 from xrpl.utils import xrp_to_drops
 
 ACCOUNT = WALLET.address
@@ -248,6 +256,38 @@ class TestTransaction(IntegrationTestCase):
         # the corresponding field in transaction is not populated
         self.assertIsNone(transaction.network_id)
         self.assertEqual(client.network_id, 1)
+
+    @test_async_and_sync(
+        globals(),
+        ["xrpl.transaction.autofill", "xrpl.account.get_next_valid_seq_number"],
+    )
+    async def test_batch_autofill(self, client):
+        tx = Batch(
+            account="rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh",
+            raw_transactions=[
+                DepositPreauth(
+                    account=WALLET.address,
+                    authorize="rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh",
+                ),
+                DepositPreauth(
+                    account=WALLET.address,
+                    authorize="rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn",
+                ),
+            ],
+        )
+        transaction = await autofill(tx, client)
+        self.assertEqual(len(transaction.tx_ids), 2)
+
+        sequence = await get_next_valid_seq_number(WALLET.address, client)
+        for i in range(len(transaction.raw_transactions)):
+            raw_tx = transaction.raw_transactions[i]
+            self.assertIsNotNone(raw_tx.batch_txn)
+            self.assertEqual(
+                raw_tx.batch_txn.outer_account, "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh"
+            )
+            self.assertEqual(raw_tx.batch_txn.sequence, sequence + i)
+            self.assertEqual(raw_tx.batch_txn.batch_index, i)
+            self.assertEqual(raw_tx.get_hash(), transaction.tx_ids[i])
 
 
 class TestSubmitAndWait(IntegrationTestCase):
--- a/xrpl/asyncio/transaction/main.py
+++ b/xrpl/asyncio/transaction/main.py
@@ -3,7 +3,7 @@
 import math
 from typing import Any, Dict, List, Optional, Tuple, cast
 
-from typing_extensions import Final
+from typing_extensions import Final, TypeVar
 
 from xrpl.asyncio.account import get_next_valid_seq_number
 from xrpl.asyncio.clients import Client, XRPLRequestFailureException
@@ -15,7 +15,7 @@ from xrpl.core.keypairs.main import sign as keypairs_sign
 from xrpl.models.requests import ServerInfo, ServerState, SubmitOnly
 from xrpl.models.response import Response
 from xrpl.models.transactions import EscrowFinish
-from xrpl.models.transactions.batch import Batch, BatchInnerTransaction
+from xrpl.models.transactions.batch import Batch
 from xrpl.models.transactions.transaction import Signer, Transaction
 from xrpl.models.transactions.transaction import (
     transaction_json_to_binary_codec_form as model_transaction_to_binary_codec,
@@ -214,9 +214,12 @@ def _prepare_transaction(
     return transaction_json
 
 
+T = TypeVar("T", bound=Transaction)
+
+
 async def autofill(
-    transaction: Transaction, client: Client, signers_count: Optional[int] = None
-) -> Transaction:
+    transaction: T, client: Client, signers_count: Optional[int] = None
+) -> T:
     """
     Autofills fields in a transaction. This will set `sequence`, `fee`, and
     `last_ledger_sequence` according to the current state of the server this Client is
@@ -251,7 +254,7 @@ async def autofill(
         transaction_json["raw_transactions"] = inner_txs
         if "tx_ids" not in transaction_json:
             transaction_json["tx_ids"] = tx_ids
-    return Transaction.from_dict(transaction_json)
+    return cast(T, Transaction.from_dict(transaction_json))
 
 
 async def _get_network_id_and_build_version(client: Client) -> None:
@@ -493,18 +496,18 @@ async def _fetch_owner_reserve_fee(client: Client) -> int:
 
 async def _autofill_batch(
     client: Client, transaction: Batch
-) -> Tuple[List[BatchInnerTransaction], List[str]]:
+) -> Tuple[List[Transaction], List[str]]:
     account_sequences: Dict[str, int] = {}
     batch_index = 0
     tx_ids: List[str] = []
-    inner_txs: List[BatchInnerTransaction] = []
+    inner_txs: List[Transaction] = []
 
     for raw_txn in transaction.raw_transactions:
-        if raw_txn.BatchTxn is not None:
+        if raw_txn.batch_txn is not None:
             inner_txs.append(raw_txn)
             continue
 
-        batch_txn = {"outer_account": transaction.account}
+        batch_txn: Dict[str, Any] = {"outer_account": transaction.account}
 
         if raw_txn.account in account_sequences:
             batch_txn["sequence"] = account_sequences[raw_txn.account]
@@ -518,8 +521,9 @@ async def _autofill_batch(
         batch_index += 1
 
         raw_txn_dict = raw_txn.to_dict()
-        raw_txn_dict["RawTransaction"]["BatchTxn"] = batch_txn
-        inner_txs.append(BatchInnerTransaction.from_dict(raw_txn_dict))
-        tx_ids.append(raw_txn.get_hash())
+        raw_txn_dict["batch_txn"] = batch_txn
+        new_raw_txn = Transaction.from_dict(raw_txn_dict)
+        inner_txs.append(new_raw_txn)
+        tx_ids.append(new_raw_txn.get_hash())
 
     return inner_txs, tx_ids
--- a/xrpl/asyncio/transaction/main.py
+++ b/xrpl/asyncio/transaction/main.py
@@ -87,11 +87,12 @@ def sign(
     Returns:
         The signed transaction blob.
     """
+    transaction_json = _prepare_transaction(transaction, wallet)
     if multisign:
         signature = keypairs_sign(
             bytes.fromhex(
                 encode_for_multisigning(
-                    transaction.to_xrpl(),
+                    transaction_json,
                     wallet.address,
                 )
             ),
@@ -107,7 +108,6 @@ def sign(
         ]
         return Transaction.from_dict(tx_dict)
 
-    transaction_json = _prepare_transaction(transaction, wallet)
     serialized_for_signing = encode_for_signing(transaction_json)
     serialized_bytes = bytes.fromhex(serialized_for_signing)
     signature = keypairs_sign(serialized_bytes, wallet.private_key)
@@ -194,8 +194,12 @@ def _prepare_transaction(
 
     Raises:
         XRPLException: if both LastLedgerSequence and `ledger_offset` are provided, or
-            if an address tag is provided that does not match the X-Address tag.
+            if an address tag is provided that does not match the X-Address tag, or if
+            attempting to directly sign a Batch inner transaction.
     """
+    if transaction.batch_txn is not None:
+        raise XRPLException("Cannot directly sign a batch inner transaction.")
+
     transaction_json = transaction.to_xrpl()
     transaction_json["SigningPubKey"] = wallet.public_key
 
@@ -214,7 +218,7 @@ def _prepare_transaction(
     return transaction_json
 
 
-T = TypeVar("T", bound=Transaction)
+T = TypeVar("T", bound=Transaction, default=Transaction)
 
 
 async def autofill(
@@ -385,7 +389,7 @@ def _convert_to_classic_address(json: Dict[str, Any], field: str) -> None:
         field: the field in `json` that may contain an X-Address
     """
     if field in json and is_valid_xaddress(json[field]):
-        json[field] = xaddress_to_classic_address(json[field])
+        json[field] = xaddress_to_classic_address(json[field])[0]
 
 
 def transaction_json_to_binary_codec_form(dictionary: Dict[str, Any]) -> Dict[str, Any]:
--- a/xrpl/asyncio/transaction/main.py
+++ b/xrpl/asyncio/transaction/main.py
@@ -1,6 +1,7 @@
 """High-level transaction methods with XRPL transactions."""
+
 import math
-from typing import Any, Dict, Optional, cast
+from typing import Any, Dict, List, Optional, Tuple, cast
 
 from typing_extensions import Final
 
@@ -14,6 +15,7 @@ from xrpl.core.keypairs.main import sign as keypairs_sign
 from xrpl.models.requests import ServerInfo, ServerState, SubmitOnly
 from xrpl.models.response import Response
 from xrpl.models.transactions import EscrowFinish
+from xrpl.models.transactions.batch import Batch, BatchInnerTransaction
 from xrpl.models.transactions.transaction import Signer, Transaction
 from xrpl.models.transactions.transaction import (
     transaction_json_to_binary_codec_form as model_transaction_to_binary_codec,
@@ -244,6 +246,11 @@ async def autofill(
     if "last_ledger_sequence" not in transaction_json:
         ledger_sequence = await get_latest_validated_ledger_sequence(client)
         transaction_json["last_ledger_sequence"] = ledger_sequence + _LEDGER_OFFSET
+    if transaction.transaction_type == TransactionType.BATCH:
+        inner_txs, tx_ids = await _autofill_batch(client, cast(Batch, transaction))
+        transaction_json["raw_transactions"] = inner_txs
+        if "tx_ids" not in transaction_json:
+            transaction_json["tx_ids"] = tx_ids
     return Transaction.from_dict(transaction_json)
 
 
@@ -482,3 +489,37 @@ async def _fetch_owner_reserve_fee(client: Client) -> int:
     server_state = await client._request_impl(ServerState())
     fee = server_state.result["state"]["validated_ledger"]["reserve_inc"]
     return int(fee)
+
+
+async def _autofill_batch(
+    client: Client, transaction: Batch
+) -> Tuple[List[BatchInnerTransaction], List[str]]:
+    account_sequences: Dict[str, int] = {}
+    batch_index = 0
+    tx_ids: List[str] = []
+    inner_txs: List[BatchInnerTransaction] = []
+
+    for raw_txn in transaction.raw_transactions:
+        if raw_txn.BatchTxn is not None:
+            inner_txs.append(raw_txn)
+            continue
+
+        batch_txn = {"outer_account": transaction.account}
+
+        if raw_txn.account in account_sequences:
+            batch_txn["sequence"] = account_sequences[raw_txn.account]
+            account_sequences[raw_txn.account] += 1
+        else:
+            sequence = await get_next_valid_seq_number(raw_txn.account, client)
+            account_sequences[raw_txn.account] = sequence + 1
+            batch_txn["sequence"] = sequence
+
+        batch_txn["batch_index"] = batch_index
+        batch_index += 1
+
+        raw_txn_dict = raw_txn.to_dict()
+        raw_txn_dict["RawTransaction"]["BatchTxn"] = batch_txn
+        inner_txs.append(BatchInnerTransaction.from_dict(raw_txn_dict))
+        tx_ids.append(raw_txn.get_hash())
+
+    return inner_txs, tx_ids
--- a/xrpl/core/binarycodec/definitions/definitions.json
+++ b/xrpl/core/binarycodec/definitions/definitions.json
@@ -261,6 +261,16 @@
       }
     ],
     [
+      "BatchIndex",
+      {
+        "nth": 20,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "UInt8"
+      }
+    ],
+    [
       "LedgerEntryType",
       {
         "nth": 1,
@@ -364,6 +374,16 @@
       "HookApiVersion",
       {
         "nth": 20,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "UInt16"
+      }
+    ],
+    [
+      "LedgerFixType",
+      {
+        "nth": 21,
         "isVLEncoded": false,
         "isSerialized": true,
         "isSigningField": true,
@@ -1981,6 +2001,16 @@
       }
     ],
     [
+      "InnerResult",
+      {
+        "nth": 30,
+        "isVLEncoded": true,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "Blob"
+      }
+    ],
+    [
       "Account",
       {
         "nth": 1,
@@ -2141,6 +2171,16 @@
       }
     ],
     [
+      "OuterAccount",
+      {
+        "nth": 24,
+        "isVLEncoded": true,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "AccountID"
+      }
+    ],
+    [
       "Indexes",
       {
         "nth": 1,
@@ -2174,6 +2214,16 @@
       "NFTokenOffers",
       {
         "nth": 4,
+        "isVLEncoded": true,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "Vector256"
+      }
+    ],
+    [
+      "TxIDs",
+      {
+        "nth": 5,
         "isVLEncoded": true,
         "isSerialized": true,
         "isSigningField": true,
@@ -2551,6 +2601,46 @@
       }
     ],
     [
+      "RawTransaction",
+      {
+        "nth": 33,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "STObject"
+      }
+    ],
+    [
+      "BatchExecution",
+      {
+        "nth": 34,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "STObject"
+      }
+    ],
+    [
+      "BatchTxn",
+      {
+        "nth": 35,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "STObject"
+      }
+    ],
+    [
+      "BatchSigner",
+      {
+        "nth": 36,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "STObject"
+      }
+    ],
+    [
       "Signers",
       {
         "nth": 3,
@@ -2739,6 +2829,36 @@
         "isSigningField": true,
         "type": "STArray"
       }
+    ],
+    [
+      "BatchExecutions",
+      {
+        "nth": 26,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "STArray"
+      }
+    ],
+    [
+      "RawTransactions",
+      {
+        "nth": 27,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "STArray"
+      }
+    ],
+    [
+      "BatchSigners",
+      {
+        "nth": 28,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": false,
+        "type": "STArray"
+      }
     ]
   ],
   "TRANSACTION_RESULTS": {
@@ -2808,6 +2928,7 @@
     "temEMPTY_DID": -254,
     "temARRAY_EMPTY": -253,
     "temARRAY_TOO_LARGE": -252,
+    "temINVALID_BATCH": -251,
 
     "tefFAILURE": -199,
     "tefALREADY": -198,
@@ -2830,6 +2951,7 @@
     "tefTOO_BIG": -181,
     "tefNO_TICKET": -180,
     "tefNFTOKEN_IS_NOT_TRANSFERABLE": -179,
+    "tefINVALID_LEDGER_FIX_TYPE": -178,
 
     "terRETRY": -99,
     "terFUNDS_SPENT": -98,
@@ -2923,7 +3045,8 @@
     "tecINVALID_UPDATE_TIME": 188,
     "tecTOKEN_PAIR_NOT_FOUND": 189,
     "tecARRAY_EMPTY": 190,
-    "tecARRAY_TOO_LARGE": 191
+    "tecARRAY_TOO_LARGE": 191,
+    "tecBATCH_FAILURE": 192
   },
   "TRANSACTION_TYPES": {
     "Invalid": -1,
@@ -2974,6 +3097,8 @@
     "DIDDelete": 50,
     "OracleSet": 51,
     "OracleDelete": 52,
+    "LedgerStateFix": 53,
+    "Batch": 54,
     "EnableAmendment": 100,
     "SetFee": 101,
     "UNLModify": 102
--- a/xrpl/models/transactions/batch.py
+++ b/xrpl/models/transactions/batch.py
@@ -1,0 +1,63 @@
+"""Model for Batch transaction type."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import List, Optional
+
+from xrpl.models.nested_model import NestedModel
+from xrpl.models.required import REQUIRED
+from xrpl.models.transactions.transaction import Signer, Transaction
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class BatchSigner(NestedModel):
+    """Represents a Batch signer."""
+
+    account: str = REQUIRED  # type: ignore
+
+    signing_pub_key: Optional[str] = None
+
+    txn_signature: Optional[str] = None
+
+    signers: Optional[List[Signer]] = None
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class BatchTxn(NestedModel):
+    """Represents the info indicating a Batch transaction."""
+
+    outer_account: str = REQUIRED  # type: ignore
+
+    sequence: Optional[int] = None
+
+    ticket_sequence: Optional[int] = None
+
+    batch_index: int = REQUIRED  # type: ignore
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class BatchInnerTransaction(Transaction):
+    """Represents a Batch inner transaction."""
+
+    BatchTxn: BatchTxn = REQUIRED  # type: ignore
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class Batch(Transaction):
+    """Represents a Batch transaction."""
+
+    raw_transactions: List[BatchInnerTransaction] = REQUIRED  # type: ignore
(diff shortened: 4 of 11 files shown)


mvadari:
cc @dangell7 @tequdev

xrpl/transaction/batch_signers.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    involved_accounts = set(tx.account for tx in transaction.raw_transactions)
+    if wallet.address not in involved_accounts:
+        raise XRPLException("Must be signing for an address included in the Batch.")
+
+    fields_to_sign: Dict[str, Any] = {
+        "flags": transaction.flags,
+        "transaction_ids": [tx.get_hash() for tx in transaction.raw_transactions],
+    }
+    if multisign:
```
ckeshava, line 40: What is the difference between this `multisign` case and the `else` case below?

xrpl/transaction/batch_signers.py
```diff
@@ -0,0 +1,135 @@
...
+            or tx.signers is not None
+        ):
+            raise XRPLException("Transaction must be unsigned.")
+
+    batch_txs = cast(List[Batch], decoded_txs)
+    _validate_batch_equivalence(batch_txs)
+
+    return encode(_get_batch_with_all_signers(batch_txs).to_xrpl())
+
+    return ""
```
ckeshava, line 107:
```suggestion
```

tests/integration/transactions/test_batch.py
```diff
@@ -0,0 +1,51 @@
...
+            amount="1",
+            destination=WALLET.address,
+        )
+        batch = Batch(
+            account=WALLET.address,
+            flags=BatchFlag.TF_ALL_OR_NOTHING,
+            raw_transactions=[payment, payment2],
+        )
+        autofilled = await autofill(batch, client, 2)
+        signed = sign_multiaccount_batch(DESTINATION, autofilled)
```
ckeshava, line 48:
```suggestion
        signed = sign_multiaccount_batch(WALLET, autofilled)
```

Since the outer batch transaction is sent from `WALLET.address`, shouldn't this be changed?

xrpl/asyncio/transaction/main.py
```diff
@@ -300,7 +315,7 @@ def _tx_needs_networkID(client: Client) -> bool:
     Returns:
         bool: whether the transactions required network ID to be valid
     """
-    if client.network_id and client.network_id > _RESTRICTED_NETWORKS:
+    if client.network_id is not None and client.network_id > _RESTRICTED_NETWORKS:
```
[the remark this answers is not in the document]
mvadari: No, this statement is `False` if `client.network_id == 0`

xrpl/asyncio/transaction/main.py
```diff
@@ -215,28 +220,27 @@ async def simulate(
...
 
     Raises:
         XRPLException: if both LastLedgerSequence and `ledger_offset` are provided, or
-            if an address tag is provided that does not match the X-Address tag.
+            if an address tag is provided that does not match the X-Address tag, or if
+            attempting to directly sign a Batch inner transaction.
     """
+    if transaction.has_flag(TransactionFlag.TF_INNER_BATCH_TXN):
+        raise XRPLException("Cannot directly sign a batch inner transaction.")
+
```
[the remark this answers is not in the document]
mvadari: `_prepare_transaction` is a helper function that is only used in `sign`.

xrpl/transaction/main.py
```diff
@@ -108,8 +112,8 @@ def autofill_and_sign(
 
 
 def autofill(
-    transaction: Transaction, client: SyncClient, signers_count: Optional[int] = None
-) -> Transaction:
+    transaction: T, client: SyncClient, signers_count: Optional[int] = None
+) -> T:
     """
     Autofills fields in a transaction. This will set `sequence`, `fee`, and
     `last_ledger_sequence` according to the current state of the server this Client is
```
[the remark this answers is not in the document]
mvadari: What exceptions?

tests/integration/sugar/test_transaction.py
```diff
@@ -326,6 +335,37 @@ async def test_simulate(self, client):
...
+        for i in range(len(transaction.raw_transactions)):
+            raw_tx = transaction.raw_transactions[i]
+            self.assertTrue(raw_tx.has_flag(TransactionFlag.TF_INNER_BATCH_TXN))
+            self.assertEqual(raw_tx.sequence, sequence + i)
+            self.assertIsNone(raw_tx.network_id)
+            self.assertIsNone(raw_tx.last_ledger_sequence)
+            self.assertEqual(raw_tx.fee, "0")
+            self.assertEqual(raw_tx.signing_pub_key, "")
+            self.assertEqual(raw_tx.txn_signature, "")
+
```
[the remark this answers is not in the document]
mvadari: The outer Batch transaction is a normal transaction, so it's validated normally.

xrpl/transaction/batch_signers.py
mvadari: Whether or not the account is signing the transaction with multisign.

tests/integration/transactions/test_batch.py
mvadari: `WALLET` is signing the transaction normally in the line below. This line gets the `DESTINATION` signature into the Batch as a `BatchSigner`.

lmaisons:
I could only get this to work by signing the interior transactions in addition to signing the full batch, otherwise I get an exception about being unable to retrieve the hash from an unsigned transaction. At least to my ignorance, that's surprising given [if I recall correctly] the spec specifically states that the transaction atoms are specifically unsigned.
```python
 def toy(url: str):
     xrpl_client = JsonRpcClient(url=url)
     alice = Wallet.from_seed(seed='OMITTED', algorithm=CryptoAlgorithm.ED25519)
     bob = Wallet.from_seed(seed='OMITTED', algorithm=CryptoAlgorithm.ED25519)
     alice_account_info = xrpl_client.request(AccountInfo(account=alice.address)).result['account_data']
     bob_account_info = xrpl_client.request(AccountInfo(account=bob.address)).result['account_data']
     alice_sequence_start = alice_account_info['Sequence']
     bob_sequence_start = bob_account_info['Sequence']
-    payment_alice_to_bob = Payment(account=alice.address, amount=str(1), destination=bob.address, fee=str(0), sequence=alice_sequence_start + 1)
-    payment_bob_to_alice = Payment(account=bob.address, amount=str(1), destination=alice.address, fee=str(0), sequence=bob_sequence_start)
+    payment_alice_to_bob = transaction.sign(Payment(account=alice.address, amount=str(1), destination=bob.address, fee=str(0), sequence=alice_sequence_start + 1), alice)
+    payment_bob_to_alice = transaction.sign(Payment(account=bob.address, amount=str(1), destination=alice.address, fee=str(0), sequence=bob_sequence_start), bob)
     base_fee = 10
     number_of_signers = 2
     number_of_transactions = 2
     fee = (number_of_signers + 2 + number_of_transactions) * base_fee
     batch = Batch(account=alice.address, raw_transactions=[payment_alice_to_bob, payment_bob_to_alice], sequence=alice_sequence_start, fee=str(fee))
     bob_consent = transaction.sign_multiaccount_batch(alice, batch)
     alice_submission = transaction.sign(bob_consent, alice)
     batch_response = transaction.submit(transaction=alice_submission, client=xrpl_client)
     print('Payment ' + ('successful' if batch_response.is_successful() else 'failed'))
```
Else:
```
  File "/Users/lmaisons/src/rippled_sandbox/main.py", line 155, in <module>
    sys.exit(main())
  File "/Users/lmaisons/src/rippled_sandbox/main.py", line 151, in main
    junk(url)
  File "/Users/lmaisons/src/rippled_sandbox/main.py", line 28, in junk
    bob_consent = transaction.sign_multiaccount_batch(alice, batch)
  File "/Users/lmaisons/src/xrpl-py/xrpl/transaction/batch_signers.py", line 38, in sign_multiaccount_batch
    "transaction_ids": [tx.get_hash() for tx in transaction.raw_transactions],
  File "/Users/lmaisons/src/xrpl-py/xrpl/transaction/batch_signers.py", line 38, in <listcomp>
    "transaction_ids": [tx.get_hash() for tx in transaction.raw_transactions],
  File "/Users/lmaisons/src/xrpl-py/xrpl/models/transactions/transaction.py", line 444, in get_hash
    raise XRPLModelException(
xrpl.models.exceptions.XRPLModelException: Cannot get the hash from an unsigned Transaction.
```

mvadari:
@lmaisons the immediate cause to that issue is that the inner transactions don't have the `tfInnerBatchTxn` flag. It's added via `autofill`, which it looks like you're not using. 

Let me see if I can come up with a better solution for that, where you don't need to use `autofill`. But in the meanwhile, adding that flag should fix your issue.

lmaisons:
Confirming this worked as described:
```
-    payment_alice_to_bob = Payment(account=alice.address, amount=str(1), destination=bob.address, fee=str(0), sequence=alice_sequence_start + 1)
-    payment_bob_to_alice = Payment(account=bob.address, amount=str(1), destination=alice.address, fee=str(0), sequence=bob_sequence_start)
+    payment_alice_to_bob = Payment(account=alice.address, amount=str(1), destination=bob.address, fee=str(0), sequence=alice_sequence_start + 1, flags=TransactionFlag.TF_INNER_BATCH_TXN)
+    payment_bob_to_alice = Payment(account=bob.address, amount=str(1), destination=alice.address, fee=str(0), sequence=bob_sequence_start, flags=TransactionFlag.TF_INNER_BATCH_TXN)
```

xrpl/asyncio/transaction/main.py
```diff
@@ -498,3 +519,76 @@ async def _fetch_owner_reserve_fee(client: Client) -> int:
...
+                "Cannot have a Batch transaction inside a Batch transaction."
+            )
+
+        if not raw_txn.has_flag(TransactionFlag.TF_INNER_BATCH_TXN):
+            if isinstance(raw_txn.flags, int):
+                raw_txn_dict["flags"] |= TransactionFlag.TF_INNER_BATCH_TXN
+            elif isinstance(raw_txn.flags, dict):
+                raw_txn_dict["flags"]["TF_INNER_BATCH_TXN"] = True
+            else:  # is List[int]
+                raw_txn_dict["flags"].append(TransactionFlag.TF_INNER_BATCH_TXN)
```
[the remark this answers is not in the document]
ckeshava: Understood, this is a convenience provided by the client libraries, is that correct?

Based on the specification, `rippled` appears to support only one numeric field for the flag (or multiple `OR`'ed flags)

xrpl/asyncio/transaction/main.py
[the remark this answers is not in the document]
ckeshava: Yes, however that need not be the case during future updates to this file. Other developers might not be aware that `_prepapre_transaction` is supposed to used only in `sign`.

The code would be more maintainable if we dis-entagle such dependencies.

xrpl/asyncio/transaction/main.py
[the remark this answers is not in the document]
ckeshava:
```suggestion
    if client.network_id and (client.network_id != 0 and client.network_id > _RESTRICTED_NETWORKS):
```

I feel this code is better in that, it explicitly indicates the possibility of 0 ^

It might miss the minds of developers that `None` case includes zero too.

xrpl/asyncio/transaction/main.py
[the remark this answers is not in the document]
mvadari: Yes, just in the libraries

xrpl/asyncio/transaction/main.py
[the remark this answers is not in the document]
mvadari: It's pretty trivial to determine that. Doing a refactor of that is out of scope of this PR IMO.

xrpl/core/binarycodec/main.py
```diff
@@ -73,6 +71,28 @@ def encode_for_signing_claim(json: Dict[str, Any]) -> str:
...
+    prefix = _BATCH_PREFIX
+    flags = UInt32.from_value(json["flags"])
+    transaction_ids = json["transaction_ids"]
+    len_transaction_ids = UInt32.from_value(len(transaction_ids))
+
+    buffer = prefix + bytes(flags) + bytes(len_transaction_ids)
+    for tx in transaction_ids:
+        buffer += bytes(Hash256.from_value(tx))
+
+    return buffer.hex().upper()
```
ckeshava, line 93: Where can I find the description of the serialization/deserialization algorithm for the `Batch` transaction?

I didn't observe it in the [spec](https://github.com/XRPLF/XRPL-Standards/discussions/162)

More generally, is this an accepted norm for encoding new transaction-types? I'm asking about points like how to determine order of the fields, what data types can we use for each of the fields, etc.

For instance, the length of the transaction-IDs could have been represented with a UINT8 (since its at most 8 elements) and a UINT32. How did you decide on one type?

xrpl/core/binarycodec/definitions/definitions.json
```diff
@@ -2523,25 +2563,56 @@
...
+    ],
+    [
+      "AcceptedCredentials",
+      {
+        "isSerialized": true,
+        "isSigningField": true,
         "isVLEncoded": false,
+        "nth": 28,
+        "type": "STArray"
+      }
```
ckeshava, line 2591: `AcceptedCredentials` is stored twice in this JSON file. Check line 2614.

xrpl/transaction/batch_signers.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return encode(_get_batch_with_all_signers(batch_txs).to_xrpl())
+
+
+def _validate_batch_equivalence(transactions: List[Batch]) -> None:
+    example_tx = transactions[0]
+    for tx in transactions:
+        if (
+            tx.flags != example_tx.flags
+            or tx.raw_transactions != example_tx.raw_transactions
```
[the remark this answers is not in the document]
mvadari: > Do you know if this comparison is deep? I think the below condition is more robust: `or tx.raw_transactions is not example_tx.raw_transactions`. This condition compares the memory addresses of the left and right operands.

That is an incorrect check. We don't want to check the memory addresses, we want to check the data equivalence.

> Furthermore, the `Transaction` class does not have a `__eq__` method defined for its class. Given the widespread usage of that class, the below method ensures correctness and is more efficient:
> 
> ```
> def __eq__(self, other):
>     return self.get_hash() == other.get_hash()
> ```

Dataclasses automatically generate an `__eq__`: https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass

tests/unit/core/binarycodec/test_main.py
```diff
@@ -401,6 +402,32 @@ def test_claim(self):
...
+                [
+                    # hash prefix
+                    "42434800",
+                    # flags
+                    "00000001",
+                    # transaction_ids length
+                    "00000002",
+                    # transaction_ids
+                    "ABE4871E9083DF66727045D49DEEDD3A6F166EB7F8D1E92FE868F02E76B2C5CA",
+                    "795AAC88B59E95C3497609749127E69F12958BC016C600C770AEEB1474C840B4",
```
ckeshava, line 426: I like that this output is split according to different fields. It helps me understand better, nice work

xrpl/transaction/batch_signers.py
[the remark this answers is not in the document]
mvadari: No it doesn't, because they can be edited separately.
```
>>> a = {1: 2}
>>> b = {1: 2}
>>> a is b
False
>>> a == b
True
```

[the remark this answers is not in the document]
ckeshava: > Dataclasses automatically generate an __eq__

The generated __eq__ method is inefficient. It compares all the fields of the `Transaction` object. My suggestion to use the transaction hashes is faster (assuming it is computationally hard to generate the hash collisions)

image



Furthermore, the [docs specify](https://docs.python.org/3/reference/datamodel.html#object.__eq__)  that `dataclass.object` uses `is` operator to implement its `__eq__` method


image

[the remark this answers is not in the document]
mvadari: The level of computational difference doesn't make much of a difference for this library.

ckeshava:
I didn't see a mention of batchnet in this PR. Did you intend to add that faucet-host in this python library? Here's the [relevant file](https://github.com/XRPLF/xrpl.js/pull/2801/files#diff-06e6fcfbef84f4790b1e91e8790acd725ebe144d33bf3c10c233540cbaae48c0) from JS library PR: 

[This](https://github.com/XRPLF/xrpl-py/blob/f1488c175c368584636ad217479bf96739623142/xrpl/asyncio/wallet/wallet_generation.py#L17) appears to be a suitable place for such an addition.

mvadari:
> I didn't see a mention of batchnet in this PR. Did you intend to add that faucet-host in this python library? Here's the [relevant file](https://github.com/XRPLF/xrpl.js/pull/2801/files#diff-06e6fcfbef84f4790b1e91e8790acd725ebe144d33bf3c10c233540cbaae48c0) from JS library PR:
> 
> [This](https://github.com/XRPLF/xrpl-py/blob/f1488c175c368584636ad217479bf96739623142/xrpl/asyncio/wallet/wallet_generation.py#L17) appears to be a suitable place for such an addition.

I published a beta for xrpl.js, so the faucet was more necessary to include there. It's not super necessary for Python given that I haven't been asked to create a beta (given that Batchnet will no longer be necessary by the time this PR is merged).

xrpl/core/binarycodec/main.py
mvadari: https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0056d-batch#2132-signingpubkey-and-txnsignature

xrpl/asyncio/transaction/main.py
```diff
@@ -90,40 +93,42 @@ def sign(
...
             wallet.private_key,
         )
-        tx_dict = transaction.to_dict()
-        tx_dict["signers"] = [
-            Signer(
-                account=wallet.address,
-                txn_signature=signature,
-                signing_pub_key=wallet.public_key,
-            )
+        transaction_json["Signers"] = [
```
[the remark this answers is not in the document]
mvadari: Note the difference between the use of `to_dict` vs `to_xrpl`

xrpl/asyncio/transaction/main.py
```diff
@@ -90,40 +93,42 @@ def sign(
...
+                    "TxnSignature": signature,
+                    "SigningPubKey": wallet.public_key,
+                }
+            }
         ]
-        return Transaction.from_dict(tx_dict)
+        return cast(T, Transaction.from_xrpl(transaction_json))
 
-    transaction_json = _prepare_transaction(transaction, wallet)
+    transaction_json["SigningPubKey"] = wallet.public_key
```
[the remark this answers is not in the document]
mvadari: The result is returned in line 116 in the multisign case, so this line isn't hit then.

xrpl/transaction/batch_signers.py
```diff
@@ -0,0 +1,133 @@
...
+def sign_multiaccount_batch(
+    wallet: Wallet, transaction: Batch, multisign: Optional[bool] = False
+) -> Batch:
+    """
+    Sign a multi-account Batch transaction.
+
+    Args:
+        wallet: Wallet instance.
+        transaction: The Batch transaction to sign.
+        multisign: Specify true/false to use multisign. Defaults to False.
```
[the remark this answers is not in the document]
mvadari: Fair

xrpl/asyncio/transaction/main.py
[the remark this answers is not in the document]
mvadari: I suggest looking at the code as a whole and not just the diff. It should make a lot of things more clear.

mvadari:
The `develop` Docker container needs to be updated before integration tests will pass. They do pass locally.

xrpl/transaction/batch_signers.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    involved_accounts = set(tx.account for tx in transaction.raw_transactions)
+    if wallet.address not in involved_accounts:
+        raise XRPLException("Must be signing for an address included in the Batch.")
+
+    fields_to_sign: Dict[str, Any] = {
+        "flags": transaction.flags,
+        "transaction_ids": [tx.get_hash() for tx in transaction.raw_transactions],
+    }
+    if multisign:
```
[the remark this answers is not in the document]
mvadari: fixed

xrpl/transaction/batch_signers.py
```diff
@@ -0,0 +1,134 @@
...
+            tx.signing_pub_key != ""
+            or tx.txn_signature is not None
+            or tx.signers is not None
+        ):
+            raise XRPLException("Transaction must be unsigned.")
+
+    batch_txs = cast(List[Batch], decoded_txs)
+    _validate_batch_equivalence(batch_txs)
+
+    return encode(_get_batch_with_all_signers(batch_txs).to_xrpl())
```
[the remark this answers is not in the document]
mvadari: That's already the signature of that function, there is no linter error

.ci-config/rippled.cfg
```diff
@@ -183,20 +183,24 @@ fixEmptyDID
 fixXChainRewardRounding
 fixPreviousTxnID
 # 2.3.0 Amendments
-fixAMMv1_1
```
ckeshava, line 186: Why is `fixAMMv1_1` being removed from this file?

.ci-config/rippled.cfg
```diff
@@ -183,20 +183,24 @@ fixEmptyDID
...
-fixAMMv1_2
 AMMClawback
-InvariantsV1_1
```
ckeshava, line 189: Did you mean to remove `InvariantsV1_1` amendment?

tests/integration/sugar/test_transaction.py
```diff
@@ -326,6 +335,37 @@ async def test_simulate(self, client):
...
+        )
+        transaction = await autofill(tx, client)
+
+        sequence = await get_next_valid_seq_number(WALLET.address, client)
+        for i in range(len(transaction.raw_transactions)):
+            raw_tx = transaction.raw_transactions[i]
+            self.assertTrue(raw_tx.has_flag(TransactionFlag.TF_INNER_BATCH_TXN))
+            self.assertEqual(raw_tx.sequence, sequence + i)
+            self.assertEqual(raw_tx.network_id, 63456)
+            self.assertIsNone(raw_tx.last_ledger_sequence)
```
ckeshava, line 364: Since the `last_ledger_sequence` parameter is `None` in the inner-transactions, I'm thinking about the possibility of transaction replay attacks.

Suppose this batch transaction fails (for an unspecified `tem...` error code. Note: `tem...` code does not increment the sequence number of the sender `WALLET` account), I can grab the `tx_blob` and re-submit it at an appropriate later date. As a malicious attacker, I have time until the `sequence number` of the WALLET account changes.

Am I missing something here? Has this been addressed?

xrpl/models/transactions/batch.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def _get_errors(self: Self) -> Dict[str, str]:
+        errors = super()._get_errors()
+        for i in range(len(self.raw_transactions)):
+            tx = self.raw_transactions[i]
+            if not tx.has_flag(TransactionFlag.TF_INNER_BATCH_TXN):
+                errors[f"raw_transactions[{i}]"] = (
+                    "RawTransaction must have tfInnerBatchTxn flag set."
+                )
+        return errors
```
ckeshava, line 120: Can you add a check + unit-test on the number of raw-transactions ? It needs to be at least two as per this implementation: https://github.com/XRPLF/rippled/blob/8bf4a5cbff4392fcca9990d931ba528085b9a22a/src/xrpld/app/tx/detail/Batch.cpp#L214

xrpl/models/transactions/batch.py
```diff
@@ -0,0 +1,149 @@
...
+            value: The value to construct the Batch from.
+
+        Returns:
+            A new Batch object, constructed using the given parameters.
+
+        Raises:
+            XRPLModelException: If the dictionary provided is invalid.
+        """
+        new_value = {**value}
+        new_value["raw_transactions"] = [
```
ckeshava, line 137: The `__post_init__` dunder of the Batch class already initializes a `self.raw_transactions` field. Why do we need to do this manually inside the `from_dict` method?

.ci-config/rippled.cfg
mvadari: It's not listed here: https://livenet.xrpl.org/amendments

It's marked as not supported yet: https://github.com/XRPLF/rippled/blob/develop/include/xrpl/protocol/detail/features.macro#L55

.ci-config/rippled.cfg
mvadari: typo

tests/integration/sugar/test_transaction.py
mvadari: 1. The _inner_ transaction doesn't have a LastLedgerSequence. Nothing about the _outer_ transaction. An inner transaction is not valid on its own.
2. A `tem` error is only available to the node that the transaction is submitted to.
3. That's not what `LastLedgerSequence` is intended to be about (and it doesn't do anything for it) - its purpose is just so that your transaction doesn't suddenly become valid hours or days after you submit it, when you didn't intend that.

xrpl/models/transactions/batch.py
mvadari: The `from_dict` doesn't add a `raw_transactions` field, it updates each field in `self.raw_transactions` to remove the wrapper `RawTransaction` inner object.

xrpl/models/transactions/batch.py
mvadari: done

xrpl/models/transactions/batch.py
mvadari: Also, the `__post_init__` doesn't create that field, that's already created by the dataclass-created constructor. That function just automatically adds the `tfInnerBatchTxn` flag.

xrpl/models/transactions/batch.py
```diff
@@ -0,0 +1,151 @@
...
+                new_raw_transactions.append(tx)
+            else:
+                tx_dict = tx.to_dict()
+                if "flags" not in tx_dict:
+                    tx_dict["flags"] = TransactionFlag.TF_INNER_BATCH_TXN
+                elif isinstance(tx_dict["flags"], int):
+                    tx_dict["flags"] += TransactionFlag.TF_INNER_BATCH_TXN
+                elif isinstance(tx_dict["flags"], dict):
+                    tx_dict["flags"]["TF_INNER_BATCH_TXN"] = True
+                elif isinstance(self.flags, list):
```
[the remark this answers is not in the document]
ckeshava: @mvadari  this appears to be a valid concern from code-rabbit ^^

xrpl/asyncio/transaction/main.py
```diff
@@ -480,12 +496,18 @@ async def _calculate_fee_per_transaction_type(
...
         TransactionType.ACCOUNT_DELETE,
         TransactionType.AMM_CREATE,
     ):
         base_fee = await _fetch_owner_reserve_fee(client)
 
+    elif transaction.transaction_type == TransactionType.BATCH:
+        batch = cast(Batch, transaction)
+        base_fee = base_fee * (2 + len(batch.raw_transactions))
+        if batch.batch_signers is not None:
+            base_fee += base_fee * len(batch.batch_signers)
```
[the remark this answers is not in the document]
ckeshava: Hello, code-rabbit appears to be correct in raising concerns about the fee calculation. 

Furthermore, I adhered to the cpp implementation and wrote the below snippet. (the snippet does not have any of the math overflow checks). It is not clear to me if your implementation is identical to the cpp implementation.

Reference: https://github.com/XRPLF/rippled/blob/8bf4a5cbff4392fcca9990d931ba528085b9a22a/src/xrpld/app/tx/detail/Batch.cpp#L54
```
...
elif transaction.transaction_type == TransactionType.BATCH: 
    batch = cast(Batch, transaction)
    total_batch_fee = 0
    
    inner_txn_fees = 0
    # compute the fees for all the inner transactions
    for inner_txn in batch.raw_transactions:
        inner_txn_fees += _calculate_fee_per_transaction_type(txn)
        
    signer_count = 0
    # compute the fees pertaining to multiple-signers
    if batch.batch_signers is not None:
        for signer in batch.batch_signers:
            if signer.txn_signature is not None:
                signer_count += 1
            if signer.signers is not None:
                signer_count += len(signer.signers)
                
    signer_fees = signer_count * base_fee
                
    total_batch_fee = (2 * base_fee) + inner_txn_fees + signer_fees

```

xrpl/models/transactions/batch.py
```diff
@@ -0,0 +1,151 @@
...
+    def _get_errors(self: Self) -> Dict[str, str]:
+        errors = super()._get_errors()
+        for i in range(len(self.raw_transactions)):
+            tx = self.raw_transactions[i]
+            if not tx.has_flag(TransactionFlag.TF_INNER_BATCH_TXN):
+                errors[f"raw_transactions[{i}]"] = (
+                    "RawTransaction must have tfInnerBatchTxn flag set."
+                )
+        if len(self.raw_transactions) < 2:
+            errors["raw_transactions"] = "Batch must contain at least 2 transactions."
```
ckeshava, line 119:
```suggestion
            errors["raw_transactions"] += "Batch must contain at least 2 transactions."
```

xrpl/asyncio/transaction/main.py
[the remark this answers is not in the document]
mvadari: fixed

xrpl/models/transactions/batch.py
mvadari: The error message keys are different, so this isn't necessary (and would also result in errors most of the time, since the `raw_transactions` key wouldn't exist).

xrpl/models/transactions/batch.py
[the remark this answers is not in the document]
mvadari: fixed

xrpl/asyncio/transaction/main.py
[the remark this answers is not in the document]
mvadari: I removed the `BatchSigners` fee logic and just amended the docstring instead to include that in the `signers_count` (similar to multi-signed transactions), as `autofill` runs before we have signatures.

tests/unit/models/transactions/test_batch.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_basic(self):
+        payment = Payment(
+            account=_ACCOUNT,
+            amount="1",
+            destination=_DESTINATION,
+        )
+        batch = Batch(
+            account=_ACCOUNT,
+            flags=BatchFlag.TF_ALL_OR_NOTHING,
+            raw_transactions=[payment, payment],
```
ckeshava, line 20: I think this unit test can be improved. This example will throw a `temREDUNDANT` because it has not been autofilled with unique sequence numbers (it has identical `payment` transactions). Readers could be misled by this example.

xrpl/transaction/batch_signers.py
```diff
@@ -0,0 +1,134 @@
...
+    }
+    if multisign:
+        signer = Signer(
+            account=wallet.address,
+            signing_pub_key=wallet.public_key,
+            txn_signature=sign(
+                encode_for_signing_batch(fields_to_sign), wallet.private_key
+            ),
+        )
+        batch_signer = BatchSigner(account=wallet.address, signers=[signer])
```
ckeshava, line 49: I'm confused: In the case of multi-sign, all of the accounts involved as a "sender" in the raw-transactions need to sign the entire batch transaction. Where is that happening? Can you point me to the relevant code?

This piece of code only adds the signature of `wallet.address`, which is the sender of the outer transaction.

Are users expected to piece together disparate signatures manually by using the `combine_batch_signers` method? Apart from the unit tests, I did not find an integration test (or) a snippet which demonstrates using this method.

xrpl/models/transactions/batch.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def to_dict(self: Self) -> Dict[str, Any]:
+        """
+        Returns the dictionary representation of a Batch.
+
+        Returns:
+            The dictionary representation of a Batch.
+        """
+        tx_dict = super().to_dict()
+        tx_dict["raw_transactions"] = [
```
ckeshava, line 155: Help me understand this question: Why are you manually unpacking/packing the `RawTransactions` field? There are other "inner-object" (in rippled parlance)/ `NestedModel` fields which do not have this special code in their `from_dict` and `to_dict` method definitions. (`AuthAccount`, `BatchSigner` are two examples).

Aren't inner-objects handled by the JSON library automatically? Do we need to spell out arbitrary depth nesting of the JSON object structure?

(The below structure is my mental model of the `RawTransactions` STArray inner object)

```
RawTransactions: [
    {
      RawTransaction: {
        TransactionType: "OfferCreate",
        ...
      }
    },
    {
      RawTransaction: {
        TransactionType: "Payment",
        ...
      }
    },
  ]
```

xrpl/models/transactions/batch.py
```diff
@@ -0,0 +1,158 @@
...
+            # Ensure that every inner transaction has the tfInnerBatchTxn flag set.
+            # If it does not, set it.
+            if tx.has_flag(TransactionFlag.TF_INNER_BATCH_TXN):
+                new_raw_transactions.append(tx)
+            else:
+                tx_dict = tx.to_dict()
+                if "flags" not in tx_dict:
+                    tx_dict["flags"] = TransactionFlag.TF_INNER_BATCH_TXN
+                elif isinstance(tx_dict["flags"], int):
+                    tx_dict["flags"] += TransactionFlag.TF_INNER_BATCH_TXN
```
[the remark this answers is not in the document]
ckeshava: in this context, there is at most one possible flag. Hence, both the operations are equivalent.

tests/integration/transactions/test_batch.py
```diff
@@ -0,0 +1,51 @@
...
+            account=DESTINATION.address,
+            amount="1",
+            destination=WALLET.address,
+        )
+        batch = Batch(
+            account=WALLET.address,
+            flags=BatchFlag.TF_ALL_OR_NOTHING,
+            raw_transactions=[payment, payment2],
+        )
+        autofilled = await autofill(batch, client, 2)
```
ckeshava: Do we expect the users to accurately fill-in the value for `signers_count` in this `autofill` method?

If you execute this test with `autofilled = await autofill(batch, client)` -- I have not specified the number of `signers_count`, then the test fails (with `telINSUF_FEE_P` error).

This is because both the `_check_fee` and the `autofill` methods incorrectly assume that the `signers_count == 0`.

Whose responsibility is it to accurately fill the `signers_count` value? 
- If we ask the users to do it, it degrades the user-experience. Users need to be careful in computing the total number of signers, especially in cases where the multi-signing accounts also have `SignerLists` set up for them.

- Do we have any utility methods which compute the `signer_count` automatically?

ckeshava (changes requested):

xrpl/models/transactions/batch.py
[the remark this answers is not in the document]
mvadari: Made the change anyways because it doesn't hurt

tests/unit/models/transactions/test_batch.py
mvadari: I highly doubt users are using unit tests as sample code - and I sure hope they aren't, because that's really not what they're supposed to be showing. Not to mention, there's actually nothing wrong with this example - this is pre-`autofill`, which is when all the sequence numbers would be filled in, so this is a perfectly valid transaction construction.

xrpl/transaction/batch_signers.py
mvadari: Here's an integration test: https://github.com/XRPLF/xrpl-py/blob/batch/tests/integration/transactions/test_batch.py#L31

In this case, `wallet` is not the sender of the outer transaction - it's the signer.

xrpl/models/transactions/batch.py
mvadari: Note that `RawTransactions` is a list of **Transactions**, and `Transaction` does not inherit from `NestedModel`. We could theoretically have a model called `RawTransaction` that just has a `Transaction` field, but that's much less user-friendly IMO, so instead we have to do the nesting in the `Batch` class. If you take a look at the `NestedModel` class, you'll see the code handling it is pretty similar.

tests/integration/transactions/test_batch.py
mvadari: Unfortunately, there's no way to calculate it automatically - it has to be done by the user. `autofill` is executed before any of the signatures are there, and we need the signatures to know how many signers are going to be there. Theoretically users could fill in the rest of the e.g. `Signer` fields (`Account`/`PublicKey`) but in practice that's really not worth the effort just for the purpose of `autofill`, it's much easier and more user-friendly to provide the `signer_count` field

xrpl/asyncio/transaction/main.py
```diff
@@ -480,16 +501,28 @@ async def _calculate_fee_per_transaction_type(
...
+                int(await _calculate_fee_per_transaction_type(raw_txn, client))
+                for raw_txn in batch.raw_transactions
+            ]
+        )
+
+    # Multi-signed/Multi-Account Batch Transactions
     # BaseFee × (1 + Number of Signatures Provided)
     if signers_count is not None and signers_count > 0:
-        base_fee += net_fee * (1 + signers_count)
+        if transaction.transaction_type == TransactionType.BATCH:
```
ckeshava: There is a mismatch between the fee-calculations in the xrpl-py implementation and that of xrpl.js. Here is a snapshot:

image

The xrpl.js implementation does not have this transaction-type based discretion of `net_fee * signers_count` versus `net_fee * (1 + signers_count)`.

Furthermore, the current implementation does not match the rippled cpp code. 
image

I would expect to iterate over all the entries in the BatchSigners list to accurately compute the fees. 

I understand that `autofill` does not have access to the signatures and hence cannot precisely calculate this fee. Should we put out a disclaimer that the client libraries do not match the rippled node in the fee expectations?

xrpl/models/transactions/batch.py
ckeshava: You could have implemented this idea by making use of `NestedModel`.
```
class RawTransaction(NestedModel):
    raw_transaction: Transaction = REQUIRED # type: ignore

class Batch(Transaction):
    """Represents a Batch transaction."""

    raw_transactions: List[RawTransaction] = REQUIRED  # type: ignore
    # instead of the current design
    # raw_transactions: List[Transaction] = REQUIRED  # type: ignore
    
    # rest of the code
```

I don't see how this lessens the user-experience. As a bonus, you wouldn't have had to unpack/pack the raw-transactions yourself.

xrpl/transaction/batch_signers.py
ckeshava: Okay, I understand. The integ tests did not make use of the `combine_batch_signers` method (because there is only one `BatchSigner`). That's why I didn't understand the intended use of that method.

ckeshava: > In this case, wallet is not the sender of the outer transaction - it's the signer

Correction: In that integ test, `WALLET` is the sender of the outer-transaction. `DESTINATION` is one of the signers (because it sends an inner transaction).

xrpl/asyncio/transaction/main.py
mvadari: 1. Turns out that both are wrong and it should just be just `+ base_fee * signersCount` in both places - this has been fixed (and tested with integration tests).
2. rippled sees the transactions after they have been signed. This isn't comparable. The libraries see the transactions before they have been signed, so some parts of the transaction aren't available yet. The same process is done for regular multisign.

xrpl/models/transactions/batch.py
mvadari: Yes, I could have, but now every Batch looks like this:
```
Batch(account="r...", 
    raw_transactions=[
        RawTransaction(Payment(account="r...", ...)),
        RawTransaction(AccountSet(account="r...", ...)),
    ]
)
```
This is much weirder/trickier to use, instead of just having a list of transactions, which is how I would want to use it. Users don't have to do any additional work with the existing solution anyways, it "magically" works because of the additional `NestedModel`-esque code.

xrpl/transaction/batch_signers.py
mvadari: `combine_batch_signers` is analogous to the multisign methods.

xrpl/asyncio/transaction/main.py
ckeshava: 1. > Turns out that both are wrong
OK

2. It would be helpful to document the differences in fee-calculation between the client libraries and rippled nodes. Users should be aware of this quirk. This is useful for debugging erroring Batch transactions.

mvadari: I updated the docs around autofill to call it out.

CHANGELOG.md
```diff
@@ -7,17 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Add `MPTCurrency` support in `Issue` (rippled internal type)
+- Fix the implementation error in get_latest_open_ledger_sequence method. The change uses the "current" ledger for extracting sequence number
+- Increase default maximum payload size for websocket client
+- Fix the default behavior of flags field when preparing transactions. By default, flags are not part of the transaction if not explicitly provided
+- Add support for `amm_info` to `Request.from_dict`
+- Improve error handling for `amm_info`
+- Handle autofilling better when multisigning transactions
+- Improve typing for transaction-related helper functions
+- Improve handling of `TicketSequence`
+- Fix issue with failing on a higher than expected fee
```
ckeshava, line 25: if you can explain the context of this CL entry, that will be helpful.

mvadari: https://github.com/XRPLF/xrpl-py/commit/341d91a7bf2f51cb040cfb34ae639df9ce980b2c

Essentially, the `max` operator wasn't working properly since it was comparing strings, not numbers.

PR #758, opened 2024-10-16
Updates to FeeSettings ledger object

ckeshava:
This PR scrubs the xrpl-py client library of any references to "10 drops" as the Base transaction cost. The commits have been organized with the following intent:
1. Mandate the usage of `Client` parameter inside the `check_fee` method: Short of reading the latest validated ledger, I do not know of a way to grab the accurate Fees values. 
2. Re-organize the tests to incorporate point-1
3. Execute integration tests with non-historical Fees values: I've used arbitrary values to run sanity check on the code. It is difficult to determine the exact value of future `FeeSettings` ledger objects.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Release Notes

- **New Features**
	- Introduced a new `[voting]` section in server configuration with parameters for transaction fees and reserves.
	- Added `include_deleted` parameter in `ledger_entry` requests.
	- Enhanced support for transaction fee calculations, including the handling of issued currencies.

- **Bug Fixes**
	- Fixed fee settings retrieval to ensure accurate transaction cost calculations.
	- Updated tests to accurately reflect transaction fees in balance calculations.

- **Breaking Changes**
	- Removed support for Python 3.7; Python 3.8 is now the default version.
	- Updated default API version to rippled API v2.

### DIFF
--- a/.github/workflows/snippet_test.yml
+++ b/.github/workflows/snippet_test.yml
@@ -15,15 +15,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Load cached .local
         id: cache-poetry
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: /home/runner/.local
           key: dotlocal-${{ env.POETRY_VERSION }}
@@ -35,7 +35,7 @@ jobs:
           echo "$HOME/.local/bin" >> $GITHUB_PATH
 
       - name: Install Python + Retrieve Poetry dependencies from cache
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: "poetry"
--- a/.github/workflows/snippet_test.yml
+++ b/.github/workflows/snippet_test.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 env:
-  POETRY_VERSION: 1.4.2
+  POETRY_VERSION: 1.8.3
 
 jobs:
   snippet-test:
--- a/tests/integration/sugar/test_transaction.py
+++ b/tests/integration/sugar/test_transaction.py
@@ -462,6 +462,17 @@ class TestSubmitAndWait(IntegrationTestCase):
             client,
         )
 
-        # If FeeSettings ledger object is updated, the below assert needs to be changed
-        # correspondingly
-        self.assertEqual(fee, "200")
+        # The expected fee is read from the below-specified config file
+        expected_fee = ""
+        with open(".ci-config/rippled.cfg", "r", encoding="utf-8") as file:
+            lines = file.readlines()  # Read all lines into a list
+
+            for value in lines:
+                kv_pairs = value.split()
+                # This step assumes that no non-`voting` section in the config file
+                # uses the reference_fee key-value pair.
+                if "reference_fee" in kv_pairs:
+                    expected_fee = kv_pairs[2]
+                    break
+
+        self.assertEqual(fee, expected_fee)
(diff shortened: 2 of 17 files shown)


tests/integration/sugar/test_transaction.py
```diff
@@ -431,3 +436,32 @@ async def test_sign_and_submit(self, client):
...
+                    issuer="rweYz56rfmQ98cAdRaeTxQS9wVMGnrdsFp",
+                    value="0.0000002831214446",
+                ),
+            ),
+            client,
+        )
+
+        # If FeeSettings ledger object is updated, the below assert needs to be changed
+        # correspondingly
+        self.assertEqual(fee, "200")
```
achowdhry-ripple, line 467: nit: could we use the reference_fee variable from the config file here so that only that file needs to be updated in the future?

xrpl/asyncio/transaction/main.py
```diff
@@ -444,10 +443,10 @@ async def _calculate_fee_per_transaction_type(
         The expected Transaction fee in drops
     """
     # Reference Transaction (Most transactions)
-    if client is None:
-        net_fee = 10  # 10 drops
-    else:
-        net_fee = int(await get_fee(client))  # Usually 0.00001 XRP (10 drops)
+
+    net_fee = int(
+        await get_fee(client)
```
achowdhry-ripple, line 448: thought we should default to 200 drops always? Is the new recommendation that we don't have a default fee anymore?

achowdhry-ripple (approved):
LGTM - curious why we no longer want a default fee but otherwise looks good!

xrpl/transaction/__init__.py
```diff
@@ -1,9 +1,11 @@
 """Methods for working with transactions on the XRP Ledger."""
+
 from xrpl.asyncio.transaction import (
     XRPLReliableSubmissionException,
     transaction_json_to_binary_codec_form,
 )
 from xrpl.transaction.main import (
+    _calculate_fee_per_transaction_type,
```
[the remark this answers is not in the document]
ckeshava: This import is necessary for integration testing purposes. This method is used for internal purposes only.

tests/integration/sugar/test_transaction.py
ckeshava: completed in 16099ec2
I tried to use external libraries like `toml` and `configparser`. Thy could not parse the `.ci-config/rippled.cfg` file. Let me know if I can accomplish this without manually reading the file contents.

xrpl/asyncio/transaction/main.py
ckeshava: No, we do not have a default value. I used 200 drops as an example value, it was chosen arbitrarily.

We will know the transaction fees once all the validators have reached a compromise. Here is more detail on the Fee Voting process: https://xrpl.org/docs/concepts/consensus-protocol/fee-voting#voting-process

.github/workflows/snippet_test.yml
```diff
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 env:
-  POETRY_VERSION: 1.4.2
+  POETRY_VERSION: 1.8.3
```
pdp2121, line 10: Why do we need version bump here?

ckeshava: poetry released version `1.8.3` 6 months ago. I believe it is stable enough for our adoption: https://github.com/python-poetry/poetry/releases

Initially, I assumed the older version was causing a lint error, but the poetry version is not related to the linter.

I can revert back the poetry-related changes, if required.

pdp2121: No issue at all. Just curious since poetry.lock already use 1.8.3 but I guess it doesn't update automatically in all places.
https://github.com/XRPLF/xrpl-py/blob/main/poetry.lock#L1

pdp2121 (approved):

xrpl/transaction/__init__.py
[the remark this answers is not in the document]
mvadari: In Python this isn't necessary and honestly is still bad practice. THe better route is to fix the integration tests to support this, which was done in #793.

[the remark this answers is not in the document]
ckeshava: @mvadari  Hmm, so you'd recommend exporting a non-private method? One without a leading `_` character?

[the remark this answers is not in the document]
mvadari: No, it's not useful for external people, so it shouldn't be exported or have a sync version. See what I did in the PR linked above (I can split it out into a separate PR if you'd prefer).

PR #763, opened 2024-10-30
Add API key support for JSON-RPC requests

joshuahamsa:
This change allows JsonRpcBase to include custom HTTP headers when connecting to XRPL nodes that require authentication or metadata. The `__init__` method now accepts a headers argument, which is stored as an attribute and merged with any per-request headers passed to `_request_impl`.
	•	Updated the `_request_impl` method to merge default and request-specific headers.
	•	Added a headers parameter to the `__init__` method for flexible instantiation.
	•	This change was tested by connecting to a private XRPL server that required authentication using a custom header.

This PR introduces support for custom header injection in the `JsonRpcBase` class. An optional headers parameter has been added to the `__init__` method, allowing users to specify HTTP headers (such as API keys, Dhali payment-claims, or other authentication schemes) to include in each `JSON-RPC` request. The `_request_impl` method also accepts an optional headers argument to override or add headers on a per-request basis.

This makes it easier to integrate `xrpl-py` with private infrastructure providers and secure environments, while maintaining full backward compatibility.

### Context of Change

This change is driven by the increasing use of XRPL infrastructure that relies on API key or header-based authentication—such as rate-limited services, Dhali monetization, or internal infrastructure. Previously, `JsonRpcBase` did not support adding HTTP headers, limiting its usefulness in many production environments.

By making this update in the shared `json_rpc_base.py` layer, all inheriting classes benefit from the new feature. It allows for greater flexibility without impacting users who do not require any headers. Additionally, the per-request header support allows for advanced use cases like dynamic tokens, session headers, or monetized access (e.g. via Dhali).

### Type of Change

- [x] Yes

## Test Plan

The change can be tested by:
	•	Connecting to a private XRPL server that requires authentication using a custom header and verifying that requests are processed successfully.
	•	Ensuring compatibility with public nodes by omitting the headers parameter, confirming that no issues arise when using unauthenticated connections.
	•	Passing request-specific headers to `_request_impl` and verifying that they are correctly merged with any global headers.
	•	Confirming that all existing requests function identically when headers are not used.

## Future Tasks
Future tasks related to this PR could include:
	•	Elevating header support to the top-level `Client` class to ensure WebSocket clients can also benefit from custom headers.
	•	Adding support for header presets or strategies (e.g. `BearerTokenAuth`, `DhaliHeaderProvider`) to reduce boilerplate.
	•	Adding additional test coverage in CI for both with-header and without-header use cases.


mvadari:
To make this a bit more general/future-proof, I would write this to allow any headers in the init (and also in the request method, in case you want to include some headers just for some requests).

ckeshava:
@joshuahamsa  Is the `GenericRequest` not helpful for this case? It allows for a generic dictionary of key-value pairs in the request. Here's the [source code](https://github.com/XRPLF/xrpl-py/blob/f9bae8c341f3a23b5f86757c06d1c77ce977b330/xrpl/models/requests/generic_request.py#L31)

mvadari:
> @joshuahamsa  Is the `GenericRequest` not helpful for this case? It allows for a generic dictionary of key-value pairs in the request. Here's the [source code](https://github.com/XRPLF/xrpl-py/blob/f9bae8c341f3a23b5f86757c06d1c77ce977b330/xrpl/models/requests/generic_request.py#L31)

Headers aren't included in `GenericRequest`, they're a separate field.

ckeshava:
> > @joshuahamsa  Is the `GenericRequest` not helpful for this case? It allows for a generic dictionary of key-value pairs in the request. Here's the [source code](https://github.com/XRPLF/xrpl-py/blob/f9bae8c341f3a23b5f86757c06d1c77ce977b330/xrpl/models/requests/generic_request.py#L31)
> 
> Headers aren't included in `GenericRequest`, they're a separate field.

I understand that, but `GenericRequest` accepts a `**kwargs` input field. This field could theoretically accommodate any header key-value pairs.

The API keys appear to be required at the POST request construction.

```
api_keys: Optional[str] = None
if request.method == RequestMethod.GENERIC_REQUEST and request.api_key:
    api_keys = request.api_keys

async with AsyncClient(timeout=timeout) as http_client:
            response = await http_client.post(
                self.url,
                json=request_to_json_rpc(request),
                headers=api_keys,  # Include the headers with the optional API key
            )
```

Alternatively, it would be appropriate to make these changes at the top-most `Client` class. If a websocket connection needs to specify API-keys, this PR change cannot accommodate such use cases.

mvadari:
> > Headers aren't included in `GenericRequest`, they're a separate field.
> 
> I understand that, but `GenericRequest` accepts a `**kwargs` input field. This field could theoretically accommodate any header key-value pairs.
> 
> The API keys appear to be required at the POST request construction.
> 
> ```
> api_keys: Optional[str] = None
> if request.method == RequestMethod.GENERIC_REQUEST and request.api_key:
>     api_keys = request.api_keys
> 
> async with AsyncClient(timeout=timeout) as http_client:
>             response = await http_client.post(
>                 self.url,
>                 json=request_to_json_rpc(request),
>                 headers=api_keys,  # Include the headers with the optional API key
>             )
> ```

That wouldn't be very developer-friendly. Why should headers be limited to just the `GenericRequest` model, and not allowed in other requests? `GenericRequest` is only supposed to be for requests that we don't have models for, such as admin methods - which you're highly unlikely to be using an API key for.

> Alternatively, it would be appropriate to make these changes at the top-most `Client` class. If a websocket connection needs to specify API-keys, this PR change cannot accommodate such use cases.

Agreed, but adding it to just JSONRPC clients would be a good start.

ckeshava:
Hmm, I assumed that this was a use case for admin-methods. rippled operators could provide a wrapper with API-keys for their users to access admin methods. But I agree with you, other `Request`'s should also provide support for this header.

desimmons:
For what it's worth, we are very interested in this feature for the purpose of enabling RPC clients to pay for access to XRPL nodes using Dhali.

As others have mentioned above, generic headers would be desirable. For Dhali, the RPC client would have the `payment-claim` header injected into it.

ckeshava:
Are these headers not useful in `Websocket` Subscriptions? If so, we should consider propagating these changes up in the class inheritance hierarchy.

desimmons:
We don't currently support websockets connection. However, we do intend to. So this would be important for us.

xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -40,10 +52,18 @@ async def _request_impl(
 
         :meta private:
         """
+        # Prepare headers, including the API key if it’s available
+        headers = {"Content-Type": "application/json"}
+        if self.api_key:
+            headers[
+                "Authorization"
+            ] = f"Bearer {self.api_key}"  # Adjust key name if necessary
+
```
desimmons, line 61: Rather than having this as an injectable key, just for `Authorisation: Bearer ...`, could it be a dict-like object so that we can support the injection of arbitrary headers?

Our usecase is supporting the monetisation of XRPL-infra like [xrplcluster](https://xrpl.org/docs/tutorials/public-servers#commercial).

mvadari:
@joshuahamsa any update on this?

joshuahamsa:
@mvadari I just haven’t circled back to it. I’ll try to make some time to follow up on it.

xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -21,16 +22,36 @@ class JsonRpcBase(Client):
...
+        headers: Optional[Dict[str, str]] = None,
     ) -> Response:
         """
         Base ``_request_impl`` implementation for JSON RPC.
 
         Arguments:
             request: An object representing information about a rippled request.
             timeout: The duration within which we expect to hear a response from the
-            rippled validator.
+                rippled validator.
```
mvadari, line 53:
```suggestion
                rippled server.
```

xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -53,4 +82,4 @@ async def _request_impl(
                         "error": response.status_code,
                         "error_message": response.text,
                     }
-                )
+                ) from None
```
mvadari, line 85: Why `from None` here?

xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -3,6 +3,7 @@
 from __future__ import annotations
```
mvadari, line 3: You may want to update the client base type and cover websockets too (cc @desimmons)

Please also update the changelog

xrpl/asyncio/clients/json_rpc_base.py
joshuahamsa: @mvadari a nitpick from Coderabbit suggested to explicitly suppress the traceback from the original `JSONDecodeError` by using `from None`, presumably to keeps the error output clean and intentional.

xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -21,16 +22,37 @@ class JsonRpcBase(Client):
...
+        url: str,
+        *,
+        headers: Optional[Dict[str, str]] = None,
+    ) -> None:
+        """
+        Initializes a new JsonRpcBase client.
+
+        Arguments:
+            url: The URL of the XRPL node to connect to.
+            headers: Optional default headers for all requests (e.g. API key or Dhali payment-claim).
```
ckeshava, line 36: Could you give me some examples of what could constitute the headers? Do you expect users to send confidential/sensitive information via these headers?

1. I'm trying to understand if these headers are end-to-end encrypted during network communications.
2. If people are using `xrpl-py` as non-admin applications on a computer, can debugging/instrumentation/sniffing tools read this section of Python's in-memory variables? If so, these might be visible to super-users of the system.

Issue #768, opened 2024-11-04
Upgrade the minimum supported Python version to 3.9


ckeshava:
The End-Of-Life of Python versions can be found here: https://devguide.python.org/versions/

Python 3.8 is past its supported timeframe.

ckeshava:
It appears that there is no urgency to upgrade to a later Python version. Hence closing this issue.

PR #769, opened 2024-11-04
Upgrade the minimum supported version to Python 3.9

ckeshava:
This fixes https://github.com/XRPLF/xrpl-py/issues/768.

### Type of Change

- [ ] Documentation Updates
- [ ] Release
There is no change in functionality. But older versions of Python (3.8 and lower) are no longer tested actively.

- [x] Yes

## Test Plan

The existing unit and integration tests must suffice.


ckeshava:
@mvadari  Since the EOL is already completed, Python 3.8 will not receive any bug-fixes. Am I understanding that correctly?

If it will not be actively maintained, will the software be reliable?

Besides, dependencies like `black` have stopped supporting Python 3.8. Check the third-bullet-point here: https://github.com/psf/black/releases/tag/24.10.0

mvadari:
> @mvadari Since the EOL is already completed, Python 3.8 will not receive any bug-fixes. Am I understanding that correctly?
> 
> If it will not be actively maintained, will the software be reliable?

Plenty of people use old versions of code. Some people still use Python 2.x. We shouldn't force them to upgrade unless we need to (e.g. if an actual dependency upgrades).

> Besides, dependencies like `black` have stopped supporting Python 3.8. Check the third-bullet-point here: https://github.com/psf/black/releases/tag/24.10.0

Black is only relevant to contributors, not users. We can run the linting on just >= 3.9.

ckeshava:
@mvadari Alright, if that's the case, I can close this PR

PR #770, opened 2024-11-08
fix: Update default cryptographic algorithm in `Wallet` class methods

ckeshava:

This change is related to this PR: https://github.com/XRPLF/xrpl-py/pull/748
I've addressed my comment (https://github.com/XRPLF/xrpl-py/pull/748#issuecomment-2438533913) in the current PR.

@anissa-ripple  @justinr1234  @dangell7  please review the changes in this PR

### Context of Change
Use `ed25519` as the default cryptographic algorithm in `Wallet.from_secret_numbers` method. This ensures consistency amongst all the methods of the `Wallet` class.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes


ckeshava:
Note to the reviewers: I've made an attempt to update the docs to better communicate these changes here: https://github.com/XRPLF/xrpl-dev-portal/pull/2844

mvadari:
This is a breaking change.

justinr1234:
@ckeshava would need to put this into a major version release --- and move to the breaking change docs

ckeshava:
@justinr1234  I've moved the update-message into the "Breaking Changes" section. There is already one other breaking change as well.

How do I indicate that the next release needs to be a major version upgrade? Since we are removing support for Python 3.7, I suspect we need to do a major version upgrade anyway.

xrpl/wallet/main.py
```diff
@@ -216,9 +215,11 @@ def from_secret_numbers(
             master_address: Include if a Wallet uses a Regular Key Pair. It must be
                 the master address of the account. The default is `None`.
             algorithm: The digital signature algorithm to generate an address for.
-                The default is `SECP256K1
-                <https://xrpl.org/cryptographic-keys.html#secp256k1-key-derivation>`_
-                (XUMM standard as of December 2022).
+                The default is ED25519. Docs:
+                https://xrpl.org/docs/concepts/accounts/cryptographic-keys#ed25519-key-derivation
```
khancode, line 219: Let's keep the link format as it was before.

xrpl/wallet/main.py
```diff
@@ -89,7 +88,7 @@ def __init__(
                     + wallet_algorithm
                     + "\nError message: "
                     + str(e)
-                )
+                ) from e
```
khancode, line 91: Is this line not needed?

ckeshava: ok, I have removed it 👍

xrpl/wallet/main.py
ckeshava: Ok, fixed in the latest commit.

khancode:
> @justinr1234 I've moved the update-message into the "Breaking Changes" section. There is already one other breaking change as well.
> 
> How do I indicate that the next release needs to be a major version upgrade? Since we are removing support for Python 3.7, I suspect we need to do a major version upgrade anyway.

This is sufficient. When we do the next release it's be a major version bump.

khancode (approved):

pdp2121 (approved):

PR #802, opened 2025-02-05
Support AMMClawback amendment

ckeshava:

Documentation for the said amendment: https://xrpl.org/docs/references/protocol/transactions/types/ammclawback
C++ implementation PR: https://github.com/XRPLF/rippled/pull/5142/files

### Context of Change

This PR implements the xrpl-py support for the said amendment.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

### DIFF
--- a/tests/integration/it_utils.py
+++ b/tests/integration/it_utils.py
@@ -315,6 +315,7 @@ def _get_non_decorator_code(function):
 
 def create_amm_pool(
     client: SyncClient = JSON_RPC_CLIENT,
+    issuer_account_enable_trustline_clawback: bool = False,
 ) -> Dict[str, Any]:
     issuer_wallet = Wallet.create()
     fund_wallet(issuer_wallet)
@@ -331,6 +332,16 @@ def create_amm_pool(
         issuer_wallet,
     )
 
+    # The below flag is required for AMMClawback tests
+    if issuer_account_enable_trustline_clawback:
+        sign_and_reliable_submission(
+            AccountSet(
+                account=issuer_wallet.classic_address,
+                set_flag=AccountSetAsfFlag.ASF_ALLOW_TRUSTLINE_CLAWBACK,
+            ),
+            issuer_wallet,
+        )
+
     sign_and_reliable_submission(
         TrustSet(
             account=lp_wallet.classic_address,
@@ -382,11 +393,13 @@ def create_amm_pool(
         "asset": asset,
         "asset2": asset2,
         "issuer_wallet": issuer_wallet,
+        "lp_wallet": lp_wallet,
     }
 
 
 async def create_amm_pool_async(
     client: AsyncClient = ASYNC_JSON_RPC_CLIENT,
+    issuer_account_enable_trustline_clawback: bool = False,
 ) -> Dict[str, Any]:
     issuer_wallet = Wallet.create()
     await fund_wallet_async(issuer_wallet)
@@ -403,6 +416,16 @@ async def create_amm_pool_async(
         issuer_wallet,
     )
 
+    # The below flag is required for AMMClawback tests
+    if issuer_account_enable_trustline_clawback:
+        await sign_and_reliable_submission_async(
+            AccountSet(
+                account=issuer_wallet.classic_address,
+                set_flag=AccountSetAsfFlag.ASF_ALLOW_TRUSTLINE_CLAWBACK,
+            ),
+            issuer_wallet,
+        )
+
     await sign_and_reliable_submission_async(
         TrustSet(
             account=lp_wallet.classic_address,
@@ -454,6 +477,7 @@ async def create_amm_pool_async(
         "asset": asset,
         "asset2": asset2,
         "issuer_wallet": issuer_wallet,
+        "lp_wallet": lp_wallet,
     }
 
 
--- a/tests/integration/it_utils.py
+++ b/tests/integration/it_utils.py
@@ -315,7 +315,7 @@ def _get_non_decorator_code(function):
 
 def create_amm_pool(
     client: SyncClient = JSON_RPC_CLIENT,
-    issuer_account_enable_trustline_clawback: bool = False,
+    enable_amm_clawback: bool = False,
 ) -> Dict[str, Any]:
     issuer_wallet = Wallet.create()
     fund_wallet(issuer_wallet)
@@ -333,7 +333,7 @@ def create_amm_pool(
     )
 
     # The below flag is required for AMMClawback tests
-    if issuer_account_enable_trustline_clawback:
+    if enable_amm_clawback:
         sign_and_reliable_submission(
             AccountSet(
                 account=issuer_wallet.classic_address,
@@ -399,7 +399,7 @@ def create_amm_pool(
 
 async def create_amm_pool_async(
     client: AsyncClient = ASYNC_JSON_RPC_CLIENT,
-    issuer_account_enable_trustline_clawback: bool = False,
+    enable_amm_clawback: bool = False,
 ) -> Dict[str, Any]:
     issuer_wallet = Wallet.create()
     await fund_wallet_async(issuer_wallet)
@@ -417,7 +417,7 @@ async def create_amm_pool_async(
     )
 
     # The below flag is required for AMMClawback tests
-    if issuer_account_enable_trustline_clawback:
+    if enable_amm_clawback:
         await sign_and_reliable_submission_async(
             AccountSet(
                 account=issuer_wallet.classic_address,
--- a/tests/integration/transactions/test_amm_clawback.py
+++ b/tests/integration/transactions/test_amm_clawback.py
@@ -1,0 +1,73 @@
+from tests.integration.integration_test_case import IntegrationTestCase
+from tests.integration.it_utils import (
+    create_amm_pool_async,
+    fund_wallet_async,
+    sign_and_reliable_submission_async,
+    test_async_and_sync,
+)
+from tests.integration.reusable_values import WALLET
+from xrpl.models.amounts.issued_currency_amount import IssuedCurrencyAmount
+from xrpl.models.currencies.issued_currency import IssuedCurrency
+from xrpl.models.currencies.xrp import XRP
+from xrpl.models.requests.amm_info import AMMInfo
+from xrpl.models.response import ResponseStatus
+from xrpl.models.transactions import (
+    AccountSet,
+    AccountSetAsfFlag,
+    AMMCreate,
+    AMMDeposit,
+)
+from xrpl.models.transactions.amm_clawback import AMMClawback
+from xrpl.models.transactions.amm_deposit import AMMDepositFlag
+from xrpl.wallet import Wallet
+
+
+class TestAMMClawback(IntegrationTestCase):
+    @test_async_and_sync(globals())
+    async def test_positive_workflow(self, client):
+        amm_pool = await create_amm_pool_async(
+            client, issuer_account_enable_trustline_clawback=True
+        )
+
+        # Asset-1 is XRP, Asset-2 is an IssuedCurrency titled "USD"
+        # The Issuer of Asset-2 is the issuer_wallet
+        # For the purposes of this test, the issuer_wallet has set the
+        # Allow Trust Line Clawback flag
+        issuer_wallet = amm_pool["issuer_wallet"]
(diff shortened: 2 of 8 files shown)


tests/integration/it_utils.py
```diff
@@ -315,6 +315,7 @@ def _get_non_decorator_code(function):
 
 def create_amm_pool(
     client: SyncClient = JSON_RPC_CLIENT,
+    issuer_account_enable_trustline_clawback: bool = False,
```
mvadari, line 318: This is a very verbose field name - why not just call it `enable_clawback`?

xrpl/models/transactions/amm_clawback.py
```diff
@@ -0,0 +1,106 @@
...
+    Specifies the asset that the issuer wants to claw back from the AMM pool. In JSON,
+    this is an object with currency and issuer fields. The issuer field must match with
+    Account.
+    """
+
+    asset2: Currency = REQUIRED  # type: ignore
+    """
+    Specifies the other asset in the AMM's pool. In JSON, this is an object with
+    currency and issuer fields (omit issuer for XRP).
+    """
```
khancode, line 73: Move optional fields after required.
```suggestion
    asset: IssuedCurrency = REQUIRED  # type: ignore
    """
    Specifies the asset that the issuer wants to claw back from the AMM pool. In JSON,
    this is an object with currency and issuer fields. The issuer field must match with
    Account.
    """

    asset2: Currency = REQUIRED  # type: ignore
    """
    Specifies the other asset in the AMM's pool. In JSON, this is an object with
    currency and issuer fields (omit issuer for XRP).
    """
    
    amount: Optional[IssuedCurrencyAmount] = None
    """
    The maximum amount to claw back from the AMM account. The currency and issuer
    subfields should match the Asset subfields. If this field isn't specified, or the
    value subfield exceeds the holder's available tokens in the AMM, all of the
    holder's tokens are clawed back.
    """
```

tests/integration/it_utils.py
```diff
@@ -331,6 +332,16 @@ def create_amm_pool(
...
+    # The below flag is required for AMMClawback tests
+    if issuer_account_enable_trustline_clawback:
+        sign_and_reliable_submission(
+            AccountSet(
+                account=issuer_wallet.classic_address,
+                set_flag=AccountSetAsfFlag.ASF_ALLOW_TRUSTLINE_CLAWBACK,
+            ),
+            issuer_wallet,
+        )
+
```
khancode, line 344: Rather than doing a second AccountSet transaction, we still do one AccountSet and OR `ASF_ALLOW_TRUSTLINE_CLAWBACK` flag with `ASF_DEFAULT_RIPPLE` when we want to enable clawback.

tests/integration/it_utils.py
```diff
@@ -403,6 +416,16 @@ async def create_amm_pool_async(
...
 
+    # The below flag is required for AMMClawback tests
+    if issuer_account_enable_trustline_clawback:
+        await sign_and_reliable_submission_async(
+            AccountSet(
+                account=issuer_wallet.classic_address,
+                set_flag=AccountSetAsfFlag.ASF_ALLOW_TRUSTLINE_CLAWBACK,
+            ),
+            issuer_wallet,
+        )
```
khancode, line 427: [ditto](https://github.com/XRPLF/xrpl-py/pull/802/files#r1944973102)

tests/integration/transactions/test_amm_clawback.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # "holder" account deposits both assets into the AMM pool
+        # Deposit assets into AMM pool
+        amm_deposit = AMMDeposit(
+            account=holder_wallet.address,
+            asset=IssuedCurrency(
+                currency="USD",
+                issuer=issuer_wallet.address,
+            ),
+            asset2=XRP(),
```
khancode, line 37: You can reference `asset` and `asset2` from `amm_pool`.

tests/integration/transactions/test_amm_clawback.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # Clawback one of the assets from the AMM pool
+        amm_clawback = AMMClawback(
+            account=issuer_wallet.address,
+            holder=holder_wallet.address,
+            asset=IssuedCurrency(
+                currency="USD",
+                issuer=issuer_wallet.address,
+            ),
+            asset2=XRP(),
```
khancode, line 58: ditto: reuse `asset`, `asset2` from `ammPool`

xrpl/models/transactions/amm_clawback.py
```diff
@@ -0,0 +1,106 @@
...
+    Account field. If this flag isn't enabled, the issuer claws back the specified
+    amount of Asset, while a corresponding proportion of Asset2 goes back to the Holder.
+    """
+
+    TF_CLAW_TWO_ASSETS = 0x00000001
+
+
+class AMMClawbackFlagInterface(FlagInterface):
+    """
+    Claw back the specified amount of Asset, and a corresponding amount of Asset2 based
```
achowdhry-ripple, line 34: nit: is it necessary to duplicate the paragraph comment thats right above?

xrpl/models/transactions/amm_clawback.py
```diff
@@ -0,0 +1,106 @@
...
+        if self.account != self.asset.issuer:
+            errors += (
+                "Asset.issuer and AMMClawback transaction sender must be identical."
+            )
+
+        if self.amount is not None and (
+            self.amount.issuer != self.asset.issuer
+            or self.amount.currency != self.asset.currency
+        ):
+            errors += "Amount issuer/currency subfield does not match Asset field. "
```
achowdhry-ripple, line 104:
```suggestion
            errors += "Amount.issuer and Amount.currency must match corresponding Asset fields."
```

tests/integration/transactions/test_amm_clawback.py
```diff
@@ -0,0 +1,63 @@
...
+from xrpl.models.currencies.issued_currency import IssuedCurrency
+from xrpl.models.currencies.xrp import XRP
+from xrpl.models.transactions import AMMDeposit
+from xrpl.models.transactions.amm_clawback import AMMClawback
+from xrpl.models.transactions.amm_deposit import AMMDepositFlag
+
+
+class TestAMMClawback(IntegrationTestCase):
+    @test_async_and_sync(globals())
+    async def test_positive_workflow(self, client):
```
achowdhry-ripple, line 17: Is it worth having a second flow here for two asset clawback?

xrpl/models/transactions/amm_clawback.py
ckeshava: I don't mind removing it. But it comes down to individual reviewers preferences

tests/integration/transactions/test_amm_clawback.py
ckeshava: I have included a unit test for the two-asset clawback

ckeshava: much of the test-preparation remains identical to this case, except for the flag parameters to the `AMMClawback` transaction.

tests/integration/transactions/test_amm_clawback.py
ckeshava: The first `asset` cannot be XRP. rippled will throw a `temMALFORMED` error. However, `asset2` can be XRP. Source: https://github.com/XRPLF/rippled/blob/fb3713bc2507d577aebe56fbc38872346811d76d/src/xrpld/app/tx/detail/AMMClawback.cpp#L63

The `create_amm_pool_async` utility method returns `asset: XRP()`. Hence, I can't directly use this value for this test.

If I refactor this method, I'm not sure if other AMM tests break.

tests/integration/it_utils.py
ckeshava: I don't want users to be confused with the `Clawback` amendment. Besides, modern IDEs auto-complete the variable names. If we have verbose names, is that a problem?

tests/integration/it_utils.py
ckeshava: That doesn't work. Here is the error:
```
File "/Users/ckeshavabs/xrpl-py/xrpl/models/base_model.py", line 228, in _from_dict_single_param
    raise XRPLModelException(error_message)
xrpl.models.exceptions.XRPLModelException: set_flag expected a typing.Optional[xrpl.models.transactions.account_set.AccountSetAsfFlag], received a <class 'int'>
```

The `AccountSet.set_flag` field accepts the enum `AccountSetAsfFlag`, not an `int`

xrpl/models/transactions/amm_clawback.py
ckeshava: a37cd1bf

ckeshava: I had to update the unit test file as well. I did it in the above commit

xrpl/models/transactions/amm_clawback.py
achowdhry-ripple: Personally I'd remove it for cleanliness, too much repeated content right next to each other with little gain IMO. I don't feel strongly though if anyone disagrees here

tests/integration/transactions/test_amm_clawback.py
achowdhry-ripple: sounds good

tests/integration/it_utils.py
achowdhry-ripple: I'd agree to simplify the name. In my opinion verbose names reduce code readability, and make it harder for new community devs to go through and understand whats going on. If you want to avoid confusion maybe you could name it` enable_amm_clawback`?

mvadari: Also, if you have verbose names, then you end up having lines of code that need to be split up between like 10 lines of space due to line length limits.

xrpl/models/transactions/amm_clawback.py
mvadari: I think it's useful for docs purposes to have it in both places.

tests/integration/it_utils.py
khancode: Why not just use `AccountSet.flags` instead?

tests/integration/it_utils.py
khancode: I agree with @achowdhry-ripple and @mvadari for the same reasons.

tests/integration/it_utils.py
ckeshava: https://xrpl.org/docs/references/protocol/transactions/types/accountset#accountset-flags

Do you know if `asf` flags can be passed into the `tf` flags input field? Based on the docs, they have a separate `set_flag` and `clear_flag` field for this purpose

tests/integration/it_utils.py
ckeshava: d9ecf89c

tests/integration/it_utils.py
khancode: Oh that's inconvenient. Why can't they be included all in one transaction rather than separate ones? I'm not sure how rippled implemented this.

mvadari: rippled ran out of bitwise flags, lol

ckeshava: I share your frustrations @khancode 
@mvadari  the flags field is a `std::uint32_t` right? We should be able to fit in `2^32 -1` values right? Am I missing something?

ckeshava: I'm referring to `sfFlags`

mvadari: They're flags, so they have to be distinct bitwise. So you've only got 32, and some of that space has to be reserved for global transaction flags.

mvadari: https://xrpl.org/docs/references/protocol/transactions/common-fields#flags-field

khancode: Ah that makes sense.. Thanks for clarifying this 😄

tests/integration/transactions/test_amm_clawback.py
khancode: Ah yes, you're right. Disregard my comment then.

khancode (approved):
LGTM, great work 👍

achowdhry-ripple (approved):

ckeshava:
Thank you for all the reviews @khancode @achowdhry-ripple @mvadari

PR #813, opened 2025-03-04
fix: update error msg for MPT float inputs; fix #796

### LINKED ISSUE #796: Error: MPTAmount value field rejects valid float values (e.g., 10.01)
When I attempt to send an MPT payment transaction with an MPTAmount value of 10.01 (a float), the SDK throws the following error:

```XRPLBinaryCodecException(f"{mpt_value} is an invalid MPT amount.")
xrpl.core.binarycodec.exceptions.XRPLBinaryCodecException: Error processing Amount: 10.01 is an invalid MPT amount.
```

This behavior is unexpected as the SDK type hints suggest that float values should be acceptable for the `MPTAmount.value` field.

Expected Behavior:

The SDK should successfully process float values for the MPTAmount value field in MPT payment transactions.

Steps to Reproduce:

* Create an MPT payment transaction with an MPTAmount value of 10.01 (or any other float value with decimal places).
* Attempt to send the transaction using the SDK.

ckeshava:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes
There is no change in behavior. The error message is updated for easier remediation.

### DIFF
--- a/xrpl/core/binarycodec/types/amount.py
+++ b/xrpl/core/binarycodec/types/amount.py
@@ -125,7 +125,9 @@ def verify_mpt_value(mpt_value: str) -> None:
     """
     # Contains no decimal point
     if not _contains_decimal(mpt_value):
-        raise XRPLBinaryCodecException(f"{mpt_value} is an invalid MPT amount.")
+        raise XRPLBinaryCodecException(
+            f"{mpt_value} contains fraction value. Please use str type to represent it."
+        )
 
     decimal = None
     try:
--- a/tests/unit/core/binarycodec/types/test_amount.py
+++ b/tests/unit/core/binarycodec/types/test_amount.py
@@ -171,6 +171,16 @@ class TestAmount(TestSerializedType):
             amount_object = amount.Amount.from_value(json)
             self.assertEqual(amount_object.to_hex(), serialized)
 
+    def test_fractional_MPT_invalid_representation(self):
+        self.assertRaises(
+            XRPLBinaryCodecException,
+            amount.Amount.from_value,
+            {
+                "mpt_issuance_id": "0000012FFD9EE5DA93AC614B4DB94D7E0FCE415CA51BED47",
+                "value": 1.1,
+            },
+        )
+
     def test_to_json_issued_currency(self):
         for json, serialized in IOU_CASES:
             parser = BinaryParser(serialized)


achowdhry-ripple (approved):

xrpl/models/amounts/mpt_amount.py
```diff
@@ -24,7 +24,7 @@ class MPTAmount(BaseModel):
     :meta hide-value:
     """
 
-    value: Union[str, int, float] = REQUIRED  # type: ignore
+    value: str = REQUIRED  # type: ignore
```
khancode, line 27: I don't think this should be changed at the xrpl package level. Doesn't binary-codec take `int` or `float` and eventually convert it to string? cc @mvadari

mvadari: I didn't review the MPT code - here it is, for reference: https://github.com/XRPLF/xrpl-py/blob/main/xrpl/core/binarycodec/types/amount.py#L304

ckeshava: 1. We have discrepancies in handling the translation of rippled's `std::uint64_t` type into a standard JSON type. For instance, `asset_price` (in `OracleSet` transaction) accepts an `int` whereas the MPTAmount.value field accepts `str` only. Mea culpa, I implemented the `PriceOracle` amendment. 

Is it acceptable to use `str` JSON type for all the `std::uint64_t` fields? Do we need to provide any int, float interfaces for these `sf<Fieldname>` values?

2. As shown in test cases inside `test_amount.py` file, we need to support scientific notation syntax in the `MPTAmount.value` fields. At the moment, scientific notation does not work for serialization.

ckeshava: @khancode 

https://github.com/XRPLF/xrpl-py/blob/13fdbd8003ddf2aa60e398f8063fed51d138e020/xrpl/core/binarycodec/types/amount.py#L128

It is not possible to do a string-based search for `.` character in a `float, int` input. Do MPT amounts not allow for fractional portion? 

It is possible to extend this check for int, float types by processing it through the `Decimal` library. We can extract the `exponent` portion of the number and determine the validity of the value.

This was one of the differences I found between serialization of IOU and MPT values. I'd prefer to retain the existing functionality instead of reducing the API surface area (in terms of allowed types for `value` field). What do you think ?

khancode: @ckeshava You can verify on the docs if it accepts it. And if you have an idea for a solution, you can verify it by testing your changes.

ckeshava: I'll address that in a separate PR. I've created an issue to track it: https://github.com/XRPLF/xrpl-py/issues/816

xrpl/core/binarycodec/types/amount.py
```diff
@@ -125,7 +125,9 @@ def verify_mpt_value(mpt_value: str) -> None:
     """
     # Contains no decimal point
     if not _contains_decimal(mpt_value):
-        raise XRPLBinaryCodecException(f"{mpt_value} is an invalid MPT amount.")
+        raise XRPLBinaryCodecException(
+            f"{mpt_value} contains fraction value. Please use str type to represent it."
```
khancode, line 129:
```suggestion
            f"{mpt_value} is a decimal value. Please use a string type to represent it."
```

khancode (approved):

PR #814, opened 2025-03-04
Single Asset Vault XLS-65d

ckeshava:

CPP implementation for SAV: https://github.com/XRPLF/rippled/pull/5224
CPP implementation of the `STNumber` type: https://github.com/XRPLF/rippled/pull/5121/files

XLS Specification for this feature: https://github.com/XRPLF/XRPL-Standards/discussions/192
Please view the updates to the specification in this PR: https://github.com/XRPLF/XRPL-Standards/pull/275/files

### Context of Change

This PR adds the following transactions: `Vault<Create, Set, Deposit, Withdraw, Clawback, Delete>`
This PR also adds the models and the unit tests for the `VaultInfo` RPC. This is the reference commit: ([a1073b3](https://github.com/XRPLF/xrpl-py/pull/814/commits/a1073b357014d71c5b38e18e2f0d226099866e77)).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan
I don't see the value of writing unit tests for the transaction models. All the `tem` codes are caught in the `preflight` method of the relevant `Transactor` class. Additional latency to an already failing transaction should not affect user-experience.

Transaction-model unit tests add unnecessary code bloat and duplication. Certain transactions (like the ones related to `XChain` amendment) do not have unit tests associated with them. Furthermore, `requests` also do not have unit tests associated with their models.


xrpl/core/binarycodec/types/number.py
```diff
@@ -0,0 +1,275 @@
...
+        A bytes object containing the serialized integer
+    """
+    serialized_bytes = bytes()
+    serialized_bytes += (value >> 24 & 0xFF).to_bytes(1)
+    serialized_bytes += (value >> 16 & 0xFF).to_bytes(1)
+    serialized_bytes += (value >> 8 & 0xFF).to_bytes(1)
+    serialized_bytes += (value & 0xFF).to_bytes(1)
+
+    return serialized_bytes
+
```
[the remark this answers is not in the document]
ckeshava: every statement appends only one byte. Hence, there is no need to specify the endianness.
Furthermore, this is implemented in the big-endian notation as evidenced by the decreasing size of the right-shift mask.

xrpl/core/binarycodec/types/number.py
```diff
@@ -0,0 +1,275 @@
...
+    serialized_bytes += (value >> 48 & 0xFF).to_bytes(1)
+    serialized_bytes += (value >> 40 & 0xFF).to_bytes(1)
+    serialized_bytes += (value >> 32 & 0xFF).to_bytes(1)
+    serialized_bytes += (value >> 24 & 0xFF).to_bytes(1)
+    serialized_bytes += (value >> 16 & 0xFF).to_bytes(1)
+    serialized_bytes += (value >> 8 & 0xFF).to_bytes(1)
+    serialized_bytes += (value & 0xFF).to_bytes(1)
+
+    return serialized_bytes
+
```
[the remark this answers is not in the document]
ckeshava: every statement appends only one byte. Hence, there is no need to specify the endianness.
Furthermore, this is implemented in the big-endian notation as evidenced by the decreasing size of the right-shift mask.

tests/integration/transactions/test_sav.py
```diff
@@ -0,0 +1,135 @@
...
+        self.assertEqual(response.status, ResponseStatus.SUCCESS)
+        self.assertEqual(response.result["engine_result"], "tesSUCCESS")
+
+        # Verify the existence of the vault with account_objects RPC call
+        account_objects_response = await client.request(
+            AccountObjects(account=vault_owner.address, type=AccountObjectType.VAULT)
+        )
+        self.assertEqual(len(account_objects_response.result["account_objects"]), 1)
+
+        VAULT_ID = account_objects_response.result["account_objects"][0]["index"]
```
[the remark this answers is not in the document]
ckeshava: Instead of adding more code to this file, I'd prefer that it fails with an error. Exception handling is not critical because this is a test file

tests/unit/core/binarycodec/types/test_number.py
```diff
@@ -0,0 +1,42 @@
...
+
+        serialized_number = Number.from_value("1.456e-45")
+        self.assertEqual(serialized_number.to_json(), "1456000000000000e-60")
+
+        serialized_number = Number.from_value("0.456e34")
+        self.assertEqual(serialized_number.to_json(), "4560000000000000e18")
+
+        serialized_number = Number.from_value("4e34")
+        self.assertEqual(serialized_number.to_json(), "4000000000000000e19")
+
```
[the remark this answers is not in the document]
ckeshava: this question has been answered in the appropriate file.

xrpl/core/binarycodec/types/number.py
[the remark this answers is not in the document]
ckeshava: I didn't get any errors or warnings from the Python interpreter

tests/unit/core/binarycodec/types/test_number.py
```diff
@@ -0,0 +1,42 @@
...
+    def test_extreme_limits(self):
+        lowest_mantissa = "-9223372036854776"
+        serialized_number = Number.from_value(lowest_mantissa + "e3")
+        self.assertEqual(serialized_number.hex(), "FFDF3B645A1CAC0800000003")
+
+        highest_mantissa = "9223372036854776"
+        serialized_number = Number.from_value(highest_mantissa + "e3")
+        self.assertEqual(serialized_number.hex(), "0020C49BA5E353F700000003")
```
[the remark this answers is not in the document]
ckeshava: this method has been updated in recent commits.

tests/unit/core/binarycodec/types/test_number.py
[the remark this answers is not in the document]
ckeshava: thanks, I have updated the code

xrpl/core/binarycodec/types/number.py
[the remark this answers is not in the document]
ckeshava: thanks, I've updated the code

ckeshava:
Note: The integration tests will not pass until the C++ code is merged into the rippled:develop branch. After this, a new docker container needs to be generated for the said branch.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Improved validation for models to also check param types
+- Support for Single Asset Vault (XLS-65d)
```
achowdhry-ripple, line 15:
```suggestion
- Support for `Single Asset Vault` (XLS-65d)
```

.ci-config/rippled.cfg
```diff
@@ -196,6 +196,7 @@ fixEnforceNFTokenTrustline
 fixReducedOffersV2
 DeepFreeze
 PermissionedDomains
+SingleAssetVault
```
achowdhry-ripple, line 199: Is this all still a part of 2.3.0 (no new releases)?

xrpl/models/requests/ledger_entry.py
```diff
@@ -315,6 +316,8 @@ class LedgerEntry(Request, LookupByLedgerRequest):
     payment_channel: Optional[str] = None
```
achowdhry-ripple, line 316: how come there is no Vault class definition in this file?

xrpl/models/transactions/types/transaction_type.py
```diff
@@ -52,6 +52,12 @@ class TransactionType(str, Enum):
     SIGNER_LIST_SET = "SignerListSet"
     TICKET_CREATE = "TicketCreate"
     TRUST_SET = "TrustSet"
+    VAULT_CREATE = "VaultCreate"
+    VAULT_CLAWBACK = "VaultClawback"
+    VAULT_DEPOSIT = "VaultDeposit"
+    VAULT_DELETE = "VaultDelete"
```
achowdhry-ripple, line 58: tiny nit: flip these two for alphabetical order

xrpl/models/requests/ledger_entry.py
ckeshava: Hello @achowdhry-ripple , this was due to the state of the cpp implementation back in April 2025. I have updated the code in [f4830a0](https://github.com/XRPLF/xrpl-py/pull/814/commits/f4830a08b7241c3647f4b9b7e6b3ddca008576f4).

.ci-config/rippled.cfg
ckeshava: I have added an additional comment here: [f4830a0](https://github.com/XRPLF/xrpl-py/pull/814/commits/f4830a08b7241c3647f4b9b7e6b3ddca008576f4)

ckeshava:
@pdp2121 @achowdhry-ripple 
I have updated this PR to match the latest version of the feature specification. Please take a look.

xrpl/models/amounts/mpt_amount.py
```diff
@@ -49,3 +49,16 @@ def to_currency(self: Self) -> MPTCurrency:
...
+    """
+    This class represents an MPT issue. It is similar to the Issue class, but
+    it is used with MPT amounts.
+    """
+
+    mpt_issuance_id: str = REQUIRED  # type: ignore
+    """
+    mpt_issuance_id is a 192-bit concatenation of a 32-bit account sequence and a
+    160-bit account id.
+    """
```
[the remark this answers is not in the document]
ckeshava: Does the MPTIssue class need other aspects of the `BaseModel` and the `dataclass` ? The present set of requirements work correctly with the current design of the `MPTIssue` class.

xrpl/models/requests/vault_info.py
```diff
@@ -0,0 +1,31 @@
...
+    Information about a vault ledger-object can be fetched by providing either the
+    vault_id or both owner and seq values. Please check the documentation for more
+    details.
+    """
+
+    vault_id: Optional[str] = None
+    owner: Optional[str] = None
+    seq: Optional[int] = None
+
+    method: RequestMethod = field(default=RequestMethod.VAULT_INFO, init=False)
```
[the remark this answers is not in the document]
ckeshava: I agree with the intent of your suggestion. However, other `Request`s in the xrpl-py library do not perform any validation on the contents of the requests. Furthermore, the rippled node will return an appropriate error-code, should there be any errors in the formation of the request.

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,76 @@
...
+
+    domain_id: Optional[str] = None
+    """The PermissionedDomain object ID associated with the shares of this Vault."""
+
+    withdrawal_policy: Optional[int] = None
+    """Indicates the withdrawal strategy used by the Vault. The below withdrawal policy
+    is supported:
+
+    Strategy Name	           Value	      Description
+    strFirstComeFirstServe	   1	          Requests are processed on a first-come-first-
```
Patel-Raj11: Shouldn't this be `vaultStrategyFirstComeFirstServe` as per the XLS spec?

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,76 @@
...
+    assets_maximum: Optional[str] = None
+    """The maximum asset amount that can be held in a vault."""
+
+    mptoken_metadata: Optional[str] = None
+    """Arbitrary metadata about the share MPT, in hex format, limited to 1024 bytes."""
+
+    domain_id: Optional[str] = None
+    """The PermissionedDomain object ID associated with the shares of this Vault."""
+
+    withdrawal_policy: Optional[int] = None
```
Patel-Raj11: IMO having an enum would have a better user experience. I don't think they might be aware which integer value to pass in for `vaultStrategyFirstComeFirstServe`.

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,76 @@
...
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class VaultCreate(Transaction):
+    """The VaultCreate transaction creates a new Vault object."""
+
+    asset: Union[Currency, MPTIssue] = REQUIRED  # type: ignore
+    """The asset (XRP, IOU or MPT) of the Vault."""
+
+    data: Optional[str] = None
```
Patel-Raj11, line 64: Having a _get_errors function that enforces the following would be helpful.

1. Data to max 256 bytes
2. DomainID to adhere to Hash256
3. MPTokenMetadata to 1024 bytes.

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,76 @@
...
+    """Arbitrary Vault metadata, limited to 256 bytes."""
+
+    assets_maximum: Optional[str] = None
```
Patel-Raj11, line 67: Shouldn't this be `int` type and have a default value of 0 as per the spec?

xrpl/models/transactions/vault_set.py
```diff
@@ -0,0 +1,34 @@
...
+    vault_id: str = REQUIRED  # type: ignore
+    """The ID of the Vault to be modified. Must be included when updating the Vault."""
+
+    domain_id: Optional[str] = None
+    """The PermissionedDomain object ID associated with the shares of this Vault."""
+
+    data: Optional[str] = None
+    """Arbitrary Vault metadata, limited to 256 bytes."""
+
+    assets_maximum: Optional[str] = None
```
Patel-Raj11, line 33: Shouldn't this be `int` type?

xrpl/models/transactions/vault_set.py
```diff
@@ -0,0 +1,34 @@
...
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class VaultSet(Transaction):
+    """The VaultSet updates an existing Vault ledger object."""
+
+    vault_id: str = REQUIRED  # type: ignore
+    """The ID of the Vault to be modified. Must be included when updating the Vault."""
+
+    domain_id: Optional[str] = None
```
Patel-Raj11, line 27: These checks can be enforced here as well.

1. Data to max 256 bytes
2. DomainID to adhere to Hash256

xrpl/models/transactions/vault_delete.py
```diff
@@ -0,0 +1,22 @@
...
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class VaultDelete(Transaction):
+    """The VaultDelete transaction deletes an existing vault object."""
+
+    vault_id: str = REQUIRED  # type: ignore
```
Patel-Raj11, line 21: Hash256 check can be enforced here.

xrpl/models/transactions/vault_withdraw.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class VaultWithdraw(Transaction):
+    """The VaultWithdraw transaction withdraws assets in exchange for the vault's
+    shares.
+    """
+
+    vault_id: str = REQUIRED  # type: ignore
```
Patel-Raj11, line 23: 1. Hash256 check for vault_id can be enforced here.
2. Amount defaults to 0 as per the spec.

xrpl/models/requests/vault_info.py
```diff
@@ -0,0 +1,31 @@
...
+
+    All information retrieved is relative to a particular version of the ledger.
+
+    Information about a vault ledger-object can be fetched by providing either the
+    vault_id or both owner and seq values. Please check the documentation for more
+    details.
+    """
+
+    vault_id: Optional[str] = None
+    owner: Optional[str] = None
```
Patel-Raj11, line 32: I don't see `owner` and `seq` mentioned in the spec for vault_info request. Not sure if they are supported.

xrpl/models/transactions/vault_withdraw.py
Patel-Raj11: Can we add unit tests once we add validation checks on all these Transaction model objects?

xrpl/core/binarycodec/types/number.py
```diff
@@ -0,0 +1,283 @@
...
+        Note: This method is faithful to rippled's `Number::to_string()` method.
+        This ensures API compatibility between rippled and xrpl-py regarding the JSON
+        representation of Number objects.
+
+        Returns:
+            A JSON string representing the Number
+        """
+        mantissa = int.from_bytes(self.buffer[:8], byteorder="big", signed=True)
+        exponent = int.from_bytes(self.buffer[8:], byteorder="big", signed=True)
```
khancode: For preciseness:
```suggestion
        exponent = int.from_bytes(self.buffer[8:12], byteorder="big", signed=True)
```

.ci-config/rippled.cfg
```diff
@@ -196,6 +196,9 @@ fixEnforceNFTokenTrustline
 fixReducedOffersV2
 DeepFreeze
 PermissionedDomains
+
+# (proposed, not yet released) 2.5.0 Amendments
```
khancode:
```suggestion
# 2.5.0 Amendments
```

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,76 @@
...
+from xrpl.models.required import REQUIRED
+from xrpl.models.transactions.transaction import Transaction
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
+
+class VaultCreateFlag(int, Enum):
+    """Flags for the VaultCreate transaction."""
+
+    TF_VAULT_PRIVATE = 0x0001
```
khancode: Per [spec](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0065d-single-asset-vault#3111-flags)
```suggestion
    TF_VAULT_PRIVATE = 0x00010000
```

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    Indicates that the vault is private. It can only be set during Vault creation.
+    """
+    TF_VAULT_SHARE_NON_TRANSFERABLE = 0x0002
```
khancode: Per [spec](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0065d-single-asset-vault#3111-flags)
```suggestion
    TF_VAULT_SHARE_NON_TRANSFERABLE = 0x00020000
```

xrpl/models/transactions/vault_create.py
khancode: The default value should also be `vaultStrategyFirstComeFirstServe`, according to the spec.

ckeshava: Which version of the spec are you guys using? [The spec appears to have been updated/changed](https://github.com/XRPLF/XRPL-Standards/discussions/192). I don't see the VaultCreate transaction (or) its associated default values in the spec anymore.

xrpl/models/requests/vault_info.py
ckeshava: This might be the case where the specification is not up-to-date with the cpp implementation. Please refer to this file for details of the VaultInfo implementation: https://github.com/XRPLF/rippled/blob/develop/src/xrpld/rpc/handlers/VaultInfo.cpp

xrpl/models/transactions/vault_create.py
ckeshava: As per [this version of the spec](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0065d-single-asset-vault), the `withdrawal strategy` has a JSON type of `number`
image. 

We will need to introduce an enum in all the client libraries for this transaction type. Is that alright? Furthermore, we will need to allow both `number` and `enum` as possible inputs values for this field. Am I understanding that correctly?

xrpl/models/transactions/vault_create.py
ckeshava: thanks, these were the values in an older version of the spec. Updated in f00a4c5c

xrpl/models/transactions/vault_create.py
ckeshava: fixed in f00a4c5c

xrpl/models/transactions/vault_create.py
ckeshava: fixed in f00a4c5c

xrpl/models/requests/vault_info.py
ckeshava: Hello Raj, regarding the error-validation and the associated unit tests -- There are 11 `tem...` error codes returned by the `Vault`-related transactions. Refer to this file for details: https://github.com/XRPLF/rippled/blob/develop/src/xrpld/app/tx/detail/VaultCreate.cpp

Are we planning to add all such error cases into our validation? The choice for the error-validation and the unit tests have become discretionary, as per the existing client library implementation.

I'm happy to add your suggestions, but I wanted to point out other error-cases too, for the sake of completeness.

xrpl/models/transactions/vault_create.py
Patel-Raj11: Yes, JS can also have enum to represent this field. Yes, your understanding is correct to allow both `number` and `enum`. We have followed such pattern in [GranularPermission](https://github.com/XRPLF/xrpl-py/blob/031174a1e6d30576419d1ba4fa625b768a0a60f3/xrpl/models/transactions/delegate_set.py#L29)

xrpl/models/requests/vault_info.py
Patel-Raj11: > This might be the case where the specification is not up-to-date with the cpp implementation. Please refer to this file for details of the VaultInfo implementation: https://github.com/XRPLF/rippled/blob/develop/src/xrpld/rpc/handlers/VaultInfo.cpp

Yes, it looks like CPP implementation supports `owner` and `seq`. Quick skim on CPP code gives an impression that either `vault_id` or (`owner` and `seq`) are supported. We can add such model validation as done for some other requests.

Patel-Raj11: > Hello Raj, regarding the error-validation and the associated unit tests -- There are 11 `tem...` error codes returned by the `Vault`-related transactions. Refer to this file for details: https://github.com/XRPLF/rippled/blob/develop/src/xrpld/app/tx/detail/VaultCreate.cpp
> 
> Are we planning to add all such error cases into our validation? The choice for the error-validation and the unit tests have become discretionary, as per the existing client library implementation.
> 
> I'm happy to add your suggestions, but I wanted to point out other error-cases too, for the sake of completeness.

I am not suggesting to add validation for tem error codes, those would appear as part of integration tests if we were to assert those. If you were referring to this comment `Can we add unit tests once we add validation checks on all these Transaction model objects?`, I meant to add the model validations as we are currently adding to transactions and requests.

xrpl/models/transactions/vault_create.py
Patel-Raj11: [this version of the spec](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0065d-single-asset-vault) I was pointing to.

khancode: I was also referring to this [version](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0065d-single-asset-vault). And yes, it's best to represent them as enum.

xrpl/models/requests/vault_info.py
ckeshava: In the xrpl-py repository, I haven't seen validations for the Request models. The files in this folder `xrpl/models/requests/*.py` do not contain any `get_errors()` method to validate the inputs to the Requests.

xrpl/core/binarycodec/definitions/definitions.json
```diff
@@ -3341,6 +3344,10 @@
     "TicketCreate": 10,
     "TrustSet": 20,
     "UNLModify": 102,
+    "VaultCreate": 65,
+    "VaultClawback": 70,
+    "VaultDeposit": 68,
+    "VaultDelete": 67,
```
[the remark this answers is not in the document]
ckeshava: addressed in  9488ae9

xrpl/core/binarycodec/definitions/definitions.json
```diff
@@ -3160,6 +3160,8 @@
     "tecUNFUNDED_OFFER": 103,
     "tecUNFUNDED_PAYMENT": 104,
     "tecWRONG_ASSET": 194,
+    "tecLIMIT_EXCEEDED": 195,
+    "tecPSEUDO_ACCOUNT": 196,
```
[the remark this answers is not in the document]
ckeshava: addressed in  9488ae9

xrpl/models/transactions/vault_withdraw.py
ckeshava: > Amount defaults to 0 as per the spec.

Regarding point-2 in your first comment, the Amount field is mandatory. Specifying a default value is not useful for mandatory fields. Do you recommend enforcing the default value in the transaction model? @Patel-Raj11 

image

xrpl/models/transactions/vault_delete.py
ckeshava: comment addressed in [ebb70ec](https://github.com/XRPLF/xrpl-py/pull/814/commits/ebb70ec7008e18409f73d8a0cf1c36c64442e4ed)

xrpl/models/transactions/vault_set.py
ckeshava: comment addressed in [ebb70ec](https://github.com/XRPLF/xrpl-py/pull/814/commits/ebb70ec7008e18409f73d8a0cf1c36c64442e4ed)

xrpl/models/transactions/vault_create.py
ckeshava: comment addressed in [ebb70ec](https://github.com/XRPLF/xrpl-py/pull/814/commits/ebb70ec7008e18409f73d8a0cf1c36c64442e4ed)

xrpl/models/transactions/vault_withdraw.py
ckeshava: Bullet point-1 is addressed in [ebb70ec](https://github.com/XRPLF/xrpl-py/pull/814/commits/ebb70ec7008e18409f73d8a0cf1c36c64442e4ed)

xrpl/models/transactions/vault_create.py
ckeshava: Hello Raj, I agree that the specification uses `number` type to represent this field. However, I chose `str` for its advantages over the number type.

Different programming languages have different precision in the `number` type representation. Consequently, languages like Javascript cannot represent large values which are supported in the XRPL eco-system. The range of values supported by `Uint64_t` (internal rippled type) and Python's `int`, Javascript's `number` are not identical.

Due to this reason, other parts of the XRPL ecosystem use strings (or) hexa-decimal numbers to represent large values. Here is an example of range-mismatch from the [OracleSet transaction](https://xrpl.org/docs/references/protocol/transactions/types/oracleset):
image

I felt that strings `str` type would be a programming-language agnostic way to represent large numeric, non-fractional values.

xrpl/models/transactions/vault_create.py
ckeshava: this suggestion is incorporated in [a2619de](https://github.com/XRPLF/xrpl-py/pull/814/commits/a2619de16b09e0074692531cb0c3b370003ef428)

ckeshava: > The default value should also be vaultStrategyFirstComeFirstServe, according to the spec.

The cpp implementation of the rippled client handles the default value for this field. Client libraries need to faithfully communicate the developer's inputs to the rippled client, the default value is not relevant for the functioning of the client library.

xrpl/models/transactions/vault_create.py
Patel-Raj11: 1. A byte accommodates two hex characters. So, `256 bytes` would actually hold 512 characters. The checks where hex characters are expected should be fixed.
2. We can have a regex to check for Hash256 field such as DomainID in some util file. That can be used everywhere.

xrpl/models/transactions/vault_create.py
Patel-Raj11: Ah, so CPP's JSON deserializer would convert `str` to the appropriate `number` as I understand. Integration tests seems to be passing, so I am fine with this approach.

xrpl/models/transactions/vault_set.py
Patel-Raj11: @ckeshava I am un-resolving the conversations that still needs some fixes.

xrpl/models/transactions/vault_withdraw.py
Patel-Raj11: I see, it defeats the purpose of amount being the mandatory field in vault_withdraw transaction by defaulting it to 0. I take the #2 back.

xrpl/models/requests/vault_info.py
Patel-Raj11: [amm_info](https://github.com/XRPLF/xrpl-py/blob/main/xrpl/models/requests/amm_info.py), [get_aggregate_price](https://github.com/XRPLF/xrpl-py/blob/main/xrpl/models/requests/get_aggregate_price.py) are two requests that has such validations. I was leaning towards adding validations to requests/transactions because incorrect request would anyway fail at rippled end and we can save a network call if such validations are present. Leave it up to you, I don't feel strongly.

Patel-Raj11:
@ckeshava Overall looks good to me. Left few comments on the transaction model validations part of the PR (size of some fields, and Hash256 validation).

xrpl/models/requests/vault_info.py
ckeshava: Yes, I wrote the implementation of the `get_aggregate_price` RPC last year. `vault_info` RPC uses one of the following inputs: `vault_id` (or) `owner` + `seq`. Without accessing the state of the ledger, we cannot perform meaningful validation on these values. 

For example, check if the `vault_id` exists in the ledger, check if the `owner` account exists, check if the `seq` value is appropriate (not too old, not too new) -- all of these checks need to read the state of the blockchain. This is out of scope for the client libraries.

I could still perform checks such as validate the `vault_id` based on a regular-expression. However, these mistakes are unlikely to occur because users obtain `vault_id` by reading the XRPL blockchain through `ledger_entry` (or) equivalent RPC calls.

I'm leaning toward not adding the latter-type of checks into the client library code base.

Patel-Raj11: I was referring to validations that was suggested by coderabbitai [here](https://github.com/XRPLF/xrpl-py/pull/814/files/d7a903f01bb274cab72a2997f82436c5c919d315#r2127459359).

xrpl/models/transactions/vault_create.py
ckeshava: `Hash256` type contains 256 bits, not 256 bytes. 

- Hence, 256 bits is represented by `256 / 8 = 32 bytes`.
- Each byte is represented by 2 hexa-decimal characters. 32 bytes are represented by `32 * 2 = 64` base-16 representation.

This PR checks if the length of the input is 64 based on the above logic.

If you'd like to see an example of the `domain_id` values, you can print out intermediate values [in this file](https://github.com/XRPLF/xrpl-py/blob/main/tests/integration/transactions/test_permissioned_domain.py). In my experiments, I came across `CE47F59928D43773A8A9CB7F525BE031977EFB72A23FF094C1C326E687D2B567` for this `domain_id` field.

Reference: Documentation for [Hash fields](https://xrpl.org/docs/references/protocol/binary-format#hash-fields).

xrpl/models/transactions/vault_delete.py
ckeshava: let me know (or) reopen this conversation, if you still have concerns.

xrpl/models/transactions/vault_create.py
Patel-Raj11: DomainID- Yes, its 256 bits, so 64 character length check is good. I was asking to add checks to validate that the value is a valid `Hash256` like it has only A-F, a-f and 0-9.

`Data` and `MPTokenMetadata` size checks are incorrect in my opinion. You can comments these checks and run one of the integration check with the values beyond the current limits that you have set and verify if its failing at rippled end. It shouldn't in my opinion.

xrpl/models/requests/vault_info.py
ckeshava: I have added that suggestion in the latest commit: 56a19e8a

xrpl/models/transactions/vault_create.py
ckeshava: > Data and MPTokenMetadata size checks are incorrect in my opinion. 

I tested the data field with 257 characters. I also disabled the model checks inside the client library. `rippled` replied with a `temMALFORMED` error. I have attached the patch file here, for your reference.
[data_field_length.patch](https://github.com/user-attachments/files/20871607/data_field_length.patch)

I also verified that a string of length 256 characters works correctly with the rippled instance.

Here is an [associated test case](https://github.com/XRPLF/rippled/blob/e2fa5c1b7cfff1c6c1617ef3bc3e95cb65e0352d/src/test/app/Vault_test.cpp#L746C13-L751C14) from rippled code base.

There is a slight difference between the cpp and Python test case. Unlike the cpp code, Python needs to convert the input into UTF-8 encoding and then into the hexadecimal format.

Does this help clarify the issue?

Patel-Raj11: `str_to_hex("A" * 256)` => Rippled will accept the transaction without the validations checks in vault_create transaction. With your existing validation checks, it will fail to submit the transaction, which it shouldn't.

> I tested the data field with 257 characters.

257 character string will turn out to be 514 hex characters using `str_to_hex` function and rippled will rightly throw `temMALFORMED` since its `257 bytes` and its greater than `256 bytes` limit.

Patel-Raj11: Reference implementation:
URI is 256 bytes in NFTokenModify transaction.

https://github.com/XRPLF/xrpl-py/blob/1482b68a4b993b4d77407a735d9e4b059ca323d6/xrpl/models/transactions/nftoken_modify.py#L15

https://xrpl.org/docs/references/protocol/transactions/types/nftokenmodify#nftokenmodify-fields

ckeshava: > DomainID- Yes, its 256 bits, so 64 character length check is good. I was asking to add checks to validate that the value is a valid Hash256 like it has only A-F, a-f and 0-9.

All the Hash256 fields are [validated for their inputs here](https://github.com/XRPLF/xrpl-py/blob/1482b68a4b993b4d77407a735d9e4b059ca323d6/xrpl/core/binarycodec/types/blob.py#L62). There are a total of 36 fields of the type `Hash256` (also called UINT256 in rippled internal types). There are many other `Hash<length_of_bits>` serialized fields, apart from these. In order to avoid duplication of code, the validation for all these `Hash`-fields are performed inside the `binary-codec` of the client library.

[Here is a unit-test](https://github.com/XRPLF/xrpl-py/blob/1482b68a4b993b4d77407a735d9e4b059ca323d6/tests/unit/core/binarycodec/types/test_hash_types.py#L126) which validates the behavior of non-hexa-decimal inputs for the Hash type.

I prefer to retain the existing pattern of validation inside the binary-codec. This reduces boiler-plate repetitive code for all the 36 existing (and potentially future) Hash256 fields.

Here is a patch file that validates the expected behavior with non-hexadecimal inputs for the `data` field inside the `VaultCreate` transaction.
[non_hex_data_field.patch](https://github.com/user-attachments/files/20872111/non_hex_data_field.patch)

Patel-Raj11: https://github.com/XRPLF/xrpl-py/pull/814#discussion_r2162512756 makes sense. 💯

ckeshava: Thanks for the clarification. I have addressed the mistakes in the length discrepencies in 4e9a8059

Patel-Raj11 (approved):
LGTM! Nice work on this!

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,106 @@
...
+
+    transaction_type: TransactionType = field(
+        default=TransactionType.VAULT_CREATE,
+        init=False,
+    )
+
+    def _get_errors(self: Self) -> Dict[str, str]:
+        errors = super()._get_errors()
+
+        if self.data is not None and len(self.data) > (256 * 2):
```
pdp2121: nit: would prefer these numbers set to variables

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,106 @@
...
+    vaultStrategyFirstComeFirstServe	   1	          Requests are processed on a first-
+                                                            come-first-serve basis.
+    """
+
+    transaction_type: TransactionType = field(
+        default=TransactionType.VAULT_CREATE,
+        init=False,
+    )
+
+    def _get_errors(self: Self) -> Dict[str, str]:
```
pdp2121, line 90: Can we test for these failure conditions as well?
https://github.com/XRPLF/XRPL-Standards/blob/master/XLS-0065d-single-asset-vault/README.md?plain=1#L347-L354

xrpl/models/transactions/vault_deposit.py
```diff
@@ -0,0 +1,26 @@
...
+from xrpl.models.amounts import Amount
+from xrpl.models.required import REQUIRED
+from xrpl.models.transactions.transaction import Transaction
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class VaultDeposit(Transaction):
```
pdp2121, line 14: Should also check for valid vault_id here

xrpl/models/requests/vault_info.py
```diff
@@ -0,0 +1,45 @@
...
```
pdp2121, line 32: Can you add descriptions for these fields? Especially as seq and owner are not specified in the XLS proposal

xrpl/models/transactions/vault_create.py
ckeshava: Hello Phu, we can check these flags, however we will need to fetch the appropriate `MPTIssuance` / `Trustline` / `AccountRoot` ledger object from the blockchain. The latency introduced by that read-operation could render the information stale. For example, the freeze-flag on a trust line could change after our read-operation and before the transaction is actually submitted to the XRPL blockchain.

Due to this reason, we don't perform such checks in the client library. This is true for all the failure conditions which return a `tec`- error-code. All of them need the latest data from the validated ledger.

xrpl/models/requests/vault_info.py
ckeshava: fixed in [3661d9a](https://github.com/XRPLF/xrpl-py/pull/814/commits/3661d9ac4477a42dae041eec10bb6b1c0c76735d)

xrpl/models/transactions/vault_deposit.py
ckeshava: fixed in [3661d9a](https://github.com/XRPLF/xrpl-py/pull/814/commits/3661d9ac4477a42dae041eec10bb6b1c0c76735d)

xrpl/models/transactions/vault_create.py
ckeshava: fixed in [3661d9a](https://github.com/XRPLF/xrpl-py/pull/814/commits/3661d9ac4477a42dae041eec10bb6b1c0c76735d)

xrpl/core/binarycodec/definitions/definitions.json
```diff
@@ -3088,7 +3088,7 @@
     "RippleState": 114,
     "SignerList": 83,
     "Ticket": 84,
-    "Vault": 132,
+    "Vault" : 132,
```
khancode:
```suggestion
    "Vault": 132,
```
Whitespace should be removed to match formatting.

xrpl/core/binarycodec/types/number.py
```diff
@@ -0,0 +1,283 @@
...
+        + r"(0|[1-9][0-9]*)"  # mantissa: a number (no leading zeroes, unless 0)
+        + r"(\.([0-9]+))?"  # (optional) decimal point and fractional part
+        + r"([eE]([+-]?)([0-9]+))?"  # (optional) E/e, optional + or -, any number
+        + r"$"  # the end of the string
+    )
+
+    matches = re.fullmatch(VALID_NUMBER_REGEX, value)
+
+    if not matches:
+        raise XRPLBinaryCodecException("Unable to parse number from the input string")
```
khancode: Will be helpful to include input string in the error message.

xrpl/models/amounts/mpt_amount.py
```diff
@@ -49,3 +49,16 @@ def to_currency(self: Self) -> MPTCurrency:
             The MPTCurrency for this MPTAmount.
         """
         return MPTCurrency(mpt_issuance_id=self.mpt_issuance_id)
+
+
+class MPTIssue:
```
khancode: You can delete this and just use the existing `MPTCurrency` model.

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,110 @@
...
+    VAULT_STRATEGY_FIRST_COME_FIRST_SERVE = 1
+    """Requests are processed on a first-come-first-serve basis."""
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class VaultCreate(Transaction):
+    """The VaultCreate transaction creates a new Vault object."""
+
+    asset: Union[Currency, MPTIssue] = REQUIRED  # type: ignore
```
khancode:
```suggestion
    asset: Union[Currency] = REQUIRED  # type: ignore
```
This can be deleted since `MPTCurrency` is covered in `Currency`.

xrpl/models/transactions/vault_delete.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import Dict
+
+from typing_extensions import Self
+
+from xrpl.models.required import REQUIRED
+from xrpl.models.transactions.transaction import Transaction
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
+_MAX_VAULT_ID_LENGTH = 32 * 2
```
khancode:
```suggestion
_MAX_VAULT_ID_LENGTH = 64
```
I would just set it as a single integer. If you want to explain how the value is derived, you can add a comment.

xrpl/models/transactions/vault_clawback.py
```diff
@@ -0,0 +1,42 @@
...
+    The Clawback transaction must respect any future fees or penalties.
+    """
+
+    vault_id: str = REQUIRED  # type: ignore
+    """The ID of the vault from which assets are withdrawn."""
+
+    holder: str = REQUIRED  # type: ignore
+    """The account ID from which to clawback the assets."""
+
+    amount: Optional[Union[IssuedCurrencyAmount, MPTAmount]] = None
```
khancode: Similar to what I did in my [xrpl.js PR](https://github.com/XRPLF/xrpl.js/pull/3008/files#diff-695e6507bbf3a86833c7318c48da6ec25fe12c7a4f8a90a3a0cfbb8c19f8b3abR196-R208), it's better to add a new amount type used for Clawback transactions: `ClawbackAmount = Union[IssuedCurrencyAmount | MPTAmount]`. You can reuse this type in `Clawback` transaction too.

xrpl/models/requests/vault_info.py
```diff
@@ -0,0 +1,48 @@
...
+    vault_id or both owner and seq values. Please check the documentation for more
+    details.
+    """
+
+    vault_id: Optional[str] = None
+    """The object ID of the Vault to be returned."""
+    owner: Optional[str] = None
+    """The account address of the Vault Owner."""
+    seq: Optional[int] = None
+    """The transaction sequence number that created the vault."""
```
khancode:
```suggestion
    vault_id: Optional[str] = None
    """The object ID of the Vault to be returned."""

    owner: Optional[str] = None
    """The account address of the Vault Owner."""
    
    seq: Optional[int] = None
    """The transaction sequence number that created the vault."""
```
Add spacing so it's easier to read and matches existing formatting.

tests/integration/transactions/test_sav.py
```diff
@@ -0,0 +1,167 @@
...
+from xrpl.models.currencies import IssuedCurrency
+from xrpl.models.requests import AccountObjects, LedgerEntry
+from xrpl.models.requests.account_objects import AccountObjectType
+from xrpl.models.response import ResponseStatus
+from xrpl.models.transactions.vault_create import WithdrawalPolicy
+from xrpl.utils import str_to_hex
+from xrpl.wallet import Wallet
+
+
+class TestSingleAssetVault(IntegrationTestCase):
```
khancode: Can you rename this test module to `test_single_asset_vault.py`? That way it'll be easier to search in a codebase.

xrpl/models/amounts/mpt_amount.py
ckeshava: thanks, I didn't realize that there is an existing definition.

xrpl/models/transactions/vault_create.py
ckeshava: thanks

xrpl/models/requests/vault_info.py
ckeshava: Hello @khancode , I'll need to revert this commit. `black` linter complains about the whitespace. Here is the [CI example](https://github.com/XRPLF/xrpl-py/actions/runs/16035911248/job/45247517657?pr=814#step:8:16).

tests/integration/transactions/test_sav.py
ckeshava: fixed in dfc19ffe

xrpl/models/transactions/vault_clawback.py
ckeshava: fixed in dfc19ffe

xrpl/models/transactions/vault_create.py
ckeshava: fixed in dfc19ffe

xrpl/models/amounts/mpt_amount.py
ckeshava: fixed in dfc19ffe

xrpl/core/binarycodec/types/number.py
ckeshava: fixed in dfc19ffe

xrpl/models/requests/vault_info.py
```diff
@@ -0,0 +1,49 @@
...
+    vault_id or both owner and seq values. Please check the documentation for more
+    details.
+    """
+
+    vault_id: Optional[str] = None
+    """The object ID of the Vault to be returned."""
+
+    owner: Optional[str] = None
+    """The account address of the Vault Owner."""
+    seq: Optional[int] = None
```
khancode:
```suggestion
    """The account address of the Vault Owner."""

    seq: Optional[int] = None
```
There should be a blank line between them but no whitespace included.

xrpl/models/requests/vault_info.py
khancode: @ckeshava oh, it looks like there was whitespace included in the blank line. Once you remove the whitespace, the error will resolve.

xrpl/models/amounts/__init__.py
```diff
@@ -14,6 +16,8 @@
 from xrpl.models.amounts.issued_currency_amount import IssuedCurrencyAmount
 from xrpl.models.amounts.mpt_amount import MPTAmount
 
+ClawbackAmount = Union[IssuedCurrencyAmount, MPTAmount]
+
```
khancode: I would move this into a new module named `clawback_amount.py`, similar to how it's done for `Amount` in `xrpl/models/amounts/amount.py`

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,109 @@
...
+from xrpl.models.currencies import Currency
+from xrpl.models.flags import FlagInterface
+from xrpl.models.required import REQUIRED
+from xrpl.models.transactions.transaction import Transaction
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
+_MAX_DATA_LENGTH = 256 * 2
+_MAX_MPTOKEN_METADATA_LENGTH = 1024 * 2
+_MAX_DOMAIN_ID_LENGTH = 32 * 2
```
khancode:
```suggestion
VAULT_MAX_DATA_LENGTH = 256 * 2
VAULT_MAX_DOMAIN_ID_LENGTH = 32 * 2
_VAULT_MAX_MPTOKEN_METADATA_LENGTH = 1024 * 2
```
'_' should be removed from the ones that are also used in `VaultSet`

ckeshava: fixed in 9144d2c5

xrpl/models/amounts/__init__.py
ckeshava: fixed in 9144d2c5

xrpl/models/requests/vault_info.py
ckeshava: fixed in 9144d2c5

tests/integration/transactions/test_single_asset_vault.py
```diff
@@ -0,0 +1,167 @@
...
+        self.assertEqual(response.status, ResponseStatus.SUCCESS)
+        self.assertEqual(response.result["engine_result"], "tesSUCCESS")
+
+        # Step-1.b: Verify the existence of the vault with account_objects RPC call
+        account_objects_response = await client.request(
+            AccountObjects(account=vault_owner.address, type=AccountObjectType.VAULT)
+        )
+        self.assertEqual(len(account_objects_response.result["account_objects"]), 1)
+
+        VAULT_ID = account_objects_response.result["account_objects"][0]["index"]
```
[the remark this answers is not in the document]
ckeshava: The existence is guaranteed due to a successful `account_objects` API response. The success is verified based on the length of the array in the `response.result`.

xrpl/models/transactions/vault_create.py
```diff
@@ -0,0 +1,112 @@
...
+    def _get_errors(self: Self) -> Dict[str, str]:
+        errors = super()._get_errors()
+
+        if self.data is not None and len(self.data) > VAULT_MAX_DATA_LENGTH:
+            errors["data"] = (
+                "Data must be less than 256 bytes (alternatively, 512 hex characters)."
+            )
+        if self.mptoken_metadata is not None and len(self.mptoken_metadata) > (
+            _VAULT_MAX_MPTOKEN_METADATA_LENGTH
+        ):
```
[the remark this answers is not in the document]
ckeshava: The existing validation is correct. `_VAULT_MAX_MPTOKEN_METADATA_LENGTH ` is the correct name. The Github CI has not reported any errors.

khancode (approved):
LGTM, awesome work! 👏

ckeshava:
thank you for the reviews _/\_ @khancode  @Patel-Raj11

PR #820, opened 2025-03-17
Fix the get_latest_open_ledger_sequence method

### LINKED ISSUE #810: get_tatest_open_ledger_sequence() uses old/incorrect API
`get_latest_validated_ledger_sequence()` works
`get_latest_open_ledger_sequence()` doesn't. 

Seems to pass `open` to the `ledger` command which isn't a thing. (anymore?)

```python
>>> get_latest_validated_ledger_sequence(client)
635824
>>> get_latest_open_ledger_sequence(client)
  File "<python-input-8>", line 1, in <module>
    get_latest_open_ledger_sequence(client)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home/emel/dev/Ripple/xrpl-py/xrpl/ledger/main.py", line 39, in get_latest_open_ledger_sequence
    return asyncio.run(main.get_latest_open_ledger_sequence(client))
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/emel/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python3.13/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/home/emel/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/emel/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python3.13/asyncio/base_events.py", line 720, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/home/emel/dev/Ripple/xrpl-py/xrpl/asyncio/ledger/main.py", line 49, in get_latest_open_ledger_sequence
xrpl.asyncio.clients.exceptions.XRPLRequestFailureException: Request failed, invalidParams: ledgerIndexMalformed
```

With no parameters, you get an `open` key back, maybe that's what it's trying to do?
```python
>>> print(json.dumps(requests.post(url=url, json={"method": "ledger", "params": []}).json()["result"]["open"], indent=2))
{
  "ledger": {
    "closed": false,
    "ledger_index": "635941",
    "parent_hash": "DFD665B6C7BE0A9E8F207B3E0B948EF2E000557707A2E210745E077FE0B6F324"
  }
}
```

Show `rippled` doesn't know what that is:

```python
>>> print(json.dumps(requests.post(url=url, json={"method": "ledger", "params": [{"ledger_index": "open"}]}).json()["result"], indent=2))
{
  "error": "invalidParams",
  "error_code": 31,
  "error_message": "ledgerIndexMalformed",
  "request": {
    "command": "ledger",
    "ledger_index": "open"
  },
  "status": "error"
}
```

ckeshava:
Fixes https://github.com/XRPLF/xrpl-py/issues/810

Reference documentation for the `ledger` RPC method: https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/ledger-methods/ledger#request-format

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

This is a Breaking Change. It changes the behavior of the external user-facing API.

However, if any user were using this method, they should have reported this bug sooner. This change can be integrated into the `main` branch without disrupting the workflow of any users because nobody seems to be using it.


xrpl/asyncio/ledger/main.py
```diff
@@ -42,9 +49,21 @@ async def get_latest_open_ledger_sequence(client: Client) -> int:
...
+    ), "`closed` key is not relevant in current ledger"
+    assert (
+        response.result["validated"] is False
+    ), "`validated` key is not relevant in current ledger"
+    assert (
+        "ledger_index" not in response.result
+    ), "`ledger_index` key is not relevant in current ledger"
+    assert (
+        "ledger_current_index" in response.result
+    ), "`ledger_current_index` key must be present in current ledger"
```
khancode, line 64: Why are you using assert statements in non-test modules?

xrpl/asyncio/ledger/main.py
```diff
@@ -42,9 +49,21 @@ async def get_latest_open_ledger_sequence(client: Client) -> int:
...
     if response.is_successful():
-        return cast(int, response.result["ledger_index"])
+        return cast(int, response.result["ledger_current_index"])
 
     raise XRPLRequestFailureException(response.result)
```
Patel-Raj, line 68: Since we are not taking user inputs, can this exception `XRPLRequestFailureException` ever occur? If there is a chance of this exception occurring, then we should assert only when the response is successful. i.e. inside ` if response.is_successful():` to avoid incorrect AssertionError to the user.

khancode: @Patel-Raj  I think it's better to keep as is just in case if an exception does occur with the response being unsuccessful, then it's flagged to the user rather than going unnoticed.

tests/integration/reqs/test_ledger.py
```diff
@@ -1,5 +1,6 @@
 from tests.integration.integration_test_case import IntegrationTestCase
 from tests.integration.it_utils import test_async_and_sync
+from xrpl.asyncio.ledger import get_latest_open_ledger_sequence
```
mvadari, line 3: This function should be tested in `tests/integration/sugar/test_ledger.py`. This file is for testing the `ledger` RPC.

xrpl/asyncio/ledger/main.py
ckeshava: thanks. I have updated them to throw Exceptions now. 25e7304c

tests/integration/reqs/test_ledger.py
ckeshava: updated in 25e7304c

khancode (approved):

xrpl/asyncio/ledger/main.py
Patel-Raj: @ckeshava If the response is not successful then I think it is better to raise `XRPLRequestFailureException(response.result)` and early exit from the function as it is done in `get_fee` function. In the current code, if the response is not successful we would throw a KeyError instead of `XRPLRequestFailureException(response.result)`

ckeshava: @Patel-Raj  thanks, I have updated the code in [8d3d932](https://github.com/XRPLF/xrpl-py/pull/820/commits/8d3d93225105d73a0dfd94d35c85b679fdc3dcc7)

Patel-Raj (approved):

PR #821, opened 2025-03-24
Fix: default flags field for transaction frozen class should be None

AvbrehtLuka:

By default flags field in transaction frozen class should be None rather than 0

### Context of Change

When creating transactions using the frozen classes, the flags should not be 

### Type of Change

- [ ] Yes

## Test Plan

Current tests 

## Future Tasks

No future tasks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - add `MPTCurrency` support in `Issue` (rippled internal type)
 - Fixed the implementation error in get_latest_open_ledger_sequence method. The change uses the "current" ledger for extracting sequence number.
 - Increase default maximum payload size for websocket client
+- Fixed the default behavior of flags field when preparing transactions. By default, flags are not part of the transaction if not explicitly provided.
 
 ### Added
 - Improved validation for models to also check param types
--- a/tests/unit/models/transactions/test_payment.py
+++ b/tests/unit/models/transactions/test_payment.py
@@ -189,3 +189,59 @@ class TestPayment(TestCase):
         }
         tx = Payment(**transaction_dict)
         self.assertTrue(tx.is_valid())
+
+    def test_simple_payment_with_zero_flag(self):
+        payment_tx_json = {
+            "Account": _ACCOUNT,
+            "Destination": _DESTINATION,
+            "TransactionType": "Payment",
+            "Amount": _XRP_AMOUNT,
+            "Fee": _FEE,
+            "Flags": 0,
+            "Sequence": _SEQUENCE,
+        }
+        payment_txn = Payment.from_xrpl(payment_tx_json)
+        payment = payment_txn.to_xrpl()
+
+        self.assertTrue("Flags" in payment)
+
+    def test_simple_payment_with_zero_flag_direct(self):
+        payment_txn = Payment(
+            account=_ACCOUNT,
+            destination=_DESTINATION,
+            amount=_XRP_AMOUNT,
+            fee=_FEE,
+            flags=0,
+            sequence=_SEQUENCE,
+        )
+        payment = payment_txn.to_xrpl()
+
+        self.assertTrue("Flags" in payment)
+
+    def test_simple_payment_with_no_flag_direct(self):
+        payment_txn = Payment(
+            account=_ACCOUNT,
+            destination=_DESTINATION,
+            amount=_XRP_AMOUNT,
+            fee=_FEE,
+            sequence=_SEQUENCE,
+        )
+        payment = payment_txn.to_xrpl()
+
+        self.assertFalse("Flags" in payment)
+
+    def test_simple_payment_with_nonzero_flag(self):
+        payment_tx_json = {
+            "Account": _ACCOUNT,
+            "Destination": _DESTINATION,
+            "TransactionType": "Payment",
+            "Amount": _XRP_AMOUNT,
+            "Fee": _FEE,
+            "Flags": 2147483648,
+            "Sequence": _SEQUENCE,
+        }
+
+        payment_txn = Payment.from_xrpl(payment_tx_json)
+        payment = payment_txn.to_xrpl()
+
+        self.assertTrue("Flags" in payment)
--- a/xrpl/models/transactions/transaction.py
+++ b/xrpl/models/transactions/transaction.py
@@ -200,7 +200,7 @@ class Transaction(BaseModel):
     details.
     """
 
-    flags: Union[Dict[str, bool], int, List[int]] = 0
+    flags: Optional[Union[Dict[str, bool], int, List[int]]] = None
     """
     A List of flags, or a bitwise map of flags, modifying this transaction's
     behavior. See `Flags Field
@@ -275,11 +275,14 @@ class Transaction(BaseModel):
         """
         # we need to override this because transaction_type is using ``field``
         # which will not include the value in the objects __dict__
-        return {
+        prepared_dict = {
             **super().to_dict(),
             "transaction_type": self.transaction_type.value,
-            "flags": self._flags_to_int(),
         }
+        flags = self._flags_to_int()
+        if flags is not None:
+            prepared_dict["flags"] = flags
+        return prepared_dict
 
     def _iter_to_int(
         self: Self,
@@ -291,7 +294,9 @@ class Transaction(BaseModel):
             accumulator |= flag
         return accumulator
 
-    def _flags_to_int(self: Self) -> int:
+    def _flags_to_int(self: Self) -> int | None:
+        if self.flags is None:
+            return None
         if isinstance(self.flags, int):
             return self.flags
         check_false_flag_definition(tx_type=self.transaction_type, tx_flags=self.flags)
@@ -372,6 +377,8 @@ class Transaction(BaseModel):
         Raises:
             XRPLModelException: if `self.flags` is invalid.
         """
+        if self.flags is None:
+            return False
         if isinstance(self.flags, int):
             return self.flags & flag != 0
         elif isinstance(self.flags, dict):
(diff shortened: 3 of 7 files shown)


mvadari:
Why do you want them to be `None` instead of `0`? There isn't really any difference in the outcome.

AvbrehtLuka:
The main problem happens when trying to "serialise" a transactions

```python
a = Payment(
            account=some_account,
            amount=some_amount,
            destination=some_destination,
            fee=some_fee,
            sequence=some_sequence,
        )
b = a.to_dict()

c = "flags" in transaction_dict # expected: false, actually true
```

Alternatively to_dict serialisation can be changed. But I think the proposed change matches the docs and python typing better.

Patel-Raj:
@AvbrehtLuka I think using this declaration would be more accurate:
flags: Optional[Union[Dict[str, bool], int, List[int]]] = None

Currently the tests are failing, can you please address the Unit tests, integration tests, and snippet tests after making this change?

mvadari:
Appears to be a problem with a helper function:
```
File "/home/runner/work/xrpl-py/xrpl-py/xrpl/models/flags.py", line 106, in check_false_flag_definition
    assert all(isinstance(flag, int) for flag in list_flags)
TypeError: 'NoneType' object is not iterable
```

mvadari:
I would prefer updating the helper function. Less confusing that way IMO.

Patel-Raj:
@LukaAvbreht Thanks for proposing the fix. Can you please look into the failing test cases? Thanks!

LukaAvbreht:
Yes!

xrpl/models/transactions/transaction.py
```diff
@@ -275,18 +275,28 @@ def to_dict(self: Self) -> Dict[str, Any]:
...
-        return {
+        prepared_dict = {
             **super().to_dict(),
             "transaction_type": self.transaction_type.value,
-            "flags": self._flags_to_int(),
         }
+        # if "flags" in prepared_dict:
+        #     del prepared_dict["flags"]
+        flags = self._flags_to_int()
+        if self.flags is None or flags == 0:
```
achowdhry-ripple, line 285: nit: it seems like if self.flags is None, then line 284 `flags` will evaluate to 0 anyways based on _iter_to_int, so you wouldn't need both of these checks

xrpl/models/transactions/transaction.py
```diff
@@ -275,18 +275,28 @@ def to_dict(self: Self) -> Dict[str, Any]:
...
+            if "flags" in prepared_dict:
+                del prepared_dict["flags"]
```
achowdhry-ripple, line 287: nit: can you add a comment for future reference on why we need to delete flags here and ensure it is None in the returned dict?

achowdhry-ripple:
@LukaAvbreht thanks for tackling this issue. Overall LGTM with some minor nits. Pipeline is currently failing due to a typing issue, should be good to go after fixing that

xrpl/models/transactions/transaction.py
```diff
@@ -200,7 +200,7 @@ class Transaction(BaseModel):
     details.
     """
 
-    flags: Union[Dict[str, bool], int, List[int]] = 0
+    flags: Optional[Union[Dict[str, bool], int, List[int]]] = None
```
Patel-Raj11, line 203: @AvbrehtLuka Can you please add this change to the following transactions and rectify the test cases accordingly:

1. enable_amendment.py
2. unl_modify.py

xrpl/models/transactions/transaction.py
```diff
@@ -275,18 +275,28 @@ def to_dict(self: Self) -> Dict[str, Any]:
         """
         # we need to override this because transaction_type is using ``field``
         # which will not include the value in the objects __dict__
-        return {
+        prepared_dict = {
```
Patel-Raj11, line 278:
```suggestion
        prepared_dict = {
            **super().to_dict(),
            "transaction_type": self.transaction_type.value,
            "flags": self._flags_to_int(),
        }

        if self.flags is None:
            del prepared_dict["flags"]
        return prepared_dict
```

This would remove the flags field only if it was not set initially. If it is set as 0 we would keep it as it is so that `blob` method generates the correct binary format and after this change there shouldn't be a need to make changes in core/binarycodec tests.

AvbrehtLuka: I believe the expected behaviour of `from_dict` and `to_xrpl` should result in a minimalist transaction. 
If providing Flags filed with value of 0 in a dict that is used to create a transaction object, the `to_dict` and `to_xrpl` should remove the flag as it has no effect.

I added 2 tests to clarify. Let me know if you disagree

xrpl/models/transactions/transaction.py
AvbrehtLuka: Added

xrpl/models/transactions/transaction.py
AvbrehtLuka: Added a comment and a test

xrpl/models/transactions/transaction.py
AvbrehtLuka: Agreed, removed the check

xrpl/models/transactions/transaction.py
Patel-Raj11: @AvbrehtLuka xrpl.js SDK does not remove the Flags field if its 0 when signing and encoding the transaction. So, if we remove it in xrpl-py then the encoded transaction would be different (equally valid though) and we want to keep these libraries in sync. So, its better to keep `flags` as 0 if its explicitly present.

@mvadari Please pitch in if the above analysis don't make sense and its fine to have different `tx_blob` for both xrpl-py and xrpl.js for the same transaction. 
Also, on another note, do we consider this a breaking change since, after this PR gets released, the `encode` in xrpl-py will generate a different output as opposed to what we had previously. If so, do we refrain from merging this in main and merge this is another `release` branch that would be released with a major version bump?

mvadari: I would personally rather just ensure that the `flags_to_int` helper function returns `None` if `flags` is None, rather than add additional code for this check.

xrpl.js can also be checked to ensure it handles no `Flags` field properly.

`encode` from the binary codec isn't changing, personally I think this is more of a bugfix than a breaking change since it's all internal changes and no behavioral changes.

AvbrehtLuka: The js lib does not really have the `from_dict` equivalent as far as I know. Even if we change  the `_flags_to_int` helper function to return None, the same problem persists, meaning if you generate a transaction from dict in which you have flags field explicitly set to `"flags": 0`, the `to_dict` method will return the  the object with flags set to 0 that has no effect. When creating a instance of frozen class, for instance a Payment instance with flags set to 0, the flags filed will not be present in the so generated dict. This is different to how the js lib handles the same example. I can see why you would prefer to keep the explicit `"flags": 0` even if it does nothing. 

Let me know:
1. Should the to_dict method have flags set to 0 if it is explicitly set to 0
2. Should we remove the flags filed if flags are set to zero and have no effect. 

As soon as you let me know I can update and add basic tests to ensure consistency in the future.

tests/unit/models/transactions/test_payment.py
```diff
@@ -189,3 +189,36 @@ def test_mpt_payment(self):
...
+            "Amount": _XRP_AMOUNT,
+            "Fee": _FEE,
+            "Flags": 2147483648,
+            "Sequence": _SEQUENCE,
+        }
+
+        payment_txn = Payment.from_xrpl(payment_tx_json)
+        payment = payment_txn.to_xrpl()
+        
+        self.assertTrue("Flags" in payment)
```
mvadari, line 224: newline at the end of files

xrpl/models/transactions/transaction.py
mvadari: Going one step further:
```suggestion
        prepared_dict = {
	            **super().to_dict(),
	            "transaction_type": self.transaction_type.value,
	        }
	
	        if flags := self._flags_to_int() is not None:
	            prepared_dict["flags"] = self._flags_to_int()
	        return prepared_dict
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - add `MPTCurrency` support in `Issue` (rippled internal type)
 - Fixed the implementation error in get_latest_open_ledger_sequence method. The change uses the "current" ledger for extracting sequence number.
 - Increase default maximum payload size for websocket client
+- Fixed the dafault bahaviour of flags field when preparing thransactions. By default flags are not part of the transaction if not explicitly provided.
```
Patel-Raj11, line 14:
```suggestion
- Fixed the default behavior of flags field when preparing transactions. By default, flags are not part of the transaction if not explicitly provided.
```

xrpl/models/flags.py
```diff
@@ -74,7 +74,7 @@ def interface_to_flag_list(
 
 def check_false_flag_definition(
     tx_type: Union[TransactionType, PseudoTransactionType],
-    tx_flags: Union[Dict[str, bool], List[int]],
+    tx_flags: Optional[Union[Dict[str, bool], int, List[int]]],
```
Patel-Raj11, line 77: This change is not needed as no codepath reaches here with `tx_flags` defined as `None`.

LukaAvbreht:
fixed typos and squashed commits. Please take a look

xrpl/models/flags.py
```diff
@@ -88,6 +88,9 @@ def check_false_flag_definition(
         XRPLModelException: Flags were not set correctly.
     """
     try:
+        if tx_flags is None:
```
Patel-Raj11, line 91: @AvbrehtLuka This part is not needed. `check_false_flag_definition` can remain as it is.

AvbrehtLuka: Nice spot, updated

Patel-Raj11 (approved):
LGTM

achowdhry-ripple (approved):
lgtm

PR #809, opened 2025-02-23
fix: add `MPTAmount` support in `Issue` model

khancode:
## High Level Overview of Change

Adds `MPTAmount` support in `Issue` model and adds missing test module for `Issue`

### Context of Change

rippled internal type `Issue` now supports `MPTAmount` which was missing.

Historically, the Issue type accounted for `XRP` and `IssuedCurrency` types only. This was implemented in Dec 2024 -- [rippled/include/xrpl/protocol/Asset.h at 33e1c42599857336d792effc753795911bdb13f0 · XRPLF/rippled](https://github.com/XRPLF/rippled/blob/33e1c42599857336d792effc753795911bdb13f0/include/xrpl/protocol/Asset.h#L40) 

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### Did you update CHANGELOG.md?

- [x] Yes
- [ ] No, this change does not impact library users

## Test Plan

Adds a missing test module for `Issue`.

PR #822, opened 2025-03-27
fix: add `MPTCurrency` type instead of `MPTAmount` for `Issue`s

mvadari:

This PR switches all `Issue` support to use a new type, `MPTCurrency`, instead of `MPTAmount` (since `Issue` doesn't have a `value` field). It also enhances some helper functions by changing them to be `TypeGuard`s.

This PR will need to make it into the next release to avoid breaking change issues.

### Context of Change

#809 mistakenly added support for `MPTAmount` to `Issue`, when really it should be a new `MPTCurrency` type.

### Type of Change

- [x] Yes

## Test Plan

CI passes.

## Future Tasks

Update xrpl.js accordingly (https://github.com/XRPLF/xrpl.js/pull/2919)

### DIFF
--- a/tests/unit/core/binarycodec/types/test_issue.py
+++ b/tests/unit/core/binarycodec/types/test_issue.py
@@ -32,7 +32,6 @@ class TestIssue(TestCase):
         # Test Issue creation for an MPT amount.
         # Use a valid 48-character hex string (24 bytes) for mpt_issuance_id.
         test_input = {
-            "value": "100",  # MPT amounts must be an integer string (no decimal point)
             "mpt_issuance_id": "BAADF00DBAADF00DBAADF00DBAADF00DBAADF00DBAADF00D",
         }
         issue_obj = Issue.from_value(test_input)
@@ -75,9 +74,9 @@ class TestIssue(TestCase):
         self.assertEqual(issue_from_parser.to_json(), expected)
 
     def test_from_parser_mpt(self):
-        # Test round-trip: serialize an MPT Issue and then parse it back.
+        # Test round-t
+        # rip: serialize an MPT Issue and then parse it back.
         test_input = {
-            "value": "100",
             "mpt_issuance_id": "BAADF00DBAADF00DBAADF00DBAADF00DBAADF00DBAADF00D",
         }
         issue_obj = Issue.from_value(test_input)
--- a/tests/unit/core/binarycodec/types/test_issue.py
+++ b/tests/unit/core/binarycodec/types/test_issue.py
@@ -74,8 +74,7 @@ class TestIssue(TestCase):
         self.assertEqual(issue_from_parser.to_json(), expected)
 
     def test_from_parser_mpt(self):
-        # Test round-t
-        # rip: serialize an MPT Issue and then parse it back.
+        # Test round-trip: serialize an MPT Issue and then parse it back.
         test_input = {
             "mpt_issuance_id": "BAADF00DBAADF00DBAADF00DBAADF00DBAADF00DBAADF00D",
         }
--- a/xrpl/constants.py
+++ b/xrpl/constants.py
@@ -42,6 +42,8 @@ Matches hex-encoded currencies in the format allowed by XRPL.
 :meta private:
 """
 
+HEX_MPTID_REGEX: Final[Pattern[str]] = re.compile(r"^[0-9A-Fa-f]{48}$")
+
 # Constants for validating amounts.
 MIN_IOU_EXPONENT: Final[int] = -96
 """
--- a/xrpl/models/amounts/amount.py
+++ b/xrpl/models/amounts/amount.py
@@ -4,7 +4,9 @@ counterparty (issuer) on the trustline that holds the value. For XRP, there is n
 counterparty.
 """
 
-from typing import Union, cast
+from typing import Union
+
+from typing_extensions import TypeGuard
 
 from xrpl.models.amounts.issued_currency_amount import IssuedCurrencyAmount
 from xrpl.models.amounts.mpt_amount import MPTAmount
@@ -12,7 +14,7 @@ from xrpl.models.amounts.mpt_amount import MPTAmount
 Amount = Union[IssuedCurrencyAmount, MPTAmount, str]
 
 
-def is_xrp(amount: Amount) -> bool:
+def is_xrp(amount: Amount) -> TypeGuard[str]:
     """
     Returns whether amount is an XRP value, as opposed to an issued currency
     or MPT value.
@@ -26,7 +28,7 @@ def is_xrp(amount: Amount) -> bool:
     return isinstance(amount, str)
 
 
-def is_issued_currency(amount: Amount) -> bool:
+def is_issued_currency(amount: Amount) -> TypeGuard[IssuedCurrencyAmount]:
     """
     Returns whether amount is an issued currency value, as opposed to an XRP
     or MPT value.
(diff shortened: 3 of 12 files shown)


xrpl/constants.py
```diff
@@ -42,6 +42,8 @@ class XRPLException(Exception):
 :meta private:
 """
 
+HEX_MPTID_REGEX: Final[Pattern[str]] = re.compile(r"^[0-9A-Fa-f]{48}$")
```
ckeshava, line 45: I was under the belief that `mpt_issuance_id` is represented by 192-bits, as explained here: https://github.com/XRPLF/rippled/blob/6cf37c4abef80fb535d1d7577a548665e139f635/include/xrpl/protocol/MPTIssue.h#L28

However, the size of `mpt_issuance_id` is much larger than 192 bits in the examples. It's four times larger in the RPC outputs.

1. Where can I find the cpp implementation of this `MPTCurrency` type?
2. Which transactions use the `MPTCurrency` type? I'd like to test the correctness of serialization/de-serialization of the `MPTCurrency` type.

xrpl/models/transactions/xchain_claim.py
```diff
@@ -81,13 +80,13 @@ def _get_errors(self: Self) -> Dict[str, str]:
...
         amount = self.amount
         if is_xrp(amount):
-            currency = XRP()
+            currency: Currency = XRP()
         elif is_issued_currency(amount):
-            currency = amount.to_currency()  # type: ignore
+            currency = amount.to_currency()
         elif is_mpt(amount):
-            currency = amount.mpt_issuance_id  # type: ignore
+            currency = amount.to_currency()
```
ckeshava, line 87: both these conditions can be combined into one line

tests/unit/core/binarycodec/types/test_issue.py
```diff
@@ -32,7 +32,6 @@ def test_from_value_mpt(self):
         # Test Issue creation for an MPT amount.
         # Use a valid 48-character hex string (24 bytes) for mpt_issuance_id.
         test_input = {
```
ckeshava, line 34: I'm making a note for future tasks: We need to add unit tests for the `MPTAmount` type. This file now exclusively tests `MPTCurrency` type.

xrpl/models/amounts/amount.py
```diff
@@ -4,15 +4,17 @@
...
+from typing_extensions import TypeGuard
 
 from xrpl.models.amounts.issued_currency_amount import IssuedCurrencyAmount
 from xrpl.models.amounts.mpt_amount import MPTAmount
 
 Amount = Union[IssuedCurrencyAmount, MPTAmount, str]
 
 
-def is_xrp(amount: Amount) -> bool:
+def is_xrp(amount: Amount) -> TypeGuard[str]:
```
ckeshava, line 17: Have you found TypeGuard useful? I'm not convinced that it improves readability in this snippet.

I get the benefits with composite type examples listed here: https://typing.python.org/en/latest/spec/narrowing.html#typeguard, but I feel these types are simple enough to read.

xrpl/models/currencies/currency.py
```diff
@@ -7,6 +7,7 @@
 from typing import Union
 
 from xrpl.models.currencies.issued_currency import IssuedCurrency
+from xrpl.models.currencies.mpt_currency import MPTCurrency
 from xrpl.models.currencies.xrp import XRP
 
-Currency = Union[IssuedCurrency, XRP]
+Currency = Union[IssuedCurrency, MPTCurrency, XRP]
```
ckeshava, line 13: Do all XRPL transactions and requests allow `MPTCurrency` ? This `Currency` type is used in [book_offers](https://github.com/XRPLF/xrpl-py/blob/109c413088e24044c3895e37e1c526b8b26d3d16/xrpl/models/requests/book_offers.py#L23) and [amm_info](https://github.com/XRPLF/xrpl-py/blob/109c413088e24044c3895e37e1c526b8b26d3d16/xrpl/models/requests/amm_info.py#L26) RPCs.

Users could mis-interpret the type hints.

xrpl/constants.py
mvadari: 192 bits is 24 bytes, which is a hex string of length 48.

1. `MPTCurrency` is equivalent to `MPTIssue` in rippled.
2. Not sure there are any. The only two features that use `STIssue` types are XChainBridge and AMM and I don't think either of them support MPTs right now (though support for AMM is being added in [XLS-82d](https://github.com/XRPLF/XRPL-Standards/discussions/231) in https://github.com/XRPLF/rippled/pull/5285).

xrpl/models/transactions/xchain_claim.py
mvadari: IMO it's clearer this way and you don't really get any performance benefit from combining it

xrpl/models/amounts/amount.py
mvadari: All the `# type: ignore`s I was able to remove were because I switched these to typeguards. It's not about readability, it's about `mypy` being able to understand the typing better if these functions are used.

xrpl/models/currencies/currency.py
mvadari: Those RPCs can be updated, but honestly I don't think that's really worth doing because MPT support is coming for those RPCs soon enough anyways.

xrpl/models/currencies/mpt_currency.py
```diff
@@ -0,0 +1,63 @@
...
+        errors = super()._get_errors()
+        if not _is_valid_mptid(self.mpt_issuance_id):
+            errors["mpt_issuance_id"] = (
+                f"Invalid mpt_issuance_id {self.mpt_issuance_id}"
+            )
+        return errors
+
+    def to_amount(self: Self, value: Union[str, int]) -> xrpl.models.amounts.MPTAmount:
+        """
+        Converts an MPTCurrency to an MPTCurrencyAmount.
```
Patel-Raj, line 53: nit: Replace MPTCurrencyAmount with MPTCurrency, so that the docs are in sync with the model names.

mvadari: typo - fixed

Patel-Raj (approved):

xrpl/constants.py
ckeshava: 1. Can we retain the same name `MPTIssue` in the client libraries. Is there a reason you prefer `MPTCurrency` ?
2. I wish there were a way to test the correctness of this serialization. Unless we communicate a serialized `MPTIssue` value to rippled, we can not validate the correctness of the client library binary codec.

As a new feature, if rippled could expose an `echo(serialized_input) => deserialized_output` -like of utility command, that would have been helpful. I'm sure there are other `SField`s which haven't been used in transactions yet.

> 192 bits is 24 bytes, which is a hex string of length 48.

Hmm yes, you are right. I had mis-understood the calculation.

tests/unit/core/binarycodec/types/test_issue.py
ckeshava: hmm yeah. It appears the other features are indirectly testing `MPTAmount`.

xrpl/models/currencies/currency.py
ckeshava: alright

xrpl/constants.py
mvadari: 1.`MPTCurrency` aligns more with what we have in the client library and makes more sense in English. I don't have a strong opinion on this, though.
2. rippled can serialize transactions via the `sign` and `submit` fields locally. You could try with the XLS-82d branch above.

ckeshava: Alright, I would slightly prefer the name `MPTIssue` over `MPTCurrency`, it aligns well with the documentation too.

The MPT amendment has been merged into the rippled codebase, I should be able to use the develop branch to test the `sign`, `submit` commands right?

> rippled can serialize transactions via the sign and submit fields locally

Hmm, this will help me with serialization, but not the other way around i.e. I'm trying to understand how I can deserialize a blob of `MPTIssue` and verify that the xrpl-py library does it correctly?

ckeshava (approved):

PR #828, opened 2025-04-10
[auth] Add custom HTTP header support to clients (API key / payment)

joshuahamsa:

Adds support for sending custom HTTP headers (e.g. API keys or [Dhali](https://dhali.io/) payment-claims) when connecting to XRPL nodes that require authentication, payment, or metadata.

- Adds an optional `headers` argument to the base `Client`, `JsonRpcBase`, `WebsocketBase`, and the sync `WebsocketClient`. Headers are applied to all requests and, for JSON-RPC, can be overridden per request via `_request_impl`.
- The base `Client` stores an owned copy of the headers so later mutation of the caller's dict cannot silently change the headers used for subsequent requests.
- WebSocket clients forward the headers via `additional_headers` on the connection handshake.
- Forces `Content-Type: application/json` on JSON-RPC requests so a custom header cannot accidentally break the JSON body.
- Adds `XRPLAuthenticationException`, raised on `401` (Unauthorized), `402` (Payment Required), and `403` (Forbidden) JSON-RPC responses. The exception's `error` field carries the exact status code.

This consolidates work that was previously split across #763, #827, and #829 (all now closed) into this single PR.

### Context of Change

This is a new feature requested by node operators and infrastructure providers who gate access behind authentication or payment. Headers are added at the base `Client` level (rather than only on `JsonRpcBase`) so the same mechanism serves both the HTTP and WebSocket clients and any future client implementations. `402` handling was added so that Dhali's payment-claim rejections surface as a catchable exception.

### Type of Change

- [x] Yes

## Test Plan

New unit tests were added and all pass locally; `black`, `isort`, `flake8`, and `mypy` are clean.

- `tests/unit/asyn/clients/test_json_rpc_base.py`
  - Global headers are sent with the request, including the default `Content-Type`.
  - Per-request headers override instance-level headers.
  - No headers supplied still works (backward compatible) and keeps the default `Content-Type`.
  - `Content-Type` cannot be overridden by instance- or request-level headers.
  - `401`/`402`/`403` responses raise `XRPLAuthenticationException`.
- `tests/unit/asyn/clients/test_websocket_base.py`
  - The connection handshake forwards `additional_headers` when headers are provided.
  - `additional_headers` is `None` when no headers are provided (default behavior preserved).

Run with:

```
poetry run python -m pytest tests/unit/asyn/clients/
```

### DIFF
--- a/xrpl/asyncio/clients/json_rpc_base.py
+++ b/xrpl/asyncio/clients/json_rpc_base.py
@@ -3,12 +3,16 @@
 from __future__ import annotations
 
 from json import JSONDecodeError
+from typing import Dict, Optional
 
 from httpx import AsyncClient
 from typing_extensions import Self
 
 from xrpl.asyncio.clients.client import REQUEST_TIMEOUT, Client
-from xrpl.asyncio.clients.exceptions import XRPLRequestFailureException
+from xrpl.asyncio.clients.exceptions import (
+    XRPLAuthenticationException,
+    XRPLRequestFailureException,
+)
 from xrpl.asyncio.clients.utils import json_to_response, request_to_json_rpc
 from xrpl.models.requests.request import Request
 from xrpl.models.response import Response
@@ -21,8 +25,28 @@ class JsonRpcBase(Client):
     :meta private:
     """
 
+    def __init__(
+        self: Self,
+        url: str,
+        *,
+        headers: Optional[Dict[str, str]] = None,
+    ) -> None:
+        """
+        Initializes a new JsonRpcBase client.
+
+        Arguments:
+            url: The URL of the XRPL node to connect to.
+            headers: Optional default headers for all requests (e.g. an API key
+                or Dhali payment-claim).
+        """
+        super().__init__(url, headers=headers)
+
     async def _request_impl(
-        self: Self, request: Request, *, timeout: float = REQUEST_TIMEOUT
+        self: Self,
+        request: Request,
+        *,
+        timeout: float = REQUEST_TIMEOUT,
+        headers: Optional[Dict[str, str]] = None,
     ) -> Response:
         """
         Base ``_request_impl`` implementation for JSON RPC.
@@ -30,21 +54,42 @@ class JsonRpcBase(Client):
         Arguments:
             request: An object representing information about a rippled request.
             timeout: The duration within which we expect to hear a response from the
-            rippled validator.
+                rippled server.
+            headers: Optional additional headers to include for this request.
 
         Returns:
             The response from the server, as a Response object.
 
         Raises:
+            XRPLAuthenticationException: if the server responds with 401, 402,
+                or 403.
             XRPLRequestFailureException: if response can't be JSON decoded.
 
         :meta private:
         """
+        # Merge global and per-request headers. Content-Type is applied last
+        # so it cannot be overridden: the JSON-RPC body is always JSON.
+        merged_headers = {
+            **self.headers,
+            **(headers or {}),
+            "Content-Type": "application/json",
+        }
+
         async with AsyncClient(timeout=timeout) as http_client:
             response = await http_client.post(
                 self.url,
                 json=request_to_json_rpc(request),
+                headers=merged_headers,
             )
+            if response.status_code in (401, 402, 403):
+                raise XRPLAuthenticationException(
+                    {
+                        "error": response.status_code,
+                        "error_message": (
+                            f"Authentication or payment required: {response.text}"
+                        ),
+                    }
+                )
             try:
                 return json_to_response(response.json())
             except JSONDecodeError:
@@ -53,4 +98,4 @@ class JsonRpcBase(Client):
                         "error": response.status_code,
                         "error_message": response.text,
                     }
-                )
+                ) from None
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add support for Batch (XLS-56) `BatchV1_1` signing.
+- Support for custom HTTP headers (e.g. API keys, Dhali payment-claims) on the base `Client`, `JsonRpcBase`, and `WebsocketBase` clients, plus `XRPLAuthenticationException` for 401/402/403 responses.
 
 ### Removed
 
--- a/tests/unit/asyn/clients/test_json_rpc_base.py
+++ b/tests/unit/asyn/clients/test_json_rpc_base.py
@@ -1,0 +1,95 @@
+from unittest.mock import AsyncMock, patch
+
+import pytest
+from httpx import Response
+
+from xrpl.asyncio.clients.exceptions import XRPLAuthenticationException
+from xrpl.asyncio.clients.json_rpc_base import JsonRpcBase
+from xrpl.models.requests import ServerInfo
+
+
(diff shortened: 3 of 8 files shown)


joshuahamsa:
cc @mvadari @ckeshava @desimmons

xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -21,16 +22,36 @@ class JsonRpcBase(Client):
     :meta private:
     """
 
+    def __init__(
+        self: Self,
+        url: str,
+        *,
+        headers: Optional[Dict[str, str]] = None,
```
ckeshava, line 29: Can you add a similar update to the WebSocket class's constructor?

xrpl/asyncio/clients/json_rpc_base.py
```diff
@@ -21,16 +22,36 @@ class JsonRpcBase(Client):
...
+    ) -> None:
+        """
+        Initializes a new JsonRpcBase client.
+
+        Arguments:
+            url: The URL of the XRPL node to connect to.
+            headers: Optional default headers for all requests (e.g. API key or Dhali payment-claim).
+        """
+        super().__init__(url, headers=headers)
+        self.headers = headers or {}
```
ckeshava, line 39:
```suggestion
```

Since the parent `Client` class initializes this data member, we don't need to handle it in the child classes.

ckeshava (commented):
Do you prefer to supercede [this PR](https://github.com/XRPLF/xrpl-py/pull/763) with the current one?

Can you also add the tests for this change? The description of the tests (in the PR abstract) seem useful.

joshuahamsa:
> Do you prefer to supercede [this PR](https://github.com/XRPLF/xrpl-py/pull/763) with the current one?
> 
Yes I realize that this would be a better approach than the other.

> 
> 
> Can you also add the tests for this change? The description of the tests (in the PR abstract) seem useful.

Will do!

xrpl/asyncio/clients/json_rpc_base.py
joshuahamsa: Would updating Client supersede this?

dangell7:
Hey can you close out your other PRs? it was very confusing tracking all these changes. 

Here is the implementation for websockets though. 

https://github.com/Transia-RnD/xrpl-py/commit/6c8bab50d04ba1e6d61d81a7dd53009a78f726de

joshuahamsa:
> Hey can you close out your other PRs? it was very confusing tracking all these changes.
> 
> Here is the implementation for websockets though.
> 
> [Transia-RnD@6c8bab5](https://github.com/Transia-RnD/xrpl-py/commit/6c8bab50d04ba1e6d61d81a7dd53009a78f726de)

Previous PRs closed

joshuahamsa (commented):
Final update to include unit tests for the custom header support introduced in @dangell7's [commit](https://github.com/Transia-RnD/xrpl-py/commit/6c8bab50d04ba1e6d61d81a7dd53009a78f726de#diff-9642981ef62b6af3cf3c9b5a68b86527a2b749459969fcbb11a7dd18e0d82c61).

This PR now contains:
	•	Test coverage for JsonRpcBase header usage and override logic (per @ckeshava)
	•	A dedicated check for `XRPLAuthenticationException` on 401/403 responses
	•	Minor cleanup only (no attempt to override the upstream implementation)

Please let me know if you’d prefer this test suite in a separate PR or adjusted to also cover `WebsocketBase`. Thanks again for integrating the core feature!

joshuahamsa:
> Hey can you close out your other PRs? it was very confusing tracking all these changes.
> 
> Here is the implementation for websockets though.
> 
> [Transia-RnD@6c8bab5](https://github.com/Transia-RnD/xrpl-py/commit/6c8bab50d04ba1e6d61d81a7dd53009a78f726de)

Love to see it! Thanks for making this happen!

mvadari:
I think you may have messed up your commits - there's only deletions in the source code.

dangell7:
@joshuahamsa you've gotta cherry pick my commit or add the code yourself. My commit will not be merged. Sorry if that was confusing. What I meant was here is the code, just copy paste it into your PR. :)

`git cherry-pick 6c8bab50d04ba1e6d61d81a7dd53009a78f726de`

joshuahamsa:
Apologies for the long delay and the earlier confusion across multiple PRs. I've cleaned this up so everything now lives in this single PR.

What changed in this update:
- Restored `client.py` and `json_rpc_base.py` (an earlier commit had deleted them outright, which is why the diff was nothing but deletions) and rebased the branch cleanly onto current `main` as a single commit.
- The feature now adds an optional `headers` argument to the base `Client`, `JsonRpcBase`, and `WebsocketBase`, so both HTTP and WebSocket clients can send default headers (e.g. API keys or Dhali payment-claims), with per-request overrides for JSON-RPC. WebSocket headers are forwarded via `additional_headers` on the handshake.
- Added `XRPLAuthenticationException`, raised on 401/403 JSON-RPC responses.
- Folded in the CHANGELOG entry that used to be in #829, so that PR is now redundant and I've closed it.

`black`, `isort`, `flake8`, and `mypy` are all clean locally, and the four unit tests in `test_json_rpc_base.py` pass. This consolidates the work that was previously scattered across #763, #827, and #829. Ready for another look whenever you have a chance — thanks for your patience.

joshuahamsa:
Addressed the CodeRabbit feedback in the latest push:

- **Defensive copy of headers** — `Client.__init__` now stores `dict(headers)` rather than aliasing the caller's dict, so later external mutation can't change the headers used for subsequent requests/reconnects.
- **Content-Type protection** — the JSON-RPC header merge now applies `Content-Type: application/json` last so it can't be overridden by an instance- or request-level header (the JSON-RPC body is always JSON). Added `test_content_type_cannot_be_overridden` to cover it.
- **WebSocket header test** — added `TestWebsocketHeaders` in `test_websocket_base.py` covering both the header-forwarding and no-header (passes `None`) branches of the handshake. Note the forwarding uses `additional_headers=` rather than `extra_headers=`, since `main` now targets the `websockets.asyncio` API.
- **Per-request override test** — added the missing `Content-Type` assertion for consistency with the other header tests.

Also fixed a gap not flagged by the bot: the sync `WebsocketClient.__init__` took `(url, timeout)` and dropped headers entirely, so `WebsocketClient(url, headers=...)` would have raised. It now accepts and forwards `headers`, matching the async clients.

`black`, `isort`, `flake8`, and `mypy` are clean; all nine client unit tests pass.

desimmons:
@joshuahamsa 

> Added XRPLAuthenticationException, raised on 401/403 JSON-RPC responses.

Could you also include 402 status codes?

joshuahamsa:
@desimmons Done — 402 is now included alongside 401 and 403. `XRPLAuthenticationException` is raised for all three, which should cover the "Payment Required" case when a Dhali payment-claim is missing or insufficient.

Since 402 isn't strictly *authentication*, I broadened the exception's docstring and the error message to "authentication or payment required" rather than "authentication failed," and updated the test to assert all three status codes. The exception's `error` field carries the exact HTTP status (401/402/403) so you can branch on it if you need to distinguish payment failures from auth failures.

If you'd prefer a dedicated `XRPLPaymentRequiredException` for 402 instead of folding it into the auth exception, say the word and I'll split it out.

joshuahamsa:
@ckeshava finally ready for review on this one when you get a chance

PR #852, opened 2025-06-30
update the lock file with newer dependency versions

ckeshava:

Update the versions of dependencies in the lock file. This is an attempt to resolve the CI errors. Here is [an example](https://github.com/XRPLF/xrpl-py/actions/runs/15981526900/job/45076783250?pr=814)

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes
Changes to dependency versions should not affect external APIs.


pdp2121 (approved):

Patel-Raj11 (approved):

ckeshava:
>  The reason of the failure is still not clear to me. But these dependency updates won't harm if it unblocks other ongoing PR's.

I agree, I don't have a good root-cause-analysis. However, the updated poetry.lock appears to be the only difference between the this and the other failing PRs.

If the issue of failing CI still persists (after merging this PR), we can explore other solutions.

PR #866, opened 2025-09-13
Support for Lending Protocol (XLS-66d)

ckeshava:

This PR adds support for the Lending Protocol. The specification can be found here: https://github.com/XRPLF/XRPL-Standards/pull/240

The cpp implementation is available here: https://github.com/XRPLF/rippled/pull/5270

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Appropriate unit and integ tests have been added.

### DIFF
--- a/.ci-config/rippled.cfg
+++ b/.ci-config/rippled.cfg
@@ -205,6 +205,7 @@ PermissionedDEX
 Batch
 TokenEscrow
 LendingProtocol
+PermissionDelegationV1_1
 
 # This section can be used to simulate various FeeSettings scenarios for rippled node in standalone mode
 [voting]


tests/integration/transactions/test_lending_protocol.py
```diff
@@ -0,0 +1,182 @@
...
+        # the Borrower.
+
+        loan_issuer_signed_txn = await autofill_and_sign(
+            LoanSet(
+                account=loan_issuer.address,
+                loan_broker_id=LOAN_BROKER_ID,
+                principal_requested="100",
+                start_date=int(datetime.datetime.now().timestamp()),
+                counterparty=borrower_wallet.address,
+            ),
```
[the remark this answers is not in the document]
ckeshava: this field is scheduled for deprecation/removal in the next revision of the specification. There is no need to update it anymore.

xrpl/models/transactions/loan_broker_cover_withdraw.py
```diff
@@ -0,0 +1,38 @@
...
+
+    amount: Amount = REQUIRED  # type: ignore
+    """
+    The Fist-Loss Capital amount to withdraw.
+    """
+
+    destination: Optional[str] = None
+    """
+    An account to receive the assets. It must be able to receive the asset.
+    """
```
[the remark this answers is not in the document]
ckeshava: fixed in [b062bc3](https://github.com/XRPLF/xrpl-py/pull/866/commits/b062bc37d16fb1f4724b7536d1fef48db99a7f5c)

xrpl/models/transactions/loan_broker_cover_deposit.py
```diff
@@ -0,0 +1,32 @@
...
+from __future__ import annotations  # Requires Python 3.7+
+
+from dataclasses import dataclass, field
+
+from xrpl.models.amounts import Amount
+from xrpl.models.required import REQUIRED
+from xrpl.models.transactions.transaction import Transaction
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
```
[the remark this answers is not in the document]
ckeshava: The suggestion is incorrect. First-Loss Capital can be deposited in XRP, IOU or MPT assets. 
image

tests/integration/transactions/test_lending_protocol.py
```diff
@@ -0,0 +1,182 @@
...
+                signing_pub_key=borrower_wallet.public_key,
+                txn_signature=borrower_txn_signature,
+            )
+        )
+
+        response = await submit(
+            Transaction.from_dict(loan_issuer_and_borrower_signature),
+            client,
+            fail_hard=True,
+        )
```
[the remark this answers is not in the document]
ckeshava: fixed in the latest commit.

xrpl/models/transactions/loan_broker_cover_withdraw.py
```diff
@@ -0,0 +1,38 @@
...
+
+from dataclasses import dataclass, field
+from typing import Optional
+
+from xrpl.models.amounts import Amount
+from xrpl.models.required import REQUIRED
+from xrpl.models.transactions.transaction import Transaction
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
```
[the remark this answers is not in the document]
ckeshava: A `LoanBrokerCoverWithdraw` should allow the withdrawal of the First-Loss capital, even if it is XRP asset. This suggestion is incorrect.

xrpl/models/transactions/loan_set.py
```diff
@@ -0,0 +1,177 @@
...
+        if (
+            self.grace_period is not None
+            and self.payment_interval is not None
+            and self.grace_period > self.payment_interval
+        ):
+            parent_class_errors["LoanSet:GracePeriod"] = (
+                "Grace period must be less than the payment interval."
+            )
+
+        return parent_class_errors
```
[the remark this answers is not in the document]
ckeshava: All these validations are already implemented in rippled (The cpp implementation of the XRPL client). 

Furthermore, we are still deliberating the exact JSON-type of these fields. Once we have clarity on the type of these fields, we can consider adding more validations for these fields.

xrpl/models/transactions/loan_broker_cover_clawback.py
```diff
@@ -0,0 +1,63 @@
...
+from xrpl.models.amounts import Amount, get_amount_value, is_xrp
+from xrpl.models.transactions.transaction import Transaction
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class LoanBrokerCoverClawback(Transaction):
+    """This transaction withdraws First Loss Capital from a Loan Broker"""
```
Patel-Raj11: The LoanBrokerCoverClawback transaction claws back the First-Loss Capital from the LoanBroker.

xrpl/models/transactions/loan_broker_cover_clawback.py
```diff
@@ -0,0 +1,63 @@
...
+
+    loan_broker_id: Optional[str] = None
+    """
+    The Loan Broker ID from which to withdraw First-Loss Capital. Must be provided if
+    the Amount is an MPT, or Amount is an IOU and issuer is specified as the Account
+    submitting the transaction.
+    """
+
+    amount: Optional[Amount] = None
```
Patel-Raj11: This cannot be XRP. We can claw back only IOU and MPTs.

xrpl/models/transactions/loan_broker_set.py
```diff
@@ -0,0 +1,62 @@
...
+    Arbitrary metadata in hex format. The field is limited to 256 bytes.
+    """
+
+    management_fee_rate: Optional[int] = None
+    """
+    The 1/10th basis point fee charged by the Lending Protocol Owner.
+    Valid values are between 0 and 10000 inclusive.
+    """
+
+    debt_maximum: Optional[int] = None
```
Patel-Raj11: Shouldn't this be Optional[str]?

xrpl/models/transactions/loan_broker_set.py
```diff
@@ -0,0 +1,62 @@
...
+
+from xrpl.models.required import REQUIRED
+from xrpl.models.transactions.transaction import Transaction
+from xrpl.models.transactions.types import TransactionType
+from xrpl.models.utils import KW_ONLY_DATACLASS, require_kwargs_on_init
+
+
+@require_kwargs_on_init
+@dataclass(frozen=True, **KW_ONLY_DATACLASS)
+class LoanBrokerSet(Transaction):
```
Patel-Raj11, line 19: Should we check the following?
1. Bounds for fields like CoverRateMinimum, CoverRateLiquidation, DebtMaximum, ManagementFeeRate, CoverRateLiquidation.
2. Bounds for Data and make sure its content is HEX?
3. Add unit tests for these checks?

xrpl/models/transactions/loan_set.py
```diff
@@ -0,0 +1,181 @@
...
+
+    counterparty: Optional[str] = None
+    """The address of the counterparty of the Loan."""
+
+    counterparty_signature: Optional[CounterpartySignature] = None
+    """
+    The signature of the counterparty over the transaction.
+    """
+
+    loan_origination_fee: Optional[int] = None
```
Patel-Raj11: Should we consistently have NUMBER XRPL type as `str` in python implementation. I think with `int` typing, users will assume that we can pass only integers, but we can pass scientific notations as well.

Similarly for other fields that have NUMBER XRPL type.

xrpl/models/transactions/loan_set.py
```diff
@@ -0,0 +1,181 @@
...
+    This enum represents those options.
+    """
+
+    TF_LOAN_OVER_PAYMENT = 0x00010000
+    """
+    Indicates that the loan supports overpayments.
+    """
+
+
+class LoanSetFlagInterface(TransactionFlagInterface):
```
Patel-Raj11, line 55: Once rippled codebase and spec solidifies, I think its good to add bound checks for other fields and their unit tests.

Snippet from xrpl.js implementation:

const MAX_DATA_LENGTH = 512
const MAX_OVER_PAYMENT_FEE_RATE = 100000
const MAX_INTEREST_RATE = 100000
const MAX_LATE_INTEREST_RATE = 100000
const MAX_CLOSE_INTEREST_RATE = 100000
const MAX_OVER_PAYMENT_INTEREST_RATE = 100000
const MIN_PAYMENT_INTERVAL = 60

xrpl/asyncio/transaction/main.py
```diff
@@ -516,6 +517,25 @@ async def _calculate_fee_per_transaction_type(
                 for raw_txn in batch.raw_transactions
             ]
         )
+    elif transaction.transaction_type == TransactionType.LOAN_SET:
```
Patel-Raj11, line 552: We should add unit/integration test to verify this logic.

tests/integration/transactions/test_lending_protocol.py
```diff
@@ -0,0 +1,186 @@
...
+        await fund_wallet_async(loan_issuer)
+
+        depositor_wallet = Wallet.create()
+        await fund_wallet_async(depositor_wallet)
+        borrower_wallet = Wallet.create()
+        await fund_wallet_async(borrower_wallet)
+
+        # Step-0: Set up the relevant flags on the loan_issuer account -- This is
+        # a pre-requisite for a Vault to hold the Issued Currency Asset
+        response = await sign_and_reliable_submission_async(
```
Patel-Raj11, line 445: Do we need this step, if XRP is the asset in the vault?

xrpl/asyncio/transaction/main.py
ckeshava: The existing integ test already uses this piece of code to automatically calculate the transaction fees

xrpl/models/transactions/loan_broker_set.py
ckeshava: updated in c6720156

xrpl/models/transactions/loan_broker_cover_clawback.py
ckeshava: updated in c6720156

xrpl/models/transactions/loan_broker_cover_clawback.py
ckeshava: updated in c6720156

xrpl/models/transactions/loan_set.py
ckeshava: updated in c6720156

tests/integration/transactions/test_lending_protocol.py
ckeshava: I'm not sure, I added it in because I encountered an error. However, I'm not able to remember the exact error now.

xrpl/models/transactions/loan_broker_set.py
ckeshava: if the data is not in hex format, it will fail the transaction-encode step. It will not be encoded into binary format at all.

ckeshava: updated in 029d65c2

xrpl/models/transactions/loan_set.py
ckeshava: 1ddf3484

xrpl/models/transactions/loan_broker_set.py
Patel-Raj11: > if the data is not in hex format, it will fail the transaction-encode step. It will not be encoded into binary format at all.

I think it will throw the error dump to the user. Not sure if thats what we want the user to see. Just check with some non hex value or out of bound value in unit test case and if thats readable, its ok it skip this check at model level.

xrpl/asyncio/transaction/main.py
Patel-Raj11: > The existing integ test already uses this piece of code to automatically calculate the transaction fees

I highly doubt that they will check `TransactionType.LOAN_SET` code path since its recently introduced. This is the most important part of this PR and getting the autofilled fee wrong will cause users to loose XRP. If it's a low hanging fruit we should test it. I recall there is a similar test for batch transaction. If it's not exported function maybe just assert the fee test in one of the integration test.

tests/integration/transactions/test_lending_protocol.py
Patel-Raj11: Can you remove this and run this test? If it fails we can add it back.

xrpl/asyncio/transaction/main.py
ckeshava: Which transaction-type are you concerned about? The integration test invokes `LoanBrokerSet`, `LoanSet` and many other Lending Protocol related transactions. All these transactions are autofilled.

The `Fees` portion of the transaction is autofilled using the `_calculate_fee_per_transaction_type` method (which I linked in my previous msg)

All the integration tests already test this code-path because `autofill` method invokes `_calculate_fee_per_transaction_type`

xrpl/models/transactions/loan_broker_set.py
ckeshava: I have added explicit hex validation in this commit: 4e5cf353
The error message should be readable now. I have also added unit tests in the binary-codec module.

xrpl/models/transactions/loan_set.py
```diff
@@ -0,0 +1,234 @@
...
+    )
+
+    MAX_DATA_LENGTH = 256
+    MAX_OVER_PAYMENT_FEE_RATE = 100_000
+    MAX_INTEREST_RATE = 100_000
+    MAX_LATE_INTEREST_RATE = 100_000
+    MAX_CLOSE_INTEREST_RATE = 100_000
+    MAX_OVER_PAYMENT_INTEREST_RATE = 100_000
+    MIN_PAYMENT_INTERVAL = 60
+
```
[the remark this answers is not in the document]
ckeshava: this discrepancy has been fixed in a later commit

xrpl/models/transactions/loan_set.py
```diff
@@ -0,0 +1,234 @@
...
+    loan_broker_id: str = REQUIRED
+    """
+    The Loan Broker ID associated with the loan.
+    """
+
+    data: Optional[str] = None
+    """
+    Arbitrary metadata in hex format. The field is limited to 256 bytes.
+    """
```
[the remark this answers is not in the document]
ckeshava: this discrepancy has been fixed in a later commit

tests/unit/models/transactions/test_loan_set.py
```diff
@@ -0,0 +1,214 @@
...
+                account=_SOURCE,
+                loan_broker_id=_ISSUER,
+                principal_requested="100000000",
+                start_date=int(datetime.datetime.now().timestamp()),
+                data="A" * 257,
+            )
+        self.assertEqual(
+            error.exception.args[0],
+            "{'LoanSet:data': 'Data must be less than 512 bytes.'}",
+        )
```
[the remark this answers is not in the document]
ckeshava: this discrepancy has been fixed in a later commit

xrpl/models/transactions/loan_broker_set.py
```diff
@@ -0,0 +1,108 @@
...
+        ):
+            errors["LoanBrokerSet:cover_rate_liquidation"] = (
+                "Cover rate liquidation must be between 0 and 100_000 inclusive."
+            )
+        if self.debt_maximum is not None and (
+            int(self.debt_maximum) < 0 or int(self.debt_maximum) > self.MAX_DEBT_MAXIMUM
+        ):
+            errors["LoanBrokerSet:debt_maximum"] = (
+                "Debt maximum must not be negative or greater than 9223372036854775807."
+            )
```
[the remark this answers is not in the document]
ckeshava: There are many fields which accept a `str` type for expressing numeric inputs. However, adding a try-catch block across all of them will unnecessarily increase the complexity of the code and increase performance overheads.

The ValueError thrown by the existing code base communicates the error clearly to the reader.

tests/integration/transactions/test_lending_protocol.py
ckeshava: I have removed these additional steps in this commit: 388553d7

xrpl/asyncio/transaction/main.py
Patel-Raj11: I am referring to a similar test like `test_basic_calculate_fee_per_transaction_type`, where you provide multiple signatures for `LOAN_SET`  transaction and assert for the `fee`.

ckeshava: Such a test is not feasible for the Lending Protocol spec. Please look at my note toward the end of this file.

The LoanSet transaction requires off-chain communication between the `LoanIssuer` and the `Borrower`. Whoever prepares the transaction needs to be aware of the `Number-of-Signers` on the counterparty account. The xrpl-py client library cannot `autofill` this value without reading the latest state of the counterparty account. However, this could increase the latency in submitting the `LoanSet` transaction.

I felt that it is easiest to let this field be populated by the `LoanInitiator` themselves.

Image

tests/unit/models/transactions/test_loan_broker_cover_clawback.py
```diff
@@ -0,0 +1,55 @@
+from unittest import TestCase
```
kuan121, line 1: I'm not familiar with our test coverage requirement. We have unit tests for 3 transaction types. Do we need unit tests for the other 6 transaction types?

tests/integration/transactions/test_lending_protocol.py
```diff
@@ -0,0 +1,372 @@
...
+from xrpl.models.currencies.xrp import XRP
+from xrpl.models.requests.account_objects import AccountObjectType
+from xrpl.models.response import ResponseStatus
+from xrpl.models.transactions.loan_manage import LoanManageFlag
+from xrpl.models.transactions.loan_set import CounterpartySignature
+from xrpl.models.transactions.vault_create import WithdrawalPolicy
+from xrpl.wallet import Wallet
+
+
+class TestLendingProtocolLifecycle(IntegrationTestCase):
```
kuan121, line 46: Could you provide a high-level summary of the scenarios you tested and explain why they are sufficient?

ckeshava: Have you read this section of the specification? The integration test in this file wallks-through these cases, as detailed in the spec.
image

ckeshava: @kuan121   The integration test covers all the new transactions introduced in the Lending Protocol amendment and also demonstrates that the existing Requests (`account_objects` and `ledger_entry`) work correctly with the Lending-Protocol ledger objects (`Loan` and `LoanBroker`). 

Did you have anything else in mind for the tests?

tests/unit/models/transactions/test_loan_broker_cover_clawback.py
ckeshava: The poetry tool enforces a minimum threshold test-coverage limit (I think 80%), I'm not aware of anything beyond that.

The unit-tests are expected to validate transactions before they are sent to a rippled node. Typically, the client-library can catch error-codes which begin with the prefix `tem...`. However, other types of error codes (`tec...`, `ter...`, etc) need read-access into the blockchain-state. Users might not want the additional latency incurred owing to the read of the blockchain state.

I have attempted to capture `tem` error codes in the unit-tests.

tests/integration/transactions/test_lending_protocol.py
kuan121: >Have you read this section of the specification? The integration test in this file wallks-through these cases, as detailed in the spec.

Thanks for pointing me to that section of the specification—I hadn’t paid close attention to it before, but it does address what I was looking for. That said, I might be missing something, but I don’t see tests covering the following scenarios:
1. First-Loss Capital (deposit, withdraw, and clawback)
2. Loan default and how First-Loss Capital is applied in that case
3. Loan deletion after it's matured and payed
4. LoanBroker deletion
5. Vault deletion
6. Unhappy paths, such as when a user tries to borrow more than the broker or vault holds (the current file seems to focus mainly on happy paths).

I’m not sure whether all of these need to be covered in integration tests, which is why I wanted to confirm whether the current coverage is sufficient, and if so, why.

tests/unit/models/transactions/test_loan_broker_cover_clawback.py
kuan121: I see. Thanks for clarifying that for me! I previously thought that we need unit tests for all new transaction types and ledger entry objects.

ckeshava: xrpl-py does not have models for all ledger-objects (however, xrpl.js has the interfaces for all the ledger-objects). In the client library integ-tests, I have validated the `LedgerEntry` and the `AccountObjects` Requests.

Let me know if you have any other questions.

tests/integration/transactions/test_lending_protocol.py
ckeshava: Hello @kuan121 , you are correct in your observation that this integ test is not exhaustive.
1. The rippled cpp implementation covers all these cases exhaustively in their unit-tests. The client libraries forward the transactions/requests to a rippled/clio node, hence the last-line-of-defense rests with the rippled software.
2. There is a different engineering team (Quality Enforcers) that performs the end-to-end testing of all the new amendments.

Historically, the client library integ tests have not been exhaustive, in order to avoid duplication of effort. Let me know if you have any other questions.

kuan121: Sounds good. Thanks for clarifying that for me. I don't have other questions.

xrpl/asyncio/transaction/main.py
Patel-Raj11: Ok

Patel-Raj11: @ckeshava As per my understanding of how Lending Protocol works, the counterparty_signature gets populated once the signed LoanSet transaction gets handed over to the second party. The signed LoanSet transaction will already have Fee in place and that cannot be changed. The second party will just fill in the loan_set.counterparty_signature values and submit the transaction as it is.

It's the first party that pays the transaction fees, we won't reach to this code path. I think this should be removed.

ckeshava: @Patel-Raj11  Are you suggesting that we remove the `if` condition on line 526?

What would be an appropriate interface for the fee-calculation method? Should we let the first party specify the "number of second-party-signers"? This info is necessary for the fee calculation.

Patel-Raj11: TransactionType.LOAN_SET should not have any special handling of fees calculation. So, I was suggesting to remove all changes in this file.

As explained in earlier message, we would not hit this code-path. If you want to try it, maybe try writing an integration test that hits this codepath with multi-signing enabled. As far as I understand counterparty_signature cannot be included in a transaction that needs to be signed. So counterparty_signature would always be None when this method is executed.

Patel-Raj11: To auto-fill the approach should be to use loan_set.counterparty instead of loan_set.counterparty_signature:

1. If loan_set.counterparty is present?
    a. Check loan_set.counterparty has multi-signing enabled. If so, get the count of minimum number of signers required to sign this transaction using another API call and perform the fee calculation that you are doing.
    b. If loan_set.counterparty does not have multi-sign enabled, then it would be net_fee.
 
2. If loan_set.counterparty is not present?
   a. Get the loan broker from LoanBrokerID and perform `#1.a and #1.b`.
   
We should write integration tests, if we implement this logic. I am planning to write such integration test in xrpl.js once, the multi-signing starts working in cpp implementation for LoanSet transaction.

ckeshava: 1.a: Even if we get the minimum quorum required for the counterparty, the actual number of signers can always exceed the minimum number of signers. In this case, the transaction will fail due to insufficient fee. 

Will users want this feature at all? It is easier to communicate the exact number of signers offline between the loan-broker and the borrower. This autofill procedure involves significant network communication and associated delays. If a user wants to use the library for time-sensitive tasks (high frequency trading, flash loans, etc), won't the delays be unacceptable?

Patel-Raj11: > 1.a: Even if we get the minimum quorum required for the counterparty, the actual number of signers can always exceed the minimum number of signers. In this case, the transaction will fail due to insufficient fee.

You made a valid point that if multi-signing is setup as three signers (1+1+1) with quorum as 2, it could happen that even though only two needs to sign, but all three can sign and the transaction is still valid. So, we need to find a balance between making the autofill feature usable and not over-estimate the fees. Using total number of signers will always ensure that the transaction passes (at a cost of increased transaction fees) and using quorum will give optimal cost but transaction might fail if more than the quorum signs that transaction. I would lean towards using  quorum to begin with and adjust to max signers if we receive many issues from library users. This calculations can always be changed without incurring a breaking change. Does this makes sense?

With the current setup, the autofill feature is totally not usable when multi-signing gets involved and I feel that Lending Protocol will be used by intuitions where multi-signing will be the norm.

> This autofill procedure involves significant network communication and associated delays. If a user wants to use the library for time-sensitive tasks (high frequency trading, flash loans, etc), won't the delays be unacceptable?

Autofilling is optional and they can completely skip it for such use cases. And on another note, we do call `server_info` to auto-fill Payment/Offer related transactions which would be used for high frequency trading.

ckeshava: @Patel-Raj11  As per our offline discussion, I have updated the design of the autofill feature for the LoanSet transaction. 2ba7a5a0

tests/integration/reqs/test_sign.py
```diff
@@ -0,0 +1,45 @@
...
+                principal_requested="100",
+                counterparty=_COUNTERPARTY_ADDRESS,
+            ),
+            client,
+            loan_issuer,
+        )
+        response = await client.request(
+            Sign(
+                transaction=loan_issuer_signed_txn,
+                signature_target="CounterpartySignature",
```
Patel-Raj11, line 32: `sign` method will not accept `signature_target`. It's `sign_for` that accepts the `signature_target`.

`sign_for` is currently not working in rippled as expected and as discussed offline, feel free to log Github but to add support for signature_target in `sign_for` and `submit_multisigned` methods.

tests/unit/models/requests/test_sign.py
```diff
@@ -67,6 +67,14 @@ def test_valid_secret(self):
         )
         self.assertTrue(request.is_valid())
 
+    def test_valid_signature_target(self):
```
Patel-Raj11, line 70: This as well should be reverted.

tests/integration/reqs/test_sign.py
ckeshava: Both `sign` and `submit` methods accept the `signature_target` field in their inputs. Both of them make use of a common parsing method: https://github.com/XRPLF/rippled/blob/4565cc280bf2d91f8f88ef74cc807ece4c91a243/src/xrpld/rpc/detail/RPCCall.cpp#L1000

This is also the reason my PR changes work correctly and pass the integration tests.

xrpl/asyncio/transaction/main.py
```diff
@@ -516,6 +549,47 @@ async def _calculate_fee_per_transaction_type(
...
+        # Compute the additional cost of each signature in the
+        # CounterpartySignature field. The  transaction fees depend on the number of
+        # signatures specified in the field.
+        loan_set = cast(LoanSet, transaction)
+        if loan_set.counterparty_signature is not None:
```
Patel-Raj11: `counterparty_signature` will always be None. Fees needs to be filled in by the first party and at that time it's always empty.
While the other party signs the transaction, they cannot fill in the Fees so this method does not get used.

tests/integration/reqs/test_sign.py
Patel-Raj11: Ah my bad, yes, `signature_target` is required in sign method, when we want to fill in CounterpartySignature (non-multisigning case).

xrpl/asyncio/transaction/main.py
Patel-Raj11: Overall looks good apart from this change.

ckeshava: thanks for identifying this case, I have removed it. 9188107b

xrpl/asyncio/transaction/main.py
```diff
@@ -516,6 +549,39 @@ async def _calculate_fee_per_transaction_type(
...
+        if loan_set.counterparty is not None:
+            counterparty_account = loan_set.counterparty
+        else:
+            # The sfCounterparty field is optional if the counterparty is the
+            # LoanBroker owner. Deduce the counterparty account from the LoanBroker
+            # ID.
+            loan_broker_object_info = await client._request_impl(
+                LedgerEntry(index=loan_set.loan_broker_id)
+            )
+            counterparty_account = loan_broker_object_info.result["node"]["Owner"]
```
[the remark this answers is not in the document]
ckeshava: > If loan_broker_id is invalid or the LedgerEntry response is malformed (missing node or Owner keys)
As part of the ledger-entry specification, these keys can be assumed to be present in all successful LedgerEntry RPC responses. There is no need for this try-catch block because we don't expect this code block to error, except for a black-swan event.

Patel-Raj11 (approved):

Patel-Raj11 (approved):

kuan121 (approved):

ckeshava:
Thank you to both the reviewers.

PR #868, opened 2025-09-15
chore: run prettier on everything

mvadari:

Title just about says it all. Prevents the auto-formatting for those who use VSCode

The changes are all just formatting, no functionality change.

### Context of Change

Does the auto-formatting for those who use VSCode

### Type of Change

- [x] Chore

- [ ] Yes

## Test Plan

CI passes.


.github/dependabot.yml
```diff
@@ -1,8 +1,8 @@
 version: 2
```
kuan121, line 1: I was wondering how often do we have to manually do this? I wonder if we should automate it instead rather than doing it this manually and periodically (if any)?

mvadari: We totally can. I'm only doing a one-off because I noticed it get triggered by the VSCode default format-on-save settings when I edited a JSON file in here. So I'd rather not automate it as a part of this PR.

kuan121 (approved):

PR #878, opened 2025-10-22
update dependencies

Patel-Raj11:

Ran - `poetry update`
https://python-poetry.org/docs/basic-usage/#updating-dependencies-to-their-latest-versions

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

## Test Plan

Tests in CI should pass


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.2.0 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.1 and should not be changed by hand.
```
mvadari: Shouldn't downgrade the Poetry version

.github/workflows/faucet_test.yml
```diff
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 
 env:
-  POETRY_VERSION: 2.1.1
+  POETRY_VERSION: 2.2.1
```
ckeshava, line 9: poetry v2.3.0 is now available.

ckeshava: https://python-poetry.org/blog/announcing-poetry-2.3.0/

xrpl/models/utils.py
```diff
@@ -298,7 +298,9 @@ def _is_kw_only_attr_defined_in_dataclass() -> bool:
         return True
     return sys.version_info.minor < 10
     """
-    return "kw_only" in dataclass.__kwdefaults__
+    return (
```
ckeshava, line 301: I'm curious why an upgrade of poetry prompted this code change, do you remember the error caused due to the lack of this code-change?

I would have expected to see this change from Python version upgrade, not poetry upgrade.

Patel-Raj11: I was attempting to upgrade all the dependencies using `poetry update` to their latest minor/patch version in one go. But it turned out that there were many breaking changes that required python version upgrades. So this change was attempted as it was only required for Python version < 3.10.

I'll close this PR as it has breaking changes to fix it in one go. We can upgrade what's necessary as per security alerts/semgrep findings.

PR #877, opened 2025-10-20
Support for Dynamic MPT (XLS-94d)

kuan121:
## High Level Overview of Change
Add support of Dynamic MPT (XLS-94d)

### Context of Change
* Specification: https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0094-dynamic-MPT
* rippled/xrpld PRs
  * https://github.com/XRPLF/rippled/pull/5705
  * https://github.com/XRPLF/rippled/pull/5797
  * https://github.com/XRPLF/rippled/pull/5820

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### Did you update CHANGELOG.md?

- [x] Yes
- [ ] No, this change does not impact library users

## Test Plan

1. Unit test: Add unit tests according to the failure conditions mentioned in https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0094-dynamic-MPT
2. Integration tests: Add integrations tests

PR #879, opened 2025-10-29
fix: better formatting handling in definitions script

mvadari:

Fix the definitions generation script so that it better handles formatting changes, like adding new lines

### Context of Change

Some newlines were introduced in `transactions.macro`, which weren't properly captured by the regex

### Type of Change

## Test Plan

Works locally.


xrpl/core/binarycodec/definitions/definitions.json
```diff
@@ -3367,6 +3386,8 @@
     "Hash160": 17,
     "Hash192": 21,
     "Hash256": 5,
+    "Int32": 10,
+    "Int64": 11,
```
kuan121, line 3390: I got [this comment](https://github.com/XRPLF/xrpl-py/pull/877#discussion_r2445605761) from coderabbit when I added these two lines in my PR. Is the comment valid and should we remove these two lines?

mvadari: It's not the only SType that isn't included in the codec, if you take a look at the full list. It doesn't matter as long as there are no fields that use the SType (which there aren't).

achowdhry-ripple (approved):

PR #895, opened 2025-12-19
Fix error in account_objects request type filter

### LINKED ISSUE #888: account_objects model unnecessarily fails validation when using canonical ledger entry name in "type" field
`rippled` version 2.4.0 added support for canonical ledger entry names ( https://github.com/XRPLF/rippled/pull/5271 ) but the `AccountObjects` request model hasn't been updated, so it rejects such usage.

example request:

```py
response = client.request(AccountObjects(
            account=wallet.address,
            ledger_hash=ledger_hash,
            type="Escrow",
            marker=marker
        ))
```

Should be valid but fails with this error:

```
{'type': "type is <class 'str'>, expected typing.Optional[xrpl.models.requests.account_objects.AccountObjectType]"}
```

ckeshava:
fix #888: Allow canonical case-insensitive ledger-object names in type filter. Please refer to the issue description for more context on this issue.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Added a new test with canonical `str` inputs to the type filer in the `account_objects` request.

### DIFF
--- a/xrpl/models/requests/account_objects.py
+++ b/xrpl/models/requests/account_objects.py
@@ -9,7 +9,7 @@ AccountLinesRequest instead.
 
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Any, Optional
+from typing import Any, Optional, Union
 
 from xrpl.models.requests.request import LookupByLedgerRequest, Request, RequestMethod
 from xrpl.models.required import REQUIRED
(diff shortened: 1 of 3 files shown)


xrpl/models/requests/account_objects.py
```diff
@@ -65,7 +65,7 @@ class AccountObjects(Request, LookupByLedgerRequest):
     """
 
     method: RequestMethod = field(default=RequestMethod.ACCOUNT_OBJECTS, init=False)
-    type: Optional[AccountObjectType] = None
+    type: Optional[Union[AccountObjectType, str]] = None
```
Patel-Raj11, line 68: I think same applies to `LedgerEntryType` based on https://github.com/XRPLF/rippled/pull/5271. Can you check and update `LedgerData` request as well?

mvadari, line 68: Instead of just using a `str` there should probably be another enum for the actual ledger entry types

ckeshava:
This work is very old, I have lost context of the issue. If somebody finds this work useful, please let me know. I can rebase and revive this work.

PR #916, opened 2026-02-23
Add workflow to perform vulns scan for pull requests

kuan121:

1. **Security vulnerability scanning in CI**: Add a new GitHub Actions workflow (`vulnerability_scan.yml`) that scans project dependencies for security vulnerabilities during CI. The workflow:
   - Generates a CycloneDX SBOM from Poetry dependencies
   - Scans the SBOM using Trivy for CRITICAL and HIGH severity vulnerabilities
   - Fails the CI if any vulnerabilities are found
   - Prints the vulnerability report to the CI log to provide developers with visibility and enable them to address any identified vulnerabilities.
2. Upgrade urllib3 to >= 2.6.3 to fix security vulnerabilities
3. **Drop Python 3.8 support**: Python 3.8 reached end-of-life in October 2024. The minimum supported Python version is now 3.9. This was required to use urllib3 >= 2.6.0 which only supports Python 3.9+.

### Context of Change

The release pipeline (`release.yml`) already includes vulnerability scanning, but it only runs during the release process. This change shifts security scanning left by running it on every PR and push to `main`, allowing developers to catch and fix vulnerabilities earlier in the development cycle. This also catches newly disclosed vulnerabilities in existing dependencies.

The workflow includes cost optimizations:
- Concurrency groups: cancels in-progress runs when new commits are pushed
- Skips draft PRs
- Caches Poetry installation and dependencies

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

1. Create a PR and verify the vulnerability scan workflow runs
2. Verify the workflow fails if a dependency with known CRITICAL/HIGH vulnerabilities is added
3. Verify the vulnerability report is printed in table format in the CI logs
5. Create a draft PR and verify the workflow is skipped
6. Verify CI passes on Python 3.9, 3.10, 3.11, 3.12, 3.13, and 3.14

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ maintainers = [
   { name = "Phu Pham", email = "ppham@ripple.com" },
 ]
 keywords = ["xrp", "xrpl", "cryptocurrency"]
-requires-python = ">=3.8.1"
+requires-python = ">=3.9"
 dynamic = [ "dependencies" ]
 
 [project.urls]
@@ -39,7 +39,7 @@ description = "A complete Python library for interacting with the XRP ledger"
 packages = [{ include = "xrpl" }, { include = "LICENSE" }]
 
 [tool.poetry.dependencies]
-python = ">=3.8.1,<4.0"
+python = ">=3.9,<4.0"
 base58 = "^2.1.0"
 ECPy = "^1.2.5"
 typing-extensions = "^4.13.2"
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -50,6 +50,8 @@ types-Deprecated = "^1.2.9"
 pycryptodome = "^3.23.0"
 
 [tool.poetry.group.dev.dependencies]
+# urllib3 >= 2.6.3 fixes CVE-2025-66418, CVE-2025-66471, CVE-2026-21441
+urllib3 = ">=2.6.3"
```
Patel-Raj11, line 54: Since `urllib3` is a transitive dependency, did we try upgrading the root dependency to see if `poetry.lock` gets the `urllib3 = ">=2.6.3"`? Or was it giving more than expected number errors and you had to pin it in `pyproject.toml`?

mvadari:
Can we deprecate Python 3.8 support in a separate PR? Otherwise it'll be a bit buried and hard to find later.

Issue #923, opened 2026-03-27
_calculate_precision incorrectly rejects valid IOU values with large exponents


sublimator:
## Bug

`verify_iou_value` rejects valid IOU values like `9999999999999999e79` with "Decimal precision out of range" even though rippled accepts them.

## Root cause

`_calculate_precision()` in `xrpl/core/binarycodec/types/amount.py` uses `quantize(Decimal(1))` which expands the number to its full decimal form (95 digits for `9999999999999999e79`), then counts all digits instead of just the 16 significant figures.

```python
def _calculate_precision(value: str) -> int:
    decimal_value = Decimal(value, Context(prec=MAX_PREC))
    if decimal_value == decimal_value.to_integral():
        return len(
            decimal_value.quantize(Decimal(1), context=Context(prec=MAX_PREC))
            .as_tuple()
            .digits
        )
    return len(decimal_value.normalize(Context()).as_tuple().digits)
```

## Fix

Count significant digits from the Decimal tuple directly:

```python
def _calculate_precision(value: str) -> int:
    decimal_value = Decimal(value)
    if decimal_value.is_zero():
        return 0
    return len(decimal_value.as_tuple().digits)
```

## Reproduction

```python
from xrpl.core.binarycodec import encode

# This is a valid XRPL IOU value (16 sig figs, exponent 79 within -96..80)
# rippled accepts it, xrpl-py rejects it
encode({"LowLimit": {
    "value": "9999999999999999e79",
    "currency": "USD",
    "issuer": "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh"
}})
# raises: Decimal precision out of range for issued currency value.
```

These values exist on mainnet ledger in historical transaction metadata (e.g. `AffectedNodes` → `ModifiedNode` → `FinalFields` → `LowLimit`).

PR #924, opened 2026-03-30
fix _calculate_precision

Patel-Raj11:

Fixed `_calculate_precision` in `xrpl/core/binarycodec/types/amount.py` to correctly count significant digits for IOU amounts. Previously, values like `9999999999999999e80` were incorrectly rejected despite having only 16 significant digits.

Fixes https://github.com/XRPLF/xrpl-py/issues/923

### Context of Change

The bug was in the integer path of `_calculate_precision`. When the value was an integer, it used `.quantize(Decimal(1))` which expanded trailing zeros into actual digits. For example, `9999999999999999e80` was expanded to a 96-digit number, causing the precision check to fail against `MAX_IOU_PRECISION (16)`.

The fix simplifies the function to always use Decimal.normalize(), which strips trailing zeros before counting digits

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan
Added unit tests for both failure and success conditions.

### DIFF
--- a/xrpl/core/binarycodec/types/amount.py
+++ b/xrpl/core/binarycodec/types/amount.py
@@ -5,7 +5,7 @@ See `Amount Fields <https://xrpl.org/serialization.html#amount-fields>`_
 
 from __future__ import annotations
 
-from decimal import Context, Decimal, InvalidOperation, localcontext
+from decimal import MAX_PREC, Context, Decimal, InvalidOperation, localcontext
 from typing import Any, Dict, Optional, Type, Union
 
 from typing_extensions import Final, Self
@@ -151,10 +151,11 @@ def verify_mpt_value(mpt_value: str) -> None:
 
 def _calculate_precision(value: str) -> int:
     """Calculate the precision of given value as a string."""
-    decimal_value = Decimal(value)
+    ctx = Context(prec=MAX_PREC)
+    decimal_value = Decimal(value, ctx)
     if decimal_value.is_zero():
         return 0
-    return len(decimal_value.normalize(Context()).as_tuple().digits)
+    return len(decimal_value.normalize(ctx).as_tuple().digits)
 
 
 def _verify_no_decimal(decimal: Decimal) -> None:
--- a/tests/unit/core/binarycodec/types/test_amount.py
+++ b/tests/unit/core/binarycodec/types/test_amount.py
@@ -181,6 +181,11 @@ class TestAmount(TestSerializedType):
             # Negative - 17 significant digits
             "-99999999999999999e80",
             "-1.1111111111111111",
+            # More than 28 significant digits (exceeds default Decimal context)
+            "12345678901234567890123456789",
+            "-12345678901234567890123456789",
+            "1.2345678901234567890123456789",
+            "-1.2345678901234567890123456789",
         ]
         for case in invalid_cases:
             self.assertRaises(


xrpl/core/binarycodec/types/amount.py
```diff
@@ -151,13 +151,9 @@ def verify_mpt_value(mpt_value: str) -> None:
...
 def _calculate_precision(value: str) -> int:
     """Calculate the precision of given value as a string."""
-    decimal_value = Decimal(value, Context(prec=MAX_PREC))
-    if decimal_value == decimal_value.to_integral():
-        return len(
-            decimal_value.quantize(Decimal(1), context=Context(prec=MAX_PREC))
-            .as_tuple()
-            .digits
-        )
+    decimal_value = Decimal(value)
```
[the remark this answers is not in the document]
Patel-Raj11: Addressed in [88317d6](https://github.com/XRPLF/xrpl-py/pull/924/commits/88317d6f63ceb64419c958c187a46749a0363650)

xrpl/core/binarycodec/types/amount.py
```diff
@@ -151,14 +151,11 @@ def verify_mpt_value(mpt_value: str) -> None:
...
-        return len(
-            decimal_value.quantize(Decimal(1), context=Context(prec=MAX_PREC))
-            .as_tuple()
-            .digits
-        )
-    return len(decimal_value.normalize(Context()).as_tuple().digits)
+    ctx = Context(prec=MAX_PREC)
+    decimal_value = Decimal(value, ctx)
+    if decimal_value.is_zero():
+        return 0
```
ckeshava, line 157: Do we need to handle the zero case separately? If so, should we add an explicit test case in the above test file?

xrpl/core/binarycodec/types/amount.py
```diff
@@ -151,14 +151,11 @@ def verify_mpt_value(mpt_value: str) -> None:
...
     """Calculate the precision of given value as a string."""
-    decimal_value = Decimal(value, Context(prec=MAX_PREC))
-    if decimal_value == decimal_value.to_integral():
-        return len(
-            decimal_value.quantize(Decimal(1), context=Context(prec=MAX_PREC))
-            .as_tuple()
-            .digits
-        )
-    return len(decimal_value.normalize(Context()).as_tuple().digits)
+    ctx = Context(prec=MAX_PREC)
```
ckeshava, line 154: I understand that the existing code uses `MAX_PREC`, however `MAX_PREC` is a really large value.

```
>>> from decimal import MAX_PREC
>>> print(f"The value of MAX_PREC is: {MAX_PREC}")
... 
The value of MAX_PREC is: 999999999999999999
```

If xrpl-py library is used in a client application, I'm concerned that an un-sanitized input can consume too much memory and cause OOM Exception DOS attack. For instance, an attacker con construct a transaction JSON with an absurdly large numeric input that consumes an unreasonable amount of memory.

Can we consider using the default-context (which has 28 significant digits) or set an upper-limit on the precision ?

xrpl/core/binarycodec/types/amount.py
Patel-Raj11: Yes, when zero is normalized in the below lines it's precision comes out to be 1, which is incorrect, so this condition was added. `_calculate_precision` is internal function and there is no code path that would exercise this condition as `verify_iou_value` already has this check, but we should keep this incase later calls to this internal function could be added without such explicit checks.

xrpl/core/binarycodec/types/amount.py
Patel-Raj11: Yes, its a good point. But by the time absurdly large numeric value is parsed into Decimal it would already have occupied large memory by being stored in a string. I think such input size validation should be enforced at the deserialization boundary, before strings are passed into the binary codec. Looks like it applies to all the fields and is beyond the scope of this fix PR.

The main aim of `_calculate_precision` internal function is to give the correct precision for the given input. So that it can be reused freely by some other functions.

Issue #992, opened 2026-04-17
Signing request reprs expose secret / seed / passphrase fields


ckeshava:
## Description

`BaseModel.__repr__` iterates over every field returned by `self.to_dict()` and includes the raw value. The request models used for remote signing — `Sign`, `SignAndSubmit`, `SignFor`, and `ChannelAuthorize` — declare `secret`, `seed`, `seed_hex`, and `passphrase` as plain optional fields, so those values appear verbatim in `repr()` (and in `str()`, which falls back to `__repr__`).

This is a distinct issue from #987. That one was about the exception message in `Wallet.__init__`; this one is about the request-object `__repr__` itself, and affects any code path that reprs / logs / traces a signing request.

## Repro

```python
from xrpl.models.requests.sign import Sign
from xrpl.models.transactions.payment import Payment

tx = Payment(
    account="rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh",
    destination="rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe",
    amount="100",
)
r = Sign(transaction=tx, seed="sEdTM1uX8pu2do5XvTnutH6HsouMaM2")

assert "sEdTM1uX8pu2do5XvTnutH6HsouMaM2" in repr(r)  # passes — seed is exposed
assert "sEdTM1uX8pu2do5XvTnutH6HsouMaM2" in str(r)   # passes — same
```

## Location

- [`xrpl/models/base_model.py:415-418`](https://github.com/XRPLF/xrpl-py/blob/8879e957/xrpl/models/base_model.py#L415-L418) — the generic `__repr__` that dumps every field.
- Request classes with secret-bearing fields:
  - [`xrpl/models/requests/sign.py`](https://github.com/XRPLF/xrpl-py/blob/8879e957/xrpl/models/requests/sign.py) (`secret`, `seed`, `seed_hex`, `passphrase`)
  - [`xrpl/models/requests/sign_and_submit.py`](https://github.com/XRPLF/xrpl-py/blob/8879e957/xrpl/models/requests/sign_and_submit.py)
  - [`xrpl/models/requests/sign_for.py`](https://github.com/XRPLF/xrpl-py/blob/8879e957/xrpl/models/requests/sign_for.py)
  - [`xrpl/models/requests/channel_authorize.py`](https://github.com/XRPLF/xrpl-py/blob/8879e957/xrpl/models/requests/channel_authorize.py)

## Impact / Severity

[...]

Issue #987, opened 2026-04-17
Secret Seed Exposed in Wallet Initialization Error Messages


ckeshava:
## Description

When wallet initialization fails with an invalid seed, the raw seed value is included verbatim in the exception message. Exception messages are commonly logged, sent to error-tracking systems (Sentry, Datadog), or displayed in UIs, resulting in secret material leakage.

```python
raise XRPLAddressCodecException(
    "Attempted to initialize a Wallet with "
    + "an invalid seed: "
    + seed              # <-- SEED EXPOSED
    + ". The cryptographic algorithm used is: "
    + wallet_algorithm
    + "\nError message: " + str(e)
)
```

## Location

[xrpl/wallet/main.py:83-92](https://github.com/XRPLF/xrpl-py/blob/8879e957/xrpl/wallet/main.py#L83-L92)

## Impact / Severity

**Critical** — Secret seed material leaks into exception messages, which are commonly logged, captured by error-tracking systems, or displayed in UIs. This can lead to wallet compromise.

## Adversarial Reviewer Notes

Verified as real. Constructing `Wallet(public_key="abc", private_key="def", seed="sInvalidSeedXXXXXXXXXXXXXX")` produces an exception containing the full seed string.

## Unit Test Patch

```python
from xrpl.wallet.main import Wallet

class TestSeedExposedInError(unittest.TestCase):
    def test_invalid_seed_not_in_exception(self):
        secret_seed = "sInvalidSeedXXXXXXXXXXXXXX"
        with self.assertRaises(Exception) as ctx:
            Wallet(public_key="abc", private_key="def", seed=secret_seed)
        self.assertNotIn(
            secret_seed, str(ctx.exception),
            "The seed value must not appear in the exception message",
        )
```

kuan121:
Close as issue has been fixed

Issue #986, opened 2026-04-17
Non-Cryptographic RNG Used for WebSocket Request IDs


ckeshava:
## Description

Request IDs are generated using `random.randrange()`, which uses a Mersenne Twister PRNG — not cryptographically secure. The ID space is limited to 1,000,000. An attacker who can observe a few request IDs can predict future ones, enabling response injection attacks on a shared WebSocket connection.

```python
from random import randrange
_REQ_ID_MAX: Final[int] = 1_000_000
request_dict["id"] = f"{request.method}_{randrange(_REQ_ID_MAX)}"
```

## Location

[xrpl/asyncio/clients/websocket_base.py:7, 21, 50](https://github.com/XRPLF/xrpl-py/blob/8879e957/xrpl/asyncio/clients/websocket_base.py#L7)

## Impact / Severity

**Critical** — Predictable request IDs from a non-cryptographic PRNG with a small ID space (1M) could allow an attacker to inject or correlate WebSocket responses.

## Adversarial Reviewer Notes

Verified as real. Source code imports `from random import randrange` and uses it directly for ID generation. Should use `secrets.randbelow`.

## Unit Test Patch

```python
import xrpl.asyncio.clients.websocket_base as ws_mod

class TestNonCryptoRNG(unittest.TestCase):
    def test_uses_secrets_module(self):
        import inspect
        src = inspect.getsource(ws_mod._inject_request_id)
        uses_secrets = "secrets" in src
        uses_random = "randrange" in src
        self.assertTrue(
            uses_secrets and not uses_random,
            "Request ID generation should use `secrets`, not `random.randrange`",
        )
```

Issue #977, opened 2026-04-17
No JSON Parse Error Handling in WebSocket Handler


ckeshava:
## Description

`json.loads(response)` in the WebSocket message handler is not wrapped in try-except. A single malformed JSON message from the server (or an attacker performing a MITM) will crash the entire handler task, silently killing the WebSocket listener without notification to the application.

## Location

[xrpl/asyncio/clients/websocket_base.py:136](https://github.com/XRPLF/xrpl-py/blob/8879e957/xrpl/asyncio/clients/websocket_base.py#L136)

## Impact / Severity

**Medium** — A single malformed message crashes the handler, silently killing all WebSocket communication. All subsequent requests hang indefinitely.

## Adversarial Reviewer Notes

Verified as real. The `_handler` method has no `try`/`except` block.

## Unit Test Patch

```python
class TestNoJsonErrorHandling(unittest.TestCase):
    def test_handler_wraps_json_loads_in_try(self):
        import inspect
        from xrpl.asyncio.clients.websocket_base import WebsocketBase
        src = inspect.getsource(WebsocketBase._handler)
        has_json_loads = "json.loads" in src
        has_try = "try:" in src or "except" in src
        self.assertTrue(not has_json_loads or has_try,
            "_handler uses json.loads without try/except error handling")
```

PR #993, opened 2026-04-17
Collection of bug-fixes

ckeshava:

  Security hardening pass: four fixes addressing silent data corruption and                                                                                                                     
  secret-material leakage in the client library. Fixes for #948, #987, #992, #986, #977. Please look at the individual commit messages and Issue descriptions for more context.                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      

<!--
Please include a summary/list of the changes.
If too broad, please consider splitting into multiple PRs.
If a relevant Asana task, please link it here.
-->


<!--
Please include the context of a change.
If a bug fix, when was the bug introduced? What was the behavior?
If a new feature, why was this architecture chosen? What were the alternatives?
If a refactor, how is this better than the previous implementation?

If there is a design document for this feature, please link it here.
-->

### Type of Change

<!--
Please check relevant options, delete irrelevant ones.
-->

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes
This is intended for defense-in-depth practice. It does not affect the external API surface area of the library

## Test Plan
  - [x] `tests/unit/core/binarycodec/types/test_amount.py::test_iou_underflow_rounds_to_zero` — `1e-82`, `1e-96`, `-1e-96` all serialize to canonical zero and round-trip to `"0"`.
  - [x] `tests/unit/wallet/test_wallet.py::test_invalid_seed_not_leaked_in_exception` — seed string is not present in `str(exc)`.                                                               
  - [x] `tests/unit/models/requests/test_sign.py::test_sensitive_fields_redacted_in_repr` — all four sensitive fields redacted in `repr()` and `str()`; `to_dict()` preserves raw values.       
  - [x] `tests/unit/models/requests/test_sign.py::test_non_sensitive_fields_appear_in_repr` — ordinary fields (nested transaction, scalars, enums) still render normally; `None`-valued         
  sensitive fields are not over-masked.  
<!--
Please describe the tests that you ran to verify your changes and provide instructions so that others can reproduce.
-->

<!--
## Future Tasks
For future tasks related to PR.
-->


xrpl/models/base_model.py
```diff
@@ -413,6 +431,17 @@ def __eq__(self: Self, other: object) -> bool:
...
+
+        Fields named in ``_SENSITIVE_FIELDS`` are replaced with a fixed
+        placeholder so secret material (seeds, passphrases, private keys)
+        never reaches logs, tracebacks, or error-reporting pipelines.
+        """
+        parts = []
+        for f in fields(self):
+            value = getattr(self, f.name)
+            if f.name in self._SENSITIVE_FIELDS and value is not None:
+                parts.append(f"{f.name}='***REDACTED***'")
```
kuan121, line 444: nit: Should we standardize on the same masking pattern (-HIDDEN-) used in xrpl/wallet/main.py:285? If you prefer ***REDACTED***, we should update `Wallet.__str__` in the same PR to keep the library consistent.

tests/unit/models/requests/test_sign.py
```diff
@@ -80,3 +80,47 @@ def test_valid_seed(self):
             transaction=_TRANSACTION, seed=_SEED, key_type=CryptoAlgorithm.SECP256K1
         )
         self.assertTrue(request.is_valid())
+
+    def test_sensitive_fields_redacted_in_repr(self):
```
kuan121, line 84: Here are some AI-generated test coverage improvements—incorporate them where they make sense.

## Suggested tests for the `__repr__` redaction

Each test below protects against a specific failure mode. The first three verify the **logic**; the last two verify the **wiring** (the `__init_subclass__` trick), which is the subtle part of this fix.

### 1. Seed value does not appear in repr

```python
def test_repr_redacts_seed_value():
    r = Sign(transaction=_valid_payment(), seed="sEdTM1uX8pu2do5XvTnutH6HsouMaM2")
    assert "sEdTM1uX8pu2do5XvTnutH6HsouMaM2" not in repr(r)
    assert "seed='-HIDDEN-'" in repr(r)
```

**Why:** This is the happy-path assertion from the original issue repro. Catches outright reverts or a subtle typo in the sensitive-field set name (e.g. `SENSITIVE_FIELDS` vs `_SENSITIVE_FIELDS`) that would silently match nothing.

---

### 2. `None`-valued secrets render as `None`, not redacted

```python
def test_repr_preserves_none_for_unset_secret_fields():
    r = Sign(transaction=_valid_payment(), seed="sEdTM1uX...")
    assert "secret=None" in repr(r)
    assert "passphrase=None" in repr(r)
    assert "seed_hex=None" in repr(r)
```

**Why:** Protects the "only hide real values" behavior. Without this test, someone could simplify the condition from `if name in SENSITIVE and value is not None:` to `if name in SENSITIVE:` — the seed-hiding test above still passes, but now unset fields misleadingly render as `secret='-HIDDEN-'`, which destroys the "is there a secret here?" signal.

---

### 3. All four secret-bearing fields are actually redacted

```python
@pytest.mark.parametrize("field_name,value", [
    ("secret",     "s1234567890abcdef"),
    ("seed",       "sEdTM1uX8pu2do5XvTnutH6HsouMaM2"),
    ("seed_hex",   "DEADBEEFDEADBEEFDEADBEEFDEADBEEF"),
    ("passphrase", "correct horse battery staple"),
])
def test_repr_redacts_every_sensitive_field(field_name, value):
    r = Sign(transaction=_valid_payment(), **{field_name: value})
    assert value not in repr(r)
```

**Why:** Test 1 only exercises `seed`. Without this, the sensitive-field set could silently lose `passphrase` or `seed_hex` during a refactor and no test would catch it.

---

### 4. A subclass that defines its own `__repr__` is not clobbered

```python
def test_init_subclass_respects_existing_repr():
    class CustomModel(BaseModel):
        def __repr__(self):
            return "CUSTOM"
        # ... implement required abstracts

    assert repr(CustomModel(...)) == "CUSTOM"
```

**Why:** Protects the guard `if "__repr__" not in cls.__dict__:` in `__init_subclass__`. That guard is easy to remove during a refactor ("why bother, `BaseModel.__repr__` works for everyone") — but removing it silently overwrites any subclass-defined `__repr__`. This test makes the guard load-bearing and documented.

---

### 5. Redaction applies to every secret-bearing request class, not just `Sign`

```python
@pytest.mark.parametrize("request_cls", [Sign, SignFor, SignAndSubmit, ChannelAuthorize])
def test_repr_redacts_across_all_signing_requests(request_cls):
    kwargs = {"seed": "sEdTM1uX8pu2do5XvTnutH6HsouMaM2"}
    if request_cls is ChannelAuthorize:
        kwargs.update({"channel_id": "0" * 64, "amount": "1000"})
    else:
        kwargs["transaction"] = _valid_payment()
    if request_cls is SignFor:
        kwargs["account"] = "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh"

    r = request_cls(**kwargs)
    assert "sEdTM1uX8pu2do5XvTnutH6HsouMaM2" not in repr(r)
```

**Why:** The fix works through `BaseModel.__init_subclass__`, which *should* uniformly install the redacting repr on every `BaseModel` subclass. This test proves it actually did — for all four request classes called out in the issue. Protects against a future change like `@dataclass(repr=True)` being added explicitly to one class, which would force dataclass to generate its own `__repr__` and silently reintroduce the leak on only that class.

tests/unit/core/binarycodec/types/test_amount.py
```diff
@@ -205,6 +205,36 @@ def test_from_value_issued_currency(self):
             amount_object = amount.Amount.from_value(json)
```
kuan121, line 205: Does it make sense for you to fix the Docker container issue for Python as part of this PR, since it’s blocking the merge? Alternatively, you could open a separate PR just for the Docker fix.

tests/unit/models/requests/test_sign.py
ckeshava: Please refer to `tests/unit/models/requests/test_sign.py` file for all the below tests:

> 1. Seed value does not appear in repr

This test is present in `test_sensitive_fields_HIDDEN_in_repr` method.

> 2. None-valued secrets render as None, not redacted

This test is present in the last part of the `test_non_sensitive_fields_appear_in_repr` test

> 3. All four secret-bearing fields are actually redacted

This test is present in `test_sensitive_fields_HIDDEN_in_repr` method.

xrpl/models/base_model.py
ckeshava: 46203d2074cbf895facd222aa1a0aef41e7513bd

tests/unit/models/requests/test_sign.py
ckeshava: thanks for the suggestions in 4, 5. It has been incorporated here: 11adb426

tests/unit/core/binarycodec/types/test_amount.py
ckeshava: dcb2a7a

tests/unit/models/test_base_model.py
```diff
@@ -70,6 +79,51 @@ def test_repr(self):
...
+                seed=seed,
+            ),
+            SignAndSubmit(transaction=transaction, seed=seed),
+            ChannelAuthorize(channel_id="0" * 64, amount="1000", seed=seed),
+        ]
+        for request in cases:
+            with self.subTest(request_cls=type(request).__name__):
+                rendered = repr(request)
+                self.assertNotIn(seed, rendered)
+                self.assertIn("seed='-HIDDEN-'", rendered)
```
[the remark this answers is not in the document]
ckeshava: this valid-looking seed does not harm anybody. I don't see the need for this change.

xrpl/asyncio/clients/websocket_base.py
```diff
@@ -133,7 +136,13 @@ async def _handler(self: Self) -> None:
...
         """
         async for response in cast(websocket_client.ClientConnection, self._websocket):
-            response_dict = json.loads(response)
+            try:
+                response_dict = json.loads(response)
+            except json.JSONDecodeError:
+                # Issue #977: a single malformed frame must not kill the
+                # handler task and leave the client deaf to every frame
+                # that follows.
+                continue
```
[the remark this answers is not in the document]
ckeshava: In the case of non Unicode chanracters (or) invalid JSON format in the response, I would expect to see a Fatal error and a stacktrace. I'm not convinced that silently catching+handling the error is the correct way to handle such use cases. The user should be informed that the rippled server is not responding with the expected response-format.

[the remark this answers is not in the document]
kuan121: @ckeshava Does this PR comment make sense?

kuan121 (dismissed):

xrpl/asyncio/clients/websocket_base.py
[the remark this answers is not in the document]
ckeshava: @coderabbitai  what would be an example response from the rippled server which causes:
(1) response_dict to be of non `Dict[str, Any]` type?
(2) trigger the `UnicodeDecodeError` / 1007-ConnectionClosed error in the `_handler` method?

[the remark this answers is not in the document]
ckeshava: a1cd8d83 
@coderabbitai I like your suggestion of printing malformed frames before dropping them. That suggestion has been incorporated.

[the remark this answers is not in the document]
ckeshava: ok, glad to clear this up.

kuan121 (dismissed):

xrpl/models/base_model.py
```diff
@@ -413,6 +431,17 @@ def __eq__(self: Self, other: object) -> bool:
...
-        return f"{type(self).__name__}({repr_items})"
+        """Returns a string representation of a BaseModel object.
+
+        Fields named in ``_SENSITIVE_FIELDS`` are replaced with a fixed
+        placeholder so secret material (seeds, passphrases, private keys)
+        never reaches logs, tracebacks, or error-reporting pipelines.
+        """
+        parts = []
+        for f in fields(self):
+            value = getattr(self, f.name)
```
cybele-ripple, line 444: Should there be some check for if `f.repr` is `false`?

xrpl/asyncio/clients/websocket_base.py
```diff
@@ -133,7 +136,16 @@ async def _handler(self: Self) -> None:
...
+                # Issue #977: a single malformed frame must not kill the
+                # handler task and leave the client deaf to every frame
+                # that follows. Surface the dropped frame so the failure is
+                # not silent.
+                # TODO: Introduce a module-level logging system across xrpl-py library
+                print(f"xrpl-py: skipping malformed WebSocket frame: {response!r}")
+                continue
 
             # if this response corresponds to request, fulfill the Future
             if "id" in response_dict and response_dict["id"] in self._open_requests:
```
cybele-ripple, line 151: Is there a chance that response_dict can return a non-dict value? Should we include a check for this?

xrpl/core/binarycodec/types/amount.py
```diff
@@ -209,7 +209,7 @@ def _serialize_issued_currency_value(value: str) -> bytes:
 
     if exp < MIN_IOU_EXPONENT or mantissa < MIN_IOU_MANTISSA:
         # Round to zero
-        _ZERO_CURRENCY_AMOUNT_HEX.to_bytes(8, byteorder="big", signed=False)
```
cybele-ripple, line 212: Do you know why this line did not cause issues before?

tests/unit/asyn/clients/test_websocket_base.py
cybele-ripple, line 1: I think that the directory should be renamed to `asyncio` instead of `asyn` because `websocker_base.py` is in `xrpl/asyncio/clients`

xrpl/wallet/main.py
```diff
@@ -81,11 +81,8 @@ def __init__(
...
-                    "Attempted to initialize a Wallet with "
-                    + "an invalid seed: "
-                    + seed
-                    + ". The cryptographic algorithm"
-                    + " used is: "
+                    "Attempted to initialize a Wallet with an invalid seed. "
+                    "The cryptographic algorithm used is: "
                     + wallet_algorithm
                     + "\nError message: "
                     + str(e)
```
cybele-ripple: Is there a way that information from `seed` can appear in `str(e)`

xrpl/core/binarycodec/types/amount.py
ckeshava: It was causing errors before, I suspect it was not discovered by the users. The unit tests show an example of a very-small-value not rounding down to zero. This represents an incompatibility rippled implementation.

xrpl/asyncio/clients/websocket_base.py
ckeshava: Here is a conversation with CodeRabbit on this question: https://github.com/XRPLF/xrpl-py/pull/993#discussion_r3156158224

If the `reponse_dict` variable has a non-`dict` value, that indicates that the rippled server is not honoring its contract. In that case, I would prefer that the user is loudly informed of the error instead of silently dropping the responses of a errorneous rippled server.

tests/unit/asyn/clients/test_websocket_base.py
ckeshava: <strike> thanks, fixed in a09b6857 </strike>

reverted back to the original version in a recent commit

xrpl/models/base_model.py
ckeshava: thanks, updated in a09b6857

xrpl/core/binarycodec/types/amount.py
ckeshava: image
  here is the unit test ^^

tests/unit/asyn/clients/test_websocket_base.py
ckeshava: Evidently, there was a valid reason for not naming this folder as `asyncio`. Here is an explanation: 
image


Basically, this folder would be discovered before the Python stdlib's `asyncio`. Hence, we cannot name-shadow this folder.

xrpl/wallet/main.py
ckeshava: thanks, I have updated the error message here: 1924fefb201a7f29d705fff2adb1998420ff0d10

The original version of the code had not considered potential errors thrown by `base58` external library. The new change should handle all such errors.

kuan121 (approved):

cybele-ripple (approved):

PR #1001, opened 2026-05-22
fix: preserve significant trailing zeros in `Amount.to_json` for IOU values

ckeshava:

The IOU branch of `Amount.to_json` finished by computing `value_str = str(value).rstrip("0").rstrip(".")` on a `Decimal` reconstructed from the 8-byte wire header. That stripping was only safe when `str(Decimal)` contained a decimal point. It does not for two common shapes of decoded value, and the first `rstrip("0")` then truncated significant data:

- **Integer-form decoded Decimal** (e.g. mantissa `10^15`, internal exponent `0`): `str(value)` is `"1000000000000000"`, and `rstrip("0")` eats every trailing zero — the JSON `value` field comes back as `"1"` for an on-ledger amount of `1,000,000,000,000,000`.
- **Scientific-notation decoded Decimal** (e.g. when the canonical exponent pushes the value outside Decimal's fixed-point window): `str(value)` is `"1.000000000000000E+20"`, and `rstrip("0")` shaves the trailing zero off the exponent itself, leaving `"1.000000000000000E+2"` — silently off by 18 orders of magnitude.

The fix forces fixed-point output via `f"{value:f}"` and guards the `rstrip` on the presence of a decimal point so it only trims fractional padding, never significant digits or exponent characters.

### Context of Change

This bug has been present since the IOU `to_json` path was written; the affected line is unchanged across recent releases. Practical impact for downstream callers includes Python indexers and exchanges that read IOU balances or `AffectedNodes` deltas from `xrpl.core.binarycodec.decode(...)` — the decoded `value` string can disagree with the authoritative on-ledger amount by up to fifteen orders of magnitude for canonical, spec-compliant inputs like `1e15`, `2.5e15`, or `1234567890000000`. The reference SDK xrpl.js (`ripple-binary-codec`) returns the correct value for the same wire bytes, so a service that ingests with xrpl-py and reports with xrpl.js would observe silent cross-SDK desynchronisation.

The fix is the one-line change recommended at the bug-site plus a guarded `rstrip` (only trims if the string contains a decimal point). No public API shape changes; the only observable difference is that decoded IOU `value` strings are now always canonical fixed-point (matching xrpl.js) instead of occasionally leaking scientific notation.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

New regression test `TestAmount::test_iou_to_json_preserves_significant_trailing_zeros` exercises both bug-faces plus a happy-path guard:

- `Amount.to_json` preserves all significant digits for integer-form decoded Decimals (`1000000000000000`, `-1000000000000000`, `1234567890000000`, `25e14`).
- `Amount.to_json` preserves a single trailing zero on `9999999999999990` / `-9999999999999990` (small x10 corruption that would slip past a loose equality check).
- `Amount.to_json` emits canonical fixed-point (never scientific notation) for the full mantissa-sign x exponent-sign matrix at `|exponent|=20`: `1e20`, `-1e20`, `1e-20`, `-1e-20`.
- `Amount.to_json` continues to strip fractional trailing zeros so `1.2000000` / `-1.2000000` still round-trip to `1.2` / `-1.2`.

The failing assertion text was verified against the pre-fix code (every case fails with a concrete corrupted value), and all 801 existing unit tests continue to pass.

### DIFF
--- a/.claude/skills/batch-deps-upgrade/SKILL.md
+++ b/.claude/skills/batch-deps-upgrade/SKILL.md
@@ -87,7 +87,7 @@ For each Python version, run the following in order:
        --detach \
        --publish 5005:5005 \
        --publish 6006:6006 \
-       --volume "$PWD/.ci-config/:/etc/opt/xrpld/" \
+       --volume "$PWD/.ci-config/:/etc/xrpld/" \
        --name xrpld-service \
        rippleci/xrpld:develop --standalone
      ```
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -46,7 +46,7 @@ jobs:
             --detach \
             --publish 5005:5005 \
             --publish 6006:6006 \
-            --volume "${{ github.workspace }}/.ci-config/":"/etc/opt/xrpld/" \
+            --volume "${{ github.workspace }}/.ci-config/":"/etc/xrpld/" \
             --name xrpld-service \
             ${{ env.XRPLD_DOCKER_IMAGE }} --standalone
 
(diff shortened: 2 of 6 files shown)


pdp2121 (approved):

kuan121 (approved):

ckeshava:
@pdp2121  @kuan121  thank you for the reviews.

PR #1002, opened 2026-05-26
feat: update definitions generation

pdp2121:

Replace the regex-based `definitions.json` generation with artifact-based approach that downloads server_definitions.json directly from rippled CI.

Added definitions validation step to CI (unit-test) to ensure definitions.json is being checked for correctness before any PR is merged.

### Context of Change

[rippled PR#6858](https://github.com/XRPLF/rippled/pull/6858) added server_definitions as artifact for every merged PRs, making it usable for downstream consumers

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

## Test Plan

CI passes.


tools/generate_definitions.py
```diff
@@ -1,376 +1,288 @@
-"""Script to generate the definitions.json file from rippled source code."""
+"""Script to generate the definitions.json file from rippled CI artifacts.
```
Patel-Raj11, line 1: Hi Phu, Based on https://github.com/XRPLF/rippled/pull/6858 and checking with Claude, the `server-definitions` artifact actually gets produced and retained for 3 days only when a cron job is run on each `weekday` (and not on every commit into develop branch, so many runs don't have those files). That means:

1. The pr:<N>, owner:branch, and arbitrary-branch flows do exactly the right API queries — but no PR-triggered or branch-push-triggered run ever uploads the artifact, so they consistently come up empty. So do we need to fix the upstream rippled workflow to allow file generate on every commit (on develop and feature branches)?

2. Also the `poetry run poe definitions` won't give server-definitions for the latest commit on develop branch, but instead it would give the file generated based on the most recent run, due to same reason as upstream generates the files on a scheduled basis. And if there are no commits for > 3 days, the `poetry run poe definitions` starts failing.

tools/generate_definitions.py
```diff
@@ -1,376 +1,288 @@
...
+
+def _download_artifact(repo: str, run_id: str, output_file: str) -> None:
+    """Download the artifact and write it as definitions.json."""
+    tmp_dir = tempfile.mkdtemp(prefix="server-definitions-")
+    try:
+        try:
+            _exec(
+                f"gh run download {run_id} --repo {repo}"
+                f' --name {ARTIFACT_NAME} --dir "{tmp_dir}"'
+            )
```
cybele-ripple: Is this an os command injection security concern?

tools/validate_definitions.py
```diff
@@ -0,0 +1,366 @@
...
+                errors.append(f"{key} must be a dict")
+                continue
+            for type_name, fields in local_value.items():
+                if not isinstance(fields, list):
+                    errors.append(
+                        f'{key} "{type_name}": value must be a list'
+                    )
+                    continue
+                for j, field in enumerate(fields):
+                    if not isinstance(field.get("name"), str):
```
cybele-ripple: Do we need to be concerned about a malformed format array where an element in that array is not a dict and instead is `null` or something else?

tools/generate_definitions.py
```diff
@@ -1,376 +1,288 @@
...
+        subprocess.run(
+            ["gh", "--version"], capture_output=True, text=True, check=True
+        )
+    except FileNotFoundError:
+        print(
+            "Error: GitHub CLI (gh) is required but not found.\n"
+            "Install from https://cli.github.com/",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
cybele-ripple: Should there be a check to ensure that the `gh` binary is authenticated?

tools/validate_definitions.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+def _exec(cmd: str) -> str:
+    result = subprocess.run(
+        cmd, shell=True, capture_output=True, text=True, check=True
+    )
+    return result.stdout.strip()
+
+
+def _check_gh_cli() -> None:
```
cybele-ripple: Should there be a check to ensure that the gh binary is authenticated?

PR #1035, opened 2026-08-27
fix: reject incomplete hex bytes in MPT metadata

rioyu123:

Reject odd-length MPT metadata at the hex-format boundary. The public validator returns its usual validation-message list, while the three transaction models reject incomplete hex bytes with `XRPLModelException`.

### Context of Change

`HEX_REGEX` checks characters but accepts strings such as `"A"` and `"abc"`. Those reach `bytes.fromhex` through `hex_to_str`, raising a plain `ValueError` that escapes `validate_mptoken_metadata` instead of returning a list.

The utility fix needs matching checks in `MPTokenIssuanceCreate`, `MPTokenIssuanceSet`, and `VaultCreate`: otherwise malformed hex would pass model validation with only an XLS-89 warning and fail later during serialization. This keeps invalid byte encoding distinct from advisory metadata-schema warnings. Empty-string metadata removal in `MPTokenIssuanceSet` is unchanged, and the shared regex is deliberately left alone.

### Type of Change

- [x] Tests

- [x] Yes

## Test Plan

- Regression tests cover one-character, lowercase, and truncated JSON hex inputs; non-hex characters; and the three transaction-model callers.
- The utility cases reproduced the uncaught `ValueError` on the original code. The model tests also caught the warning-only acceptance when only the utility was fixed.
- `python -m unittest discover -s tests/unit -q`: 1,074 tests run, 3 skipped, no failures.
- `python -m flake8 xrpl tests`: passes with the configured plugins.
- Black checks on the changed Python files: pass.

All checks were local. No ledger, faucet, wallet, or network integration tests were run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [[Unreleased]]
 
+### Fixed
+
+- Return a hex-format validation message instead of raising `ValueError` when `validate_mptoken_metadata` receives an odd-length hex string.
+- Reject odd-length `mptoken_metadata` with `XRPLModelException` in `MPTokenIssuanceCreate`, `MPTokenIssuanceSet`, and `VaultCreate` before serialization.
+
 ## [[5.1.0]]
 
 ### Added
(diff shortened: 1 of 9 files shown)


rioyu123:
Merged current main (`634ab7a`) and resolved the changelog conflict, keeping the new lending-protocol changes intact. The metadata validation fix is unchanged.

Unit tests: 1,105 run, 3 skipped, no failures. Lint and changed-file formatting checks passed. No ledger integration tests were run.

